PR #6, opened 2020-05-04
Dev/2003 plugins

vijs:


mhughesacn:
Hi Vijay,
The problem is with two files gstreamersoundplayer.cpp and .h.
This is an opensource repository but these two files have Broadcom Confidential headers which is not allowed.  The conditions in the headers are completely contrary to open sourcing.  This is not just a matter of removing the headers: you must be sure that any code here is not governed by these conditions.  All the complaining from scanning is because of these headers; protex does not find any code matches presumably because the code is not in the public domain and hence not in the code knowledge database.

vijs:
Abandoned

PR #11, opened 2020-05-14
Added AVInput, Bluetooth, DataCapture, Network, RemoteActionMapping, …

vijs:
…ScreenCapture and XCast plugins


vijs:
Added new rdkservices

PR #13, opened 2020-05-15
[build fix] make WebKitBrowser separately buildable

emutavchi:


mhughesacn:
Copyright exclusions file updated for vanilla Metrological import - scan now passes

PR #61, opened 2020-06-14
XRE-14362: Expanding DisplaySettings Api coverage

vinod-jacob546:
Reason for change: Adding the following API's to DisplaySettings
IsOutputHDR
setHdmiPreferences
getHdmiPreferences
isAudioEquivalenceEnabled
getSocIDFromSDK
isDynamicResolutionSupported
isDTCPSupported
isEnabled
getRestrictedResolution
Added Event Handler for audiomodeChanged

Test Procedure: check ticket
Risks: Low

### DIFF
--- a/DisplaySettings/DisplaySettings.cpp
+++ b/DisplaySettings/DisplaySettings.cpp
@@ -160,6 +160,15 @@ namespace WPEFramework {
             registerMethod("getTVHDRCapabilities", &DisplaySettings::getTVHDRCapabilities, this);
             registerMethod("getDefaultResolution", &DisplaySettings::getDefaultResolution, this);
             registerMethod("setScartParameter", &DisplaySettings::setScartParameter, this);
+            registerMethod("IsOutputHDR", &DisplaySettings::IsOutputHDR, this);
+            registerMethod("setHdmiPreferences", &DisplaySettings::setHdmiPreferences, this);
+            registerMethod("getHdmiPreferences", &DisplaySettings::getHdmiPreferences, this);
+            registerMethod("isAudioEquivalenceEnabled", &DisplaySettings::isAudioEquivalenceEnabled, this);
+            registerMethod("getSocIDFromSDK", &DisplaySettings::getSocIDFromSDK, this);
+            registerMethod("isDynamicResolutionSupported", &DisplaySettings::isDynamicResolutionSupported, this);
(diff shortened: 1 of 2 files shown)


vinod-jacob546:
Hi Bob,

In addition to the the API's that I have added in my pull request, the Netlfix DPI layer needs to make use of a few more APIs that are already available in DisplaySettings plugin. These are
getCurrentResolution
readEDID
getSinkAtmosCapability
getSoundMode
getConnectedVideoDisplays
getActiveInput
getSettopHDRSupport
getTVHDRCapabilities
Also, the below events too are being made use of 
ResolutionPreChange
ResolutionPostChange
connectedVideoDisplaysUpdated

From what I've seen in the DisplayInfo plugin, the only property that it exposes is DisplayInfo.1.displayinfo, from which only the below values can be fetched
{
    "jsonrpc": "2.0",
    "id": 1234567890,
    "result": {
        "totalgpuram": 381681664,
        "freegpuram": 358612992,
        "audiopassthrough": false,
        "connected": true,
        "width": 1280,
        "height": 720,
        "hdcpmajor": 2,
        "hdcpminor": 2,
        "hdrtype": "HDROff"
    }
}

Therefore, If I am to completely shift to using DisplayInfo in the place DisplaySettings (as you've mentioned that it wouldn't be proper to give Netflix that level control), then I'd have to bring all the above API's and events too in addition to the ones that I've added in the pull request.

Furthermore, the current DPI layer achieves functionality through DeviceSettings, which includes all the setters. Therefore, the Netflix app currently already has the level of control that you mentioned. And since all that DisplaySettings plugin is doing is giving Netflix app access to the same DeviceSettings APIs via itself, it seemed to be the more obvious choice for my activity of swapping the DeviceSettings APIs with rdkServices plugin APIs -  DisplaySettings would give a 1:1 mapping.

With regards to your other comments:

isEnabled – can this be renamed? Is What enabled? --> shall rename it to isVideoportEnabled

audioModeChanged – Why does Netflix need this?  --> this event leads to the invocation of getSinkAtmosCapability & getSoundMode in order to add/remove AudioProfiles.
(https://gerrit.teamccp.com/plugins/gitiles/rdk/components/cpc/netflix5_1/generic/+/stable2/partner/dpi/gstreamer/FileSystem.cpp#491)

Would like to know if you'd still prefer having DisplayInfo used in the place of DisplaySettings.

PR #80, opened 2020-06-23
[Thunder][System] DELIA-43983 System Plugin Issues

arun-madhavan-013:
Reason for change: getSerialNumber - Removed fork logic, updated setTimezoneDST to check boundary conditions.


arun-madhavan-013:
@anand-ky , please review and share feedback.

PR #86, opened 2020-06-29
RDK-28320: Enable Thunder Packager plugin

tdeva14:
Reason for change:
1. Modify Makefile to build Packager if enabled
2. Remove libprovision and libopkg dependency. Linking handled in recipe
3. Modify opkg header path
Test Procedure: Packager plugin must build fine if enabled
Risks: None

Signed-off-by: Deva <Deva_Thiyagarajan2@comcast.com>

### DIFF
--- a/Packager/CMakeLists.txt
+++ b/Packager/CMakeLists.txt
@@ -19,8 +19,6 @@ set(PLUGIN_NAME Packager)
 set(MODULE_NAME WPEFramework${PLUGIN_NAME})
 
 find_package(${NAMESPACE}Plugins REQUIRED)
-find_package(libprovision REQUIRED)
-find_package(LibOPKG REQUIRED)
 find_package(CompileSettingsDebug CONFIG REQUIRED)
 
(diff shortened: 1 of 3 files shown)


Packager/CMakeLists.txt
```diff
@@ -19,8 +19,6 @@ set(PLUGIN_NAME Packager)
 set(MODULE_NAME WPEFramework${PLUGIN_NAME})
 
 find_package(${NAMESPACE}Plugins REQUIRED)
-find_package(libprovision REQUIRED)
-find_package(LibOPKG REQUIRED)
```
goruklu, line 23: why do you remove LibOPKG ? that library is required for building the packager plugin,no?

tdeva14: Yes its required but we don't have a cmake config file for opkg currently. LDFLAGS for opkg is passed from recipe. I can remove REQUIRED if you want so that it doesn't stop the build.

vijs:
@tdeva14  Can you verify that desktop build is successful?

PR #92, opened 2020-07-06
RDK-28320: Enable Thunder Packager plugin

tdeva14:
Reason for change:
1. Makefile changes to build Packager plugin
2. libprovision dependency not required
3. Add cmake config file for OPKG
Test Procedure: Build Packager plugin
Risks: None

Signed-off-by: Deva <Deva_Thiyagarajan2@comcast.com>


vijs:
@tdeva14 Can you confirm if desktop build is successful?

Packager/CMakeLists.txt
```diff
@@ -18,8 +18,9 @@
 set(PLUGIN_NAME Packager)
 set(MODULE_NAME WPEFramework${PLUGIN_NAME})
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
 find_package(${NAMESPACE}Plugins REQUIRED)
-find_package(libprovision REQUIRED)
```
pwielders, line 22: Can we please make this conditional. We have customers using this packager relying on libprovision in combination with OPKG. So suggest to make this package optional, and if found, build it the old way, if not do it the comcast used way.
This way you do not break current deplyments

tdeva14:
@pwielders  @vijs Made the changed conditional. Could you please review and merge ?

Packager/CMakeLists.txt
```diff
@@ -18,7 +18,11 @@
 set(PLUGIN_NAME Packager)
 set(MODULE_NAME WPEFramework${PLUGIN_NAME})
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+if (BUILD_RDK)
```
pwielders, line 21: Why do we want to make this explicit ? Why can we not just say:
find_package(libprovision QUIET)

if (libprovision_FOUND)
  ...
else ()
   ...
endif ()

This way it work automagically if you install libprovision

PR #106, opened 2020-07-10
METROL-103: Update OpenCDMi

svdmaar:


goruklu:
> @goruklu you are aware of this revert I hope and that you can do it in the config. If not let me know.

@pwielders yes, we'll change the path in the config.

PR #113, opened 2020-07-13
XRE-14362: Adding methods and events to DisplayInfo & PlayerInfo

vinod-jacob:
Reason for change: Adding APIs to VAP plugins to provide
device properties for premium apps.
Plugins updated: DisplayInfo, PlayerInfo & HdcpProfile
Test Procedure: check ticket
Risks: Low

### DIFF
--- a/DisplayInfo/DisplayInfo.h
+++ b/DisplayInfo/DisplayInfo.h
@@ -16,12 +16,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
- 
+
 #pragma once
 
 #include "Module.h"
 #include <interfaces/IDisplayInfo.h>
 #include <interfaces/json/JsonData_DisplayInfo.h>
+#include <vector>
+
+#if defined(USE_IARM)
+#include "libIBus.h"
+#endif
 
 namespace WPEFramework {
 namespace Plugin {
@@ -85,11 +90,13 @@ namespace Plugin {
             , _notification(this)
         {
             RegisterAll();
+            DisplayInfo::_instance = this;
         }
 
         virtual ~DisplayInfo()
         {
             UnregisterAll();
+            DisplayInfo::_instance = nullptr;
         }
 
         BEGIN_INTERFACE_MAP(DisplayInfo)
@@ -123,6 +130,19 @@ namespace Plugin {
         void event_updated();
 
         void Info(JsonData::DisplayInfo::DisplayinfoData&) const;
+        uint32_t getCurrentResolution(const JsonObject& parameters, JsonObject& response);
+        uint32_t getConnectedVideoDisplays(const JsonObject& parameters, JsonObject& response);
+        uint32_t getTVHDRCapabilities(const JsonObject& parameters, JsonObject& response);
+        uint32_t getSettopHDRSupport(const JsonObject& parameters, JsonObject& response);
+        uint32_t IsOutputHDR(const JsonObject& parameters, JsonObject& response);
+        uint32_t setHdmiPreferences(const JsonObject& parameters, JsonObject& response);
+        uint32_t getHdmiPreferences(const JsonObject& parameters, JsonObject& response);
+        uint32_t isAudioEquivalenceEnabled(const JsonObject& parameters, JsonObject& response);
+        uint32_t readEDID(const JsonObject& parameters, JsonObject& response);
+
+        /*Events*/
+        void resolutionChanged(int width, int height);
+        void resolutionPreChange();
 
     private:
         uint8_t _skipURL;
@@ -130,6 +150,15 @@ namespace Plugin {
         Exchange::IGraphicsProperties* _graphicsProperties;
         Exchange::IConnectionProperties* _connectionProperties;
         Core::Sink<Notification> _notification;
+
+    private:
+        void getConnectedVideoDisplaysHelper(std::vector<string>& connectedDisplays);
+#if defined(USE_IARM)
+        static void ResolutionPreChange(const char *owner, IARM_EventId_t eventId, void *data, size_t len);
+        static void ResolutionPostChange(const char *owner, IARM_EventId_t eventId, void *data, size_t len);
+#endif
+    public:
+        static DisplayInfo* _instance;
     };
 
 } // namespace Plugin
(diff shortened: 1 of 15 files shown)


DisplayInfo/DisplayInfo.h
```diff
@@ -16,12 +16,17 @@
...
- 
+
 #pragma once
 
 #include "Module.h"
 #include <interfaces/IDisplayInfo.h>
 #include <interfaces/json/JsonData_DisplayInfo.h>
+#include <vector>
+
+#if defined(USE_IARM)
```
pwielders, line 27: I thought that the aim was to remove all DBus references. DBUS caters for a blackboard pattern architecture which, I think, in a closed system as ours is makes the dependencies hard to resolve.
The aim of the Thunder architecture is to create self-contained plugins that can do what they are supposed to do without delegating it to new components.
By not delegating, the dependencies are clear and resource consumption is reduced.

PlayerInfo/PlayerInfo.cpp
```diff
@@ -62,6 +77,23 @@ namespace Plugin {
         if (_player == nullptr) {
             message = _T("PlayerInfo could not be instantiated.");
         }
+
+        try
```
pwielders, line 81: Exception handling is an expensive feature. This is the embedded environment. We have exception handling by default turned off. Sofar we had never a need for it. You could also return a status from the Utils::IARM::init inidicating success or failre.

PlayerInfo/PlayerInfo.cpp
```diff
@@ -130,5 +174,217 @@ namespace Plugin {
...
+    {   //sample servicemanager response:
+        LOGINFOMETHOD();
+        bool success = true;
+        dsATMOSCapability_t atmosCapability;
+        try
+        {
+            device::AudioOutputPort aPort = device::Host::getInstance().getAudioOutputPort("HDMI0");
+            if (aPort.isConnected()) {
+                aPort.getSinkDeviceAtmosCapability (atmosCapability);
+                response["atmos_capability"] = (int)atmosCapability;
```
pwielders, line 187: For Metrological the Coding rule is applicable: in C++ code we use C++ casting. C casting does not explicitly state the type of casting and the severity of the cast can not be determined.

PlayerInfo/PlayerInfo.cpp
```diff
@@ -130,5 +174,217 @@ namespace Plugin {
         }
     }
 
+    uint32_t PlayerInfo::getSinkAtmosCapability (const JsonObject& parameters, JsonObject& response)
```
pwielders, line 177: As already explained in our Development Perception, we believe in Strong typed development. The compiler is our friend. We love the compiler to tell us when we are making mistakes, hence why we propose to use the JSON contract definitions and the using these strong type generated JSON onject classes. This way, errors are already discovered during compile time (not Intergration time) which safes a lot of development time.
Also the generated classes will avoid the large amount of magicals, like: "atmos_capability" in the code.

PlayerInfo/PlayerInfo.cpp
```diff
@@ -130,5 +174,217 @@ namespace Plugin {
...
+                {
+                    audioPort = "HDMI0";
+                }
+                else
+                {
+                    /*  * If HDMI is not connected
+                        * Get the SPDIF if it is supported by platform
+                        * If Platform does not have connected ports. Default to HDMI.
+                    */
+                    audioPort = "HDMI0";
```
pwielders, line 229: The concept of a plugin is to have magicals in the config (json part of the plugin header see any other plugin) I think the audioPort here would be such a config value. Default could still be "HDMI0" but at least it is easy to change.

helpers/utils.h
```diff
@@ -52,7 +54,7 @@
 #define returnResponse(success) \
```
pwielders, line 54: I would stronlgy discourage the use of the Preprocessor for these kind of constructions. I think C++ offers templates for these kind of constructions. Templates are namespace aware and are validated by the coompiler. The returnResponse now become a replace stament for any where in the code where the term returnResponse is used, also a variable.

helpers/utils.h
```diff
@@ -25,7 +25,9 @@
 #include <syscall.h>
```
pwielders, line 26: The architectural setup of Thunder/ThunderNanoservices is to have layers. Higher layers should only interact with lower layers. components in the same layer should not interact with the next component. The whole helper pardigm used is contradicting this architectural concept. 
TheSideCar approach has more plugins which are closed source. With this setup, those modules (if they wanted to use this helper stuff had to create a dependency on this reporsitory. This is what we would like to avoid hence why you have "supporting" libraries in thunder, (like e.g. Tracing for Tracing), Plugins for Plugins, WebSockets for http comunnication etc.
Functionality in this directory, if usefull should end up in the Thunder repository in the respective modules (Core/CryptAlgo/Tracing/WebSocket/com etc) to maintain a clean "unspaggettied" arhitecture

DisplayInfo/DisplayInfo.h
vinod-jacob: Hi Pierre,
Removing all DBUS references is indeed the final goal. In fact, we looked at possible ways to avoid making IARM calls from the plugin. However, we also wanted to make use of deviceSettings library calls to obtain the device properties. As it turns out, devicesettings library on the Arris Xi6 works in a "multi-app" mode which has a tight coupling with IARM (We are required to initialize the IARM bus and connect to it prior to making any devicesettings calls
MultiAppMode
). Due to this tight dependency, it was decided to keep the IARM calls in the plugin for now. We will be removing the IARM calls in the next phase.

helpers/utils.h
vinod-jacob: Moving the functionality of helpers directory to Thunder is perhaps better off taken up as a separate activity, as it would involve changes to almost all the other rdkservices plugins.
@bobseamon, your thoughts on this?

helpers/utils.h
vinod-jacob: Once again, as changes in helpers will touch all other plugins, sanitizing helpers module is perhaps better off taken up as a separate activity, @bobseamon?

vinod-jacob:
Will raise a new pull request, Addressing the comments

PR #115, opened 2020-07-13
RDKCMF-8648 Build fix for broadcom brcm972180hbc-refboard

sverkoye:
* set USE_IARM_BUS when USE_IARM set
* set USE_DS when USE_IARM_BUS set

These are now generic for all broadcom devices
and removed from the specific broadcom build configs


vijs:
Please include that ticket number in the commit log

sverkoye:
> Please include that ticket number in the commit log

Ok, I added ticket number in title. Is that sufficient?

PR #120, opened 2020-07-14
XIONE-818: Resolve thunder network api upcall protocol

kmortoza:
Reason for change: Do not notify address changes for unknown interfaces
Do not notify address changes for link local (non-routable) addresses
Do not add ip address attribute if not relevant
Test procedure: Full network functionality tests
Risks: Medium

### DIFF
--- a/Network/Network.cpp
+++ b/Network/Network.cpp
@@ -604,8 +604,10 @@ namespace WPEFramework
                 if  (!m_netUtils.isIPV6LinkLocal((const std::string) ipv6Addr))
 #endif
                     params["ip6Address"] = ipv6Addr;
+#ifdef NET_NO_LINK_LOCAL_ANNOUNCE
                 else
                     return;
+#endif
             }
             if (ipv4Addr != "")
             {
@@ -613,8 +615,10 @@ namespace WPEFramework
                 if (!m_netUtils.isIPV4LinkLocal((const std::string) ipv4Addr))
 #endif
                     params["ip4Address"] = ipv4Addr;
+#ifdef NET_NO_LINK_LOCAL_ANNOUNCE
                 else
                     return;
+#endif
             }
             params["status"] = string (acquired ? "ACQUIRED" : "LOST");
             sendNotify("onIPAddressStatusChanged", params);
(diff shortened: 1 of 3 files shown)


Network/Network.cpp
```diff
@@ -240,9 +240,11 @@ namespace WPEFramework
                     {
                         JsonObject interface;
                         std::string iface = m_netUtils.getInterfaceDescription(list.interfaces[i].name);
+#ifdef NET_DEFINED_INTERFACES_ONLY
```
tnykkn, line 243: Probably not needed, as an event with an empty "interface" parameter would not really be useful to anybody.

Network/Network.cpp
```diff
@@ -598,14 +600,18 @@ namespace WPEFramework
             params["interface"] = m_netUtils.getInterfaceDescription(interface);
             if (ipv6Addr != "")
             {
+#ifdef NET_NO_LINK_LOCAL_ANNOUNCE
                 if  (!m_netUtils.isIPV6LinkLocal((const std::string) ipv6Addr))
+#endif
```
tnykkn: When "NET_NO_LINK_LOCAL_ANNOUNCE" is not defined, this will result in a dangling "else" and thus a compilation error.

Network/Network.cpp
```diff
@@ -598,14 +600,18 @@ namespace WPEFramework
...
                     params["ip6Address"] = ipv6Addr;
                 else
                     return;
             }
             if (ipv4Addr != "")
             {
+#ifdef NET_NO_LINK_LOCAL_ANNOUNCE
                 if (!m_netUtils.isIPV4LinkLocal((const std::string) ipv4Addr))
+#endif
```
tnykkn: When "NET_NO_LINK_LOCAL_ANNOUNCE" is not defined, this will result in a dangling "else" and thus a compilation error.

Network/Network.cpp
kmortoza: Corrected

Network/Network.cpp
kmortoza: Corrected

PR #121, opened 2020-07-14
DELIA-44326 : [Thunder][Dis[laySettings] - HDMI Audio Delay Issues

sborushevsky:
Fixed exception handling fromp stoi call.

### DIFF
--- a/DisplaySettings/DisplaySettings.cpp
+++ b/DisplaySettings/DisplaySettings.cpp
@@ -1174,8 +1174,8 @@ namespace WPEFramework {
 			int compresionLevel = 0;
             try {
                 compresionLevel = stoi(sCompresionLevel);
-            }catch (const device::Exception& err) {
-               LOG_DEVICE_EXCEPTION1(sCompresionLevel); 
+            }catch (const std::exception &err) {
+               LOGERR("Failed to parse compresionLevel '%s'", sCompresionLevel.c_str());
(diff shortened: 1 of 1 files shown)


sborushevsky:
> Please keep both the catch statements and add logging about the exception.

LOG_DEVICE_EXCEPTION1 is for device::Exception, not for std::exception.
std::exception::what() returns "stoi", which is not very descriptive about what happened, so I didn't use it.

anand-ky:
> > Please keep both the catch statements and add logging about the exception.
> 
> LOG_DEVICE_EXCEPTION1 is for device::Exception, not for std::exception.
> std::exception::what() returns "stoi", which is not very descriptive about what happened, so I didn't use it.

ok, sounds good. As discussed, pls make similar change in other places where stoi is used.

sborushevsky:
> > > Please keep both the catch statements and add logging about the exception.
> > 
> > 
> > LOG_DEVICE_EXCEPTION1 is for device::Exception, not for std::exception.
> > std::exception::what() returns "stoi", which is not very descriptive about what happened, so I didn't use it.
> 
> ok, sounds good. As discussed, pls make similar change in other places where stoi is used.

Created https://github.com/rdkcentral/rdkservices/pull/123 for the rest of stoi calls.

vijs:
Merged https://github.com/rdkcentral/rdkservices/pull/124

PR #150, opened 2020-07-28
METROL-103: Updates WebKitBrowser

svdmaar:
Copies over work on WebKitBrowser from ThunderNanoServices.


WebKitBrowser/CMakeLists.txt
```diff
@@ -15,33 +15,11 @@
...
 
-find_package(ProxyStubGenerator REQUIRED)
-find_package(JsonGenerator REQUIRED)
-
-ProxyStubGenerator(INPUT ${CMAKE_CURRENT_SOURCE_DIR}/interfaces)
-JsonGenerator(DOCS CODE STUBS
-  INPUT ${CMAKE_CURRENT_SOURCE_DIR}/WebKitBrowserPlugin.json
-  IFDIR  ${CMAKE_CURRENT_SOURCE_DIR}/interfaces/json/
-  OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/interfaces/json/
-)
```
emutavchi, line 43: We need custom interface for RDK use cases. Please keep it.

WebKitBrowser/CMakeLists.txt
```diff
@@ -15,33 +15,11 @@
...
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/../cmake/")
-
-include(CmakeHelperFunctions)
-
-# Library installation section
-string(TOLOWER ${NAMESPACE} STORAGE_DIRECTORY)
-
-if(BUILD_REFERENCE)
-    add_definitions(-DBUILD_REFERENCE=${BUILD_REFERENCE})
-endif()
```
emutavchi, line 30: This is needed to make it possible to build just a browser plugin. Please don't remove it.

WebKitBrowser/CMakeLists.txt
```diff
@@ -118,13 +113,18 @@ target_include_directories(${MODULE_NAME}
         ${WPE_WEBKIT_INCLUDE_DIRS}
         ${WPE_WEBKIT_INCLUDE_DIRS}/WPE
         ${GLIB_INCLUDE_DIRS}
-        ${CMAKE_CURRENT_SOURCE_DIR}
-        ${CMAKE_CURRENT_SOURCE_DIR}/interfaces)
+        #${CMAKE_CURRENT_SOURCE_DIR}
```
emutavchi, line 116: We need to generate and build a custom interface. Keep this.

WebKitBrowser/InjectedBundle/RequestHeaders.cpp
```diff
@@ -25,6 +25,8 @@
 #include <WPE/WebKit/WKBundleFrame.h>
 #include <WPE/WebKit/WKURL.h>
 
+#include <interfaces/json/JsonData_WebKitBrowser.h>
+
```
emutavchi, line 29: Please keep this source file as it is.

WebKitBrowser/InjectedBundle/RequestHeaders.cpp
```diff
@@ -41,35 +43,8 @@ static PageHeaders s_pageHeaders;
 
 bool ParseHeaders(const string& json, Headers& out)
 {
-    struct HeadersData : public Core::JSON::Container
```
emutavchi, line 44: This is here to avoid extra dependency on WebKitBrowser interfaces. Please keep it as it is.

WebKitBrowser/InjectedBundle/main.cpp
```diff
@@ -97,28 +109,117 @@ static class PluginHost {
...
+        if (messageString.length() > maxStringLength) {
+            messageString = messageString.substr(0, maxStringLength);
+        }
+
+        // TODO: use "Trace" classes for different levels.
+        TRACE_GLOBAL(Trace::Information, (messageString));
+    }
+
+    WKBundleRef _bundle;
+    WebKitScriptWorld* _scriptWorld;
```
emutavchi, line 219: This would be better in a separate source file. I'd also suggest sticking with WK C API, or migrate the whole thing to GLib APIs which will be much bigger effort.

WebKitBrowser/CMakeLists.txt
pwielders: agree. @svdmaar can we add this.

WebKitBrowser/CMakeLists.txt
pwielders: This needs to be in by definition, would expect it not even to compile with out this. Will discuss with @svdmaar if it is done diffrently now.

WebKitBrowser/CMakeLists.txt
pwielders: Can we have a discussion on custom interfaces? The only reason for having an interface is if someone else is using it, would this be an interface between the WebKitPlugin and the injected bundle ?

WebKitBrowser/InjectedBundle/main.cpp
pwielders: Unfortunatly we have to support both. Hence why there is yet another Pull Request lined up to fix this. The C-API is the old one (a lot of our customers still use it) The GLIB API is the new one and we are moving towards it. So both need to be avialble (using a pre-processor flag)

WebKitBrowser/InjectedBundle/RequestHeaders.cpp
pwielders: If the "interfaces/json/JsonData_WebKitBrowser.h" is defining the struct HeadersData : public Core::JSON::Container I do prefer the way it is. If the interface changes, compile time we will notice a problem building the Injectable bundle and I still think the compiler is my best friend :-) I would not like ot see the issue while I am testing in the field.
However, this statement is whiyhout nowing the full context. Lets discuss.

pwielders:
@emutavchi I think we now need to make sure all functionality gets together. Your pull request needs to be merged first than we need to "add" that functionality, in your pull request, to this pull request. As discussed I am expecting than that we can move the interface definitions into the interfaces directory of Thunder again (these definitions are used by the DIALServer and our private ProvisioingPlugin). The latter does not have a dependency on this repository, only on Thunder (interfaces) and we want to keep the dependecy graph as simple as possible. 
After your merge request is merged, Sander or I will pick up the changes you did in your pull request and solve the issue I just raised here for revisiting the WebKitBrowser (https://github.com/rdkcentral/rdkservices/issues/175). 
So please reach out if you are going to change anything beyond the pull request that is currently pending.

MFransen69:
This PR is outdated, a new one was created in the mean time

PR #162, opened 2020-08-06
RDK-29072 Browser Plugin on PlatCo

emutavchi:


emutavchi:
> Hangdetector is good functionality, However would like to relay the action to the Thunder Frmaework to properly shutdown (Part of the basic framework, just need to trigger, and report, it)

As discussed this can be done later after Thunder provides "Hang" deactivation reason (https://github.com/rdkcentral/Thunder/issues/216)

emutavchi:
@pwielders, thanks for review. I reworked the hang detector in https://github.com/rdkcentral/rdkservices/pull/176/commits/3b3ed044191e456d6ac24bc384861de39b97e41c

PR #176, opened 2020-08-13
RDK-29486: Add option to load about:blank on suspend

emutavchi:
Enabled for HtmlApp and LightningApp callsigns


emutavchi:
> Don't we want to set the "set URL back on the resume than ?

The idea is that this suspends the browser instance and not the app. I guess the loading of the previous URL can be added, but the intended use case is for the user of the browser to set the desired URL before resuming the page.

PR #183, opened 2020-08-14
RDK-29389:API to query AudioCapabilty of the stb

akhilsarada:
Reason for change: Thunder changes for feature
Adding 2 API getSettopAudioCapabilities and
getSettopMS12Capabilities.
Test Procedure: Refer JIRA (RDK-27000)
Risks: Low
Related chage topic: RDK_27000_AudioCapability_query
Signed-off-by: Akhil Baby Sarada <Akhil_Sarada@comcast.com>


pwielders:
Metrological is adding this information as well but than on the player. There is a distinction in the plugin setup where:
DisplayInfo -> Deliver all information on the properties of the display (visualisation)
PlayerInfo -> Deliver all information on the properties of a player, like audio and video capabilities.
Given this definition we are adding the MS12 proeprties (wich are Audio properties and thus related to player capabilities) to the PlayerInfo.

akhilsarada:
@pwielders  Thanks for the comment . I will check and update

akhilsarada:
@apatel859  : I have updated according to you comments
@pwielders : Currently all the audio related API's are implemented in DisplaySettings plugin, hence I followed the same here. If we move just these implementation then it will be confusing since some will be in DisplaySettings and other in player settings.
From your comment I understand all Device specific inforamtion has to be form PlayerInfo plugin. I believe this can be taken up as a separate ticket for moving implementation of all such api's.
For now can you provide +1 to merge this with Display settings ?

akhilsarada:
The cahnges has been merged to 2009 sprint branch

PR #197, opened 2020-08-26
RDK-27957 Additional changes for TextToSpeech service

nrajan002c:

### DIFF
--- a/TextToSpeech/doc/TextToSpeechPlugin.md
+++ b/TextToSpeech/doc/TextToSpeechPlugin.md
@@ -240,7 +240,7 @@ TextToSpeech interface events:
 | Event | Description |
 | :-------- | :-------- |
 | [onspeechstart](#event.onspeechstart) | Notifies when speech starts |
-| [onspeechcancel](#event.onspeechcancel) | Notifies when speech is cancelled |
+| [onspeechinterrupted](#event.onspeechinterrupted) | Notifies when the current speech is cancelled |
 | [onnetworkerror](#event.onnetworkerror) | Notifies when network error is occurred |
 | [onplaybackerror](#event.onplaybackerror) | Notifies when playback error |
 | [onspeechcomplete](#event.onspeechcomplete) | Notifies when speech is completed |
@@ -270,10 +270,10 @@ Notifies when speech starts.
     }
 }
 ```
-<a name="event.onspeechcancel"></a>
-## *onspeechcancel <sup>event</sup>*
+<a name="event.onspeechinterrupted"></a>
+## *onspeechinterrupted <sup>event</sup>*
 
-Notifies when speech is cancelled.
+Notifies when the current speech is cancelled.
 
 ### Parameters
 
@@ -287,7 +287,7 @@ Notifies when speech is cancelled.
 ```json
 {
     "jsonrpc": "2.0",
-    "method": "client.events.1.onspeechcancel",
+    "method": "client.events.1.onspeechinterrupted",
     "params": {
         "speechid": 1
     }
--- a/TextToSpeech/impl/TTSManager.cpp
+++ b/TextToSpeech/impl/TTSManager.cpp
@@ -270,7 +270,7 @@ TTS_Error TTSManager::resume(uint32_t id) {
 TTS_Error TTSManager::shut(uint32_t id) {
     TTSLOG_TRACE("Shut");
 
-    if(m_speaker && m_speaker->cancelCurrentSpeech(id))
+    if((id != 0) && m_speaker && m_speaker->cancelSpeech(id))
         return TTS_OK;
 
     return TTS_FAIL;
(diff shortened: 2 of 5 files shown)


TextToSpeech/test/TTSThunderAPITest.cpp
```diff
@@ -263,7 +263,7 @@ int main(int argc, char *argv[]) {
                         &Handlers::onSpeechCancelledHandler) == Core::ERROR_NONE) {
                 LOGERR("Failed to Subscribe notification handler : onspeechcancelledHandler");
```
varumugam123, line 264: This could be removed until we support queuing mechanism in TTS

TextToSpeech/impl/TTSSpeaker.cpp
```diff
@@ -224,16 +226,10 @@ bool TTSSpeaker::isSpeaking(uint32_t id) {
 bool TTSSpeaker::cancelCurrentSpeech(uint32_t id) {
     TTSLOG_VERBOSE("Cancelling current speech");
     bool status = false;
-    if(m_isSpeaking) {
+    if(m_isSpeaking && m_currentSpeech && (m_currentSpeech->id == id)) {
```
varumugam123, line 229: This API can be changed as "cancelSpeech(uint32_t id)". If 0 is passed as argument, it should cancel any speech whichever is in progress - this is required for TTSSpeaker::reset(). If it is non-zero, it should match the current speech and cancel only if it succeeds.You can have a validation in TTSManager for (speechid != 0), to avoid any clients cancelling speech from someother client

With the present logic in the review, functionality of reset() is broken i.e Speaker couldn't do preemptive speeches.

TextToSpeech/doc/TextToSpeechPlugin.md
```diff
@@ -240,7 +240,7 @@ TextToSpeech interface events:
 | Event | Description |
 | :-------- | :-------- |
 | [onspeechstart](#event.onspeechstart) | Notifies when speech starts |
-| [onspeechcancel](#event.onspeechcancel) | Notifies when speech is cancelled |
```
varumugam123, line 243: We might need this cancelling functionality when we support queuing in TTS. I understand like the current requirement is to support only the preemptive speeches.

@ramasat could confirm.

When queuing is supported we should have a call to cancell all speeches. And then it would make sense to send "canceled" events to the speeches from the queue, the progressing speech would get "interrupted" event.

varumugam123 (changes requested):

TextToSpeech/test/TTSThunderAPITest.cpp
nrajan002c: Yes. I have removed the cancel event subscription and handler.

TextToSpeech/impl/TTSSpeaker.cpp
nrajan002c: I have added the id==0 handling. So this will make sure reset() works.

TextToSpeech/doc/TextToSpeechPlugin.md
ramasat: Queue'ing is required for "offline" TTS. If we already have the support, I would suggest adding the cancel event for it. If not, we can plan to add both queue'ing and canceled event together in a separate commit later.

nrajan002c: > Queue'ing is required for "offline" TTS. If we already have the support, I would suggest adding the cancel event for it. If not, we can plan to add both queue'ing and canceled event together in a separate commit later.

We will add the cancel event  as part of queue'ing use case (Offline TTS) . I feel that would be better .

TextToSpeech/impl/TTSManager.cpp
```diff
@@ -270,7 +270,7 @@ TTS_Error TTSManager::resume(uint32_t id) {
 TTS_Error TTSManager::shut(uint32_t id) {
     TTSLOG_TRACE("Shut");
 
-    if(m_speaker && m_speaker->cancelCurrentSpeech(id))
+    if((id != 0) && m_speaker && m_speaker->cancelCurrentSpeech(id))
```
varumugam123, line 273: I think the call now should be cancelSpeech(id).

varumugam123 (changes requested):

TextToSpeech/impl/TTSManager.cpp
nrajan002c: Yes. I corrected the api now.

varumugam123 (approved):

TextToSpeech/doc/TextToSpeechPlugin.md
ramasat: Ok, sounds good.

ramasat (approved):
Changes looks good to me.

PR #215, opened 2020-08-31
[RDK-28710] Add OCIContainer plugin

goruklu:


emutavchi (dismissed):
> OCIContainer/test/TestBundles/sleepy.tar.gz
I'd suggest keeping binaries out of the source tree,  otherwise looks good to me. Although, I don't have much expertise in this area.

vijs:
OCIContainer plugin is already merged

PR #224, opened 2020-09-02
METROL-103: Updates Packager

MFransen69:
Syncs Packager with the latest updates from ThunderNanoServices


pwielders (changes requested):
This is was changed delibratly to allow for Comcast to use their own provisioning and t not break the metrological way of working. So I think you can drop this pull-request

pwielders:
But we do need the EXTERNAL.... so the CMakefiles.txt changes should not be pushed, the EXTERNAL should

MFransen69:
> But we do need the EXTERNAL.... so the CMakefiles.txt changes should not be pushed, the EXTERNAL should

Done

pwielders (approved):

PR #225, opened 2020-09-02
METROL-103: Updates LocationSync

MFransen69:
Syncs LocationSync with the latest updates from ThunderNanoServices

### DIFF
--- a/LocationSync/LocationSync.config
+++ b/LocationSync/LocationSync.config
@@ -1,5 +1,7 @@
 set (autostart true)
 
+set (preconditions Network)
+
 map()
     kv(interval 10)
     kv(retries 20)
--- a/LocationSync/LocationService.h
+++ b/LocationSync/LocationService.h
@@ -28,7 +28,7 @@ namespace Plugin {
 
     struct IGeography;
 
-    class EXTERNAL LocationService
+    class LocationService
         : public PluginHost::ISubSystem::ILocation,
           public PluginHost::ISubSystem::IInternet,
           public Web::WebLinkType<Core::SocketStream, Web::Response, Web::Request, Core::ProxyPoolType<Web::Response>&> {


LocationSync/LocationSync.config
```diff
@@ -1,5 +1,7 @@
 set (autostart true)
 
+set (preconditions Network)
```
anand-ky, line 3: This precondition was removed because we currently don't have any plugin that satisfies this subsystem. We don't have NetworkControl in RDKServices. We would have to find a plugin in RDKServices (maybe Network) to satisfy this subsystem before adding the precondition.

pwielders: Thunder is a fully decoupled/modular system and is not bound to any specific implementations. 
The subsystems help in avoiding inter plugin dependencies. A plugin that states preconditions, only states these because it is really in need of these subsystems to be available.
The fact that you do *NOT* run the network plugin does not mean that the location sync nolonger requires a network :-) It is mandatory for this plugin to have network connectivity (it tries to reach something on the network)
If you add the Network plugin, the Network plugin will "register" that the subsystem Network is now nolonger under control of Thunder and that Thunder should not automagically set it. Only now the NetworkSystem requires the Network plugin.
So this is a very good fix, unless the LocationSync is now operating without network connectivity :-)

anand-ky: I understand that LocationSync needs network connectivity. Can this Network subsystem be satisfied within thunder instead of a plugin? The current issue we have with this precondition is that although there is network connectivity on the device, LocationSync plugin doesn't activate because the precondition has not been satisfied (no one called subSystem->Set(PluginHost::ISubSystem::NETWORK, nullptr)).

pwielders: Yes, as explained, if you do not configure Thunder that there is an external entity setting the NETWORK subsystem, thunder will set this susbsytem prior to starting any plugin.
This configuration is found in the config -> Controller -> SubSystems, that should *NOT* list the NETWOK susbsystem. If it is not listed, he NETWORK susbsystem is enbaled at startup (you can slo see this in the log if you run thunder in the forground.

pwielders (approved):

LocationSync/LocationSync.config
anand-ky: ok, I see. sounds good, thanks Pierre. We will check and remove Network subsystem from our thunder config.json.

PR #228, opened 2020-09-03
METROL-108: SecurityAgent: document updates based on the ACL config settings

HaseenaSainul:


pwielders (dismissed):

SecurityAgent/AccessControlList.h
```diff
@@ -237,7 +237,7 @@ namespace Plugin {
             JSONACL& operator=(const JSONACL&) = delete;
             JSONACL()
             {
-                Add(_T("assign"), &Groups);
+                Add(_T("groups"), &Groups);
```
anand-ky, line 240: This will break the existing thunder security permissions acl that has "assign" defined. Can we retain "assign"?

pwielders: @HaseenaSainul I would assume this change came from comcast? we are not actively using this plugin yet so for me the change would be oke.
If comcast is using this implementation already, I suggest, for backward compatibility, leave in both fieldnames both pointing to the same variable.

HaseenaSainul: Thanks @pwielders for the suggestion. I updated the code to point both fields, so it will work if we select either one

pwielders (approved):

SecurityAgent/AccessControlList.h
anand-ky: thanks Haseena & Pierre.

anand-ky (approved):
looks good.

PR #239, opened 2020-09-08
DELIA-43141 : ClassCastException on getCecAddresses call during HdmiCecManager initialization

sborushevsky:
Changed output of getCECAddresses to match legacy service.
Used fixed copy of Core::FromString, which doesn't add exrta zero to the buffer after base64 decoding.
Fixed ClassCastException for PhysicalAddress in cecAddressesChanged event.


anand-ky (approved):
looks good

bobseamon (approved):

PR #248, opened 2020-09-10
AMLOGIC-765: reduce locking of communication thread

emutavchi:


goruklu (approved):

pwielders (commented):
Why is this change required? Is it resolving an issue?
I think that the added overhead of the new/delete inflict more of a performance penalty than the lock, since I think it is safe to assume that the lock is available 909% of the time. The new/delete has to be done 100% of the time, also the new/delete will lead to memory fragmentation, so although technically correct I wonder if this is a usefull change?
On a side note, the new also uses locking although the level of locking might be different, it still has to lock :-)

emutavchi:
@pwielders the problem is with _adminLock which is held when sending notifications (visibility change, url change, etc). I see deadlock between JSONRPC workers trying to set properties in WPEFramework(all workes are gitting blocked in WebKitBrowser in DCOM IPC channel) and WPEProcess trying to send a notification (UI thread holds _adminLock and also blocked DCOM IPC,  communication worker is waiting for _adminLock to be released).  WPEFramework is unresponsive for ~20 seconds until DCOM fails (times out). 

The idea is to remove _adminLock from the communication thread when it is not required.

pwielders:
Ahh that makes sense, but than I still prefer to create an additional Critical section to avoid the expensive new/delete. We have seen this issue before and we just gave the notification/register/unregister their own critical section.
I think we can say that the reuse of this single Critical section for the data protection and the notification protection is a bit of dutch frugality:-)

emutavchi:
Reopening because we still need the changes to be merged

pwielders:
Sorry agree, we need a fix but 4 uur == bier uur, so must have clicked the wrong button :-(

emutavchi:
@pwielders  so, OK to merge? 

PS: Geniet van je biertje!

pwielders (approved):
Thanks ! Good beer, belgian Palm :-)
Yes good to merge. We will do a test with an additional critical section and test the performance on the master (which has been offered for you to review later on) to see what is faster. But understand the issue and this is a good fix.

PR #251, opened 2020-09-11
Modified DeviceIdentification plugin to fetch & build brcm specific code from external repository

neeraj-deshpande:


MFransen69 (dismissed):

MFransen69 (changes requested):
Hi Neeraj, think I was not clear in my last request, sorry. It is not a special flag to override when to take the sources from Gerrit, the idea was that you could specify for a component to not take the sources from Gerrit (or RDK Central) but you could specify a folder to get the sources from (so you can have them local on a disk or something similar instead)

MFransen69 (dismissed):

woutermeek (approved):

gladish (approved):

PR #265, opened 2020-09-14
DELIA-43141 : ClassCastException on getCecAddresses call during HdmiCecManager initialization

sborushevsky:
Changed output of getCECAddresses to match legacy service.
Used fixed copy of Core::FromString, which doesn't add exrta zero to the buffer after base64 decoding.
Fixed ClassCastException for PhysicalAddress in cecAddressesChanged event.


anand-ky:
looks good. Pls also cherry-pick to 2010/sprint.

PR #272, opened 2020-09-15
RDK-27957 Bring up TextToSpeech plugin code from master to sprint/2009

nrajan002c:

### DIFF
--- a/TextToSpeech/CMakeLists.txt
+++ b/TextToSpeech/CMakeLists.txt
@@ -1,0 +1,32 @@
+set(PLUGIN_NAME TextToSpeech)
+set(MODULE_NAME ${NAMESPACE}${PLUGIN_NAME})
+
+find_package(${NAMESPACE}Plugins REQUIRED)
+
+add_subdirectory(test)
+
+add_library(${MODULE_NAME} SHARED
+        TextToSpeech.cpp
+        Module.cpp
+        impl/TTSManager.cpp
+        impl/TTSSpeaker.cpp
+        impl/logger.cpp
+        )
+set_target_properties(${MODULE_NAME} PROPERTIES
+        CXX_STANDARD 11
+        CXX_STANDARD_REQUIRED YES)
+
+list(APPEND CMAKE_MODULE_PATH
+        "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
+
+find_package(GSTREAMER REQUIRED)
+
+find_package(Curl)
+
+target_include_directories(${MODULE_NAME} PRIVATE ../helpers ${GSTREAMER_INCLUDES})
+target_link_libraries(${MODULE_NAME} PRIVATE ${NAMESPACE}Plugins::${NAMESPACE}Plugins ${CURL_LIBRARY} ${GSTREAMER_LIBRARIES})
+
+install(TARGETS ${MODULE_NAME}
+        DESTINATION lib/${STORAGE_DIRECTORY}/plugins)
+
+write_config( TextToSpeech )
--- a/TextToSpeech/TextToSpeech.json
+++ b/TextToSpeech/TextToSpeech.json
@@ -1,0 +1,9 @@
+{
+ "locator":"libWPEFrameworkTextToSpeech.so",
+ "classname":"TextToSpeech",
+ "precondition":[
+  "Platform"
+ ],
+ "callsign":"org.rdk.TextToSpeech",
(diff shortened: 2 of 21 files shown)


varumugam123 (approved):

TextToSpeech/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+        impl/logger.cpp
+        )
+set_target_properties(${MODULE_NAME} PROPERTIES
+        CXX_STANDARD 11
+        CXX_STANDARD_REQUIRED YES)
+
+list(APPEND CMAKE_MODULE_PATH
+        "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
+
+find_program(PKG_CONFIG_EXECUTABLE NAMES pkg-config DOC "pkg-config executable")
```
emutavchi, line 22: Why not create FindGstreamer.cmake or similar and use find_package? note, there is PkgConfig find package already.

TextToSpeech/TextToSpeech.json
```diff
@@ -0,0 +1,9 @@
+{
```
emutavchi, line 1: Isn't this auto-generated? probably not needed to be checked in.

TextToSpeech/impl/TTSManager.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+TTSManager::TTSManager(TTSEventCallback *callback) :
+    m_callback(callback),
+    m_speaker(NULL),
+    m_ttsEnabled(false) {
+
+    TTSLOG_TRACE("TTSManager::TTSManager");
+
+    // Load configuration from file
+    loadConfigurationsFromFile(TTS_CONFIGURATION_FILE);
```
emutavchi, line 125: This should not be allowed in PROD builds.

TextToSpeech/impl/TTSSpeaker.cpp
```diff
@@ -0,0 +1,847 @@
...
+    }
+
+    m_source = gst_element_factory_make("souphttpsrc", NULL);
+
+    // create soc specific elements
+#if defined(BCM_NEXUS)
+    GstElement *decodebin = NULL;
+    decodebin = gst_element_factory_make("brcmmp3decoder", NULL);
+    m_audioSink = gst_element_factory_make("brcmpcmsink", NULL);
+#endif
```
emutavchi, line 371: I think this is against VAP requirements, i.e. we shouldn't use SoC specific plugins directly. has to be addressed sooner or later

TextToSpeech/impl/TTSManager.cpp
varumugam123: Hi Eugene, could you let us know the concern?

TextToSpeech/impl/TTSSpeaker.cpp
varumugam123: Hi Eugene, I have a doubt. Should rdk services as well follow VAP?

TextToSpeech/impl/TTSManager.cpp
emutavchi: This is the general rule. in nutshell, any dev configuration left on the box should not have any impact on prod image

TextToSpeech/impl/TTSSpeaker.cpp
emutavchi: Good question, I think it makes sense for services as well. I mean it should just work in a generic way.

TextToSpeech/CMakeLists.txt
nrajan002c: Modified the code to use FindGstreamer.cmake.

TextToSpeech/TextToSpeech.json
nrajan002c: Removed the TextToSpeech.json.

TextToSpeech/impl/TTSManager.cpp
nrajan002c: We are planning to add the default configuration as a config setting in the ticket RDK-20988.

TextToSpeech/impl/TTSSpeaker.cpp
nrajan002c: We will remove the explicit SOC specific plugins as part of RDK-20988.

emutavchi (approved):

PR #307, opened 2020-09-23
hdcp2_2RxInfo is not available in nexus v14.4-r0 (MOT7425-11257)

vijs:
Removed reference to hdcp2_2RxInfo as it is not available in one of our platforms (XG1V1)


pwielders (changes requested):
You only removed the parameter (whihc is good) but also remove the element in the formatting lin (%s), probably this line https://github.com/rdkcentral/rdkservices/blob/613de53a5bf7d3aa8871380059c47e3dd35f5519/DisplayInfo/Nexus/PlatformImplementation.cpp#L464

PR #314, opened 2020-09-24
Added NetworkService Api  to query ipaddress based on the ipaddress family.

bpunnuru:
Signed-off-by: Balaji Punnuru <balaji_punnuru@cable.comcast.com>


bobseamon (changes requested):
Can you rename getSTBip_family to getIPFamily, or getSTBIPFamily?  Would prefer not to use the underscore as part of naming conventions.

bobseamon (approved):

PR #363, opened 2020-10-14
store active plugins and log app launch time

madanagopalt:


mfiess:
This has been merged to sprint/2010.

PR #384, opened 2020-10-19
RDK-29718 : Implemented ScreenCapture for Realtek platform.

sborushevsky:
On Realtek there is vncframebuffer lib, which can be used for getting screnshots.


anand-ky (approved):
looks good

PR #396, opened 2020-10-23
DELIA-46649 : Allow overwriting 5 digit code with irdb - RAMS thunder…

jschmi201:
… plugin.

Reason for change: Couldn't do autoir if 5 digit code was loaded.
Test Procedure: Verify irdb can overwrite 5 digit code.
Risks: None.
Signed-off-by: jschmidt <Jeff_Schmidt@cable.comcast>


bobseamon (approved):
Looks good.

PR #408, opened 2020-10-27
Make TextToService outofprocess

varumugam123:


varumugam123 (commented):
It would be fail safe if a default (non-empty string) value is passed to std::stod()

ramasat (approved):

ramasat (approved):
Changes look good to me.

PR #410, opened 2020-10-27
RDK-29860 : Routing Control Changes

kselva1983:
RDK-29860 : Hdmi CEC Sink Routing control changes into 2010 sprint branch


anand-ky (approved):
looks good

PR #414, opened 2020-10-28
Configure TextToSpeech service to start out-of-process

varumugam123:
This config change was missed in https://github.com/rdkcentral/rdkservices/pull/408


ramasat (approved):
Approved.

PR #417, opened 2020-10-29
DELIA-46698: Introduce JGraphicsProperties

vinod-jacob:
Reason for change: JGraphicsProperties will
register the totalgpuram and freegpuram properties, so that
these 2 properties will become available to DisplayInfo plugin.
Updated readme file also
Test Procedure: use curl commands, check compilation
Risks: None


vinod-jacob:
@pwielders,
Changes in DisplayInfo.h and DisplayInfo.cpp to bring the JGraphicsProperties related functionalities (totalgpuram & freegpuram) into rdkservices repo. I've taken these 2 files entirely from Thundernanoservices. Could you please review this so that I can proceed for merge?

DisplayInfo/DisplayInfo.cpp
```diff
@@ -146,26 +149,34 @@ namespace Plugin {
...
         }
+        value = 0;
         if (_connectionProperties->Height(value) == Core::ERROR_NONE) {
             displayInfo.Height = value;
         }
 
-        if (static_cast<const Exchange::IConnectionProperties*>(_connectionProperties)->HDCPProtection(hdcpType) == Core::ERROR_NONE) {
-            displayInfo.Hdcpprotection = static_cast<JsonData::DisplayInfo::DisplayinfoData::HdcpprotectionType>(hdcpType);
+        Exchange::IConnectionProperties::HDCPProtectionType hdcpProtection(Exchange::IConnectionProperties::HDCPProtectionType::HDCP_Unencrypted);
+        if ((const_cast<const Exchange::IConnectionProperties*>(_connectionProperties))->HDCPProtection(hdcpProtection) == Core::ERROR_NONE) {
```
pwielders, line 171: const_cast by definition is a scry thing. I think you are now invoking the wrong method here. You probably now invoke the SET while you where looking for a get, I assume.
Looking at the sugnature, https://github.com/rdkcentral/ThunderInterfaces/blob/master/interfaces/IDisplayInfo.h#L111 the method is const so in a const method, const objects will be used (const Exchange::IConnectionProperties* _connectionProperties) which should only give you access to the getter, without the const cast.
If you get a compiler error/warning, would love to see it :-)
Bottomline, I think the const_cast should go here

pwielders (changes requested):
Looks good, except for 1 const_cast.

DisplayInfo/DisplayInfo.cpp
vinod-jacob: Removed the const_cast. Compiler didn't throw any errors or warnings. I think we're good to go

pwielders (approved):

PR #87, opened 2020-06-29
RDK-28664 : Persistent Storage Thunder API

npoltorapavlo:
Reason for change: new thunder plugin
Test Procedure: test functionality specified in docs
Risks: Low
Signed-off-by: Nikita Poltorapavlo <npoltorapavlo@productengine.com>

PR #425, opened 2020-10-30
bring persistent plugin to sprint/2010

npoltorapavlo:


NOTICE
```diff
@@ -9,3 +9,6 @@ within this component.
 
 Copyright (C) 2012 Raphael Kubo da Costa <rakuco@webkit.org>
 Licensed under the BSD-2 license
+
```
anand-ky, line 12: Do we need this license entry? What change in PersistentStore is using pxCore?

anand-ky (changes requested):

NOTICE
npoltorapavlo: Requested by RDK CMF Compliance Team in #87

anand-ky: ok sounds good.

anand-ky (approved):

PR #448, opened 2020-11-04
XRE-16568: [DisplayInfo] Converge all Soc Impl to use DeviceSettings

vinod-jacob:
Reason for change: Changes required to make all RDK ports to make use of DisplayInfo/DeviceSettings/PlatformImplementation.cpp.
1. Added ds implementations for APIs where suitable ds library functions were available (isaudiopassthrough, verticalfreq, connected)
2. Created header file SoC_Abstraction.h for abstracting implementations of APIs(height, width, totalgpuram, freegpuram)
3. Added implementation for Broadcom through abstraction file Broadcom/SoC_abstraction.cpp (XRE-16583)
4.  Added implementation for Realtek through abstraction file Realtek/SoC_abstraction.cpp (XIONE-1313)

Test Procedure: Check compilation, use curl commands
Risks: None


vinod-jacob:
**Jenkins custom verification**: https://gerrit.teamccp.com/#/c/451773/

**curl command verification:**
`curl -H "Authorization: Bearer `WPEFrameworkSecurityUtility | cut -d '"' -f 4`" --header "Content-Type: application/json"   --request POST   --data  '{"jsonrpc":"2.0","id":"3","method": "DisplayInfo.1.isaudiopassthrough"}' http://127.0.0.1:9998/jsonrpc
{"jsonrpc":"2.0","id":3,"result":false}

curl -H "Authorization: Bearer `WPEFrameworkSecurityUtility | cut -d '"' -f 4`" --header "Content-Type: application/json"   --request POST   --data  '{"jsonrpc":"2.0","id":"3","method": "DisplayInfo.1.verticalfreq"}' http://127.0.0.1:9998/jsonrpc
{"jsonrpc":"2.0","id":3,"result":59}

curl -H "Authorization: Bearer `WPEFrameworkSecurityUtility | cut -d '"' -f 4`" --header "Content-Type: application/json"   --request POST   --data  '{"jsonrpc":"2.0","id":"3","method": "DisplayInfo.1.height"}' http://127.0.0.1:9998/jsonrpc
{"jsonrpc":"2.0","id":3,"result":720}

curl -H "Authorization: Bearer `WPEFrameworkSecurityUtility | cut -d '"' -f 4`" --header "Content-Type: application/json"   --request POST   --data  '{"jsonrpc":"2.0","id":"3","method": "DisplayInfo.1.width"}' http://127.0.0.1:9998/jsonrpc
{"jsonrpc":"2.0","id":3,"result":1280}

curl -H "Authorization: Bearer `WPEFrameworkSecurityUtility | cut -d '"' -f 4`" --header "Content-Type: application/json"   --request POST   --data  '{"jsonrpc":"2.0","id":"3","method": "DisplayInfo.1.connected"}' http://127.0.0.1:9998/jsonrpc
{"jsonrpc":"2.0","id":3,"result":false}`

vinod-jacob:
@pwielders,
Could you have a look at this PR and provide me with your review comments.
These changes primarily focus on adding implementation to the DisplayInfo properties totalgpuram, freegpuram, height and width. Since these APIs are not readily available in the ds library, they have been implemented using **procfs in BCM** platforms, and **kms & procfs on Realtek** platforms.

pwielders (commented):
General remark: 
The design is not uniform. 
1) Some SOC specific properties are delivered by a the DeviceSettings manager
2) Some SOC specific properties are delivered by this plugin (what you add now)
This leads to confusion during maintenance and integration, what goes where, what if some  properties will work (not over IARM) and others will not (IARM is down)

I think, with the DeviceSettings you have chosen to delegate the SOC specifics to the DeviceSettings Manager over IARM, I would stick to this concept and extend the interface over IARM. Or move it all into this plugin, which design wise I would prefer to avoid additional dependencies and hops.
But bottomline: Keep the same paradigm.

Also to avoid flooding of the system with LOGS/TRACES, the Thudner framework offers the TRACE macro which can be turned on runtime, please use it in stead of the Compile time macros LOGINFO/LOGERR.

Exception handling is a severe penalty for embedded systems. I would strongly discourage it definitely for "just" checking if the conversion succeeds. Thunder offers (Core::NumberType<>) conversion methods that do not throw.

As these changes do not influence any deployments currently in the field (these deplyment use the other platform implementations) I would like to leave the design/architecture concerns up to the responsible architects/designers within Comcast to decide whether these concerns need attention or not

vinod-jacob:
Thank you for your feedback @pwielders 

@bobseamon, to reply to some of Pierre's comments:
I agree that the current architecture of DisplayInfo/DeviceSettings is non-uniform, and the reasons why we opted for this approach despite this non-uniformity are as follows

1. These changes are meant to serve as temporary solutions. We resorted to delegating the implementations of 4 graphics related APIs to respective Soc_abstractions because there aren't suitable DS lib APIs available presently. We're working on getting the required API into DS library as part of XRE-16610 currently and once this is completed, we'll be removing the Soc_abstractions.

2. Maintaining entirely separate backends for DisplayInfo (DisplayInfo/Nexus, DisplayInfo/Amlogic & an incoming DisplayInfo/Realtek) was becoming increasingly tedious from a Comcast perspective. This is especially true during times when a new function is added to  IDisplayInfo.h interface. During such an event, we would need to update each backend individually. This would result in the DisplayInfo plugin becoming unavailable on Xi6, Platco, XiOne for different lengths of time. This would be an obstacle for app development teams as well as maintainers of DisplayInfo, as the progress on each platform would need to be tracked separately.

3. While we work to bring the latest version of DisplayInfo into stable2, we'll be having to cherry-pick from sprint to sprint. Having all Comcast platforms converged into using DisplayInfo/DeviceSettings backend makes this cherry-picking task trivial. If on the other hand, we don't converge, individual maintainers of DisplayInfo/Amlogic, DisplayInfo/Nexus & DisplayInfo/Realtek would need to cherry-pick their changes themselves. This would again result in DisplayInfo plugin becoming unavailable on Xi6, Platco & XiOne for different durations, making life difficult for app dev teams

4. As part of XIONE-1313, Realtek was planning on introducing DisplayInfo/Realtek. The implementation was however mostly a copy of DisplayInfo/DeviceSettings. As there is no good reason to maintain 2 separate backends, both relying on DS lib, we decided to converge both. Discussions on this can be read -> https://ccp.sys.comcast.net/browse/XIONE-1313?focusedCommentId=10113293&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-10113293

5. During 2009 sprint, new APIs were added to IDisplayInfo.h Thunder interface. In order to update the DisplayInfo/Amlogic with implementations of these new APIs, they too had chosen to borrow from DeviceSettings. This would have resulted in a non-uniform Amlogic backend, where some of the APIs use amlogic library calls while the remaining ones use DS lib calls (https://gerrit.teamccp.com/#/c/451334/4/meta-platco/recipes-extended/wpe-framework/rdkservices/Displayinfo-Amlogic.patch). By getting Amlogic also to eventually converge to the architecture of this PR, at least we'll make sure that apart from 4 APIs, the remaining ones are being serviced from the same  DisplayInfo/DeviceSettings/PlatformImplementation.cpp file.

pwielders:
An architectural correction:
The statement on point 2: _"This is especially true during times when a new function is added to IDisplayInfo.h interface. During such an event, we would need to update each backend individually."_ This is not necessarily true, The COMRPC mechanism deployed in this plugin allows for interface querying. As we added, for example the HDR functionality to this plugin, the HDR interface, that was added, was done using interface navigation (QueryInterface). Since the RPI does not support HDR, we just did not implement the HDR functionality on this backend. So if the hardware does not uspport, do not fake it, just do not implement it, it is *not* mandatory.
Modularity and extensibility is key to Thunder plugins. The core libraries help you to achieve this goal.

vinod-jacob:
@pwielders , I agree with what you've mentioned.
I was, however, referring to instances where new functions are added to interfaces.
For example when WidthInCentimeters & HeightInCentimeters was added to IConnectionProperties, PlatformImplementation.cpp wasn't updated in time. This resulted in compilation errors when @vijs upgraded the git revision of Thunder repo.  As a result, he had to disable building DisplaInfo plugin on all platforms. In such a situation, if Comcast is maintaining different SoC backends (AmLogic, Nexus, Realtek), each of those would have to be updated separately by its maintainers through separate PRs. This would result in DisplayInfo being unavailable on Xi6, XiOne, Platco for different durations and thus would be a headache for app-dev teams. Therefore, having all Comcast devices running off of DeviceSettings/PlatformImplementation.cpp would facilitate matters for everyone since merging just one PR would get all platforms working again.

pwielders:
@vinod-jacob that was not a technical issue, that was a project management issue!
We really need to do design sessions before either side starts enhancing plugins. In such design sessions we could evaluate the best approach and options where we could have opted for an additional interface, or dummy implementations and than no other platform needed an update.

vinod-jacob:
@pwielders , I agree. This is something we need to work on moving forward. In fact, @vijs and I did briefly discuss decoupling the plugin from the Thunder interfaces.

bobseamon:
@pwielders - Are you satisfied with where this is right now?  If so, we can merge.

pwielders:
@bobseamon since it is contained in the platform layer (IARM/DeviceSettings) it does not impact any of our deployments so I am good to go!

pwielders:
Sorry wrong button !

vijs (approved):

PR #452, opened 2020-11-05
RDK-29867 : uploadLogs method

npoltorapavlo:
Reason for change: Add a new method to the System
RDK Service: uploadLogs.
Test Procedure: Call uploadLogs with or without "url".
Risks: None


anand-ky (changes requested):
Please add the new API to version 2 (systemVersion_2.Register)

anand-ky (approved):

PR #473, opened 2020-11-16
RDK-29274:Thunder APIs for AutoReboot

livinsunny519:
Reason for change: Added new method for
fwPendingReboot & fwDelayReboot.
'{"jsonrpc":"2.0","id":"3","method":"org.rdk.System.2.fwPendingReboot","params":{}}' 
'{"jsonrpc":"2.0","id":"3","method":"org.rdk.System.2.fwDelayReboot","params":{"fwDelayReboot":int seconds}}'
Test Procedure: Refer to Jira Ticket.

### DIFF
--- a/SystemServices/SystemServices.cpp
+++ b/SystemServices/SystemServices.cpp
@@ -40,6 +40,7 @@
 #include "SystemServices.h"
 #include "StateObserverHelper.h"
 #include "utils.h"
+#include "rfcapi.h"
 
 #if defined(USE_IARMBUS) || defined(USE_IARM_BUS)
 #include "libIARM.h"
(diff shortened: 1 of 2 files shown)


livinsunny519:
@bobseamon @vijs could you please review

bobseamon (changes requested):
These changes add new methods to the system service.  As such, the system service should be incremented to version 2.  Once that change is made, documentation needs to be updated at https://wiki.rdkcentral.com/pages/viewpage.action?pageId=105516014 .  If you don't have access, I can make the doc changes, but they will need to be documented before the change is merged.

livinsunny519:
updated and rebased the code @bobseamon

livinsunny519:
addressed in https://github.com/rdkcentral/rdkservices/pull/552.

PR #479, opened 2020-11-17
Move gst_init() to GStreame thread to reduce startup time

varumugam123:


ramasat (approved):
Changes looks good to me, approved.

ramasat (approved):
Approved!

PR #489, opened 2020-11-18
RDKTV-297 RDKTV-422 - Do IARM Init before calling BTMgrInit

anand-ky:
Reason for change:
Do IARM Init within Bluetooth thunder plugin before calling BTMgrInit. This will avoid
IARM Init within BTMgr
Test Procedure:
Refer to ticket for details
Risks: Low


anand-ky:
Changes not needed.

PR #492, opened 2020-11-18
[DeviceIdentification] Removed const from IIdentifier interface, added interfaces aggregates

lukiwa:
It is necessary in order for this plugin to work well with its ClientLibrary.


pwielders (changes requested):
@lukiwa I do not think we want or need the PluginHost::ISubSystem::IIdentifier on the external interfaces, so the removal of the const is not needed. Lets have a skype discussion on it.

lukiwa:
Closed as this is no longer relevant since the implementation of client lib has changed :)

PR #512, opened 2020-11-25
DELIA-47408 : Optimizing in response for method:

maksimsav:
org.rdk.Network.getNamedEndpoints

### DIFF
--- a/Network/Network.cpp
+++ b/Network/Network.cpp
@@ -409,10 +409,7 @@ namespace WPEFramework
         uint32_t Network::getNamedEndpoints(const JsonObject& parameters, JsonObject& response)
         {
             JsonArray namedEndpoints;
-            JsonObject endpoint;
-
-            endpoint["endpoint"] = "CMTS";
-            namedEndpoints.Add(endpoint);
+            namedEndpoints.Add("CMTS");
 
             response["endpoints"] = namedEndpoints;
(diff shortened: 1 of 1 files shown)


anand-ky (changes requested):
Why do we need to change this request signature? This will break existing clients using this API.

maksimsav:
1. Based on https://wiki.rdkcentral.com/display/RDK/Network#Network-getNamedEndpoints(v1)
we expect to get following result:

Request: {"jsonrpc":"2.0", "id":3, "method":"org.rdk.Network.1.getNamedEndpoints", "params":{}}
Response: {"jsonrpc":"2.0", "id":3, "result":{"endpoints":["CMTS"]}}

But without this change it returns:
Response: {"jsonrpc":"2.0", "id":3, "result":{"endpoints":[{"endpoint","CMTS"}]}}

2. By default, Networking plugin still disabled in servicemanager and the legacy ping service is used instead. I'm not sure that we can break something. But implementation in plugin is incorrect and don't correspond to Thunder, nor SM specs.

anand-ky (approved):
Looks good. This change is to match the implementation for getNamedEndPoints with the spec - https://wiki.rdkcentral.com/display/RDK/Network#Network-getNamedEndpoints(v1)

PR #554, opened 2020-12-10
webkitbrowser: always rank thunder decryptor no matter what

calvaris:
I found the variable is not proper set and it does not make sense that it is not always set. Setting it to `true` in all cases.


woutermeek (approved):

neeraj-deshpande (approved):

vijs:
It appears like this change set is merged in PR #709.
If anything is missing in main branch, please resubmit it to the latest sprint branch.

PR #576, opened 2020-12-14
Fix RDKShell build: missing IARMBus package find

sverkoye:
* CMakeLists.txt adds IARMBUS_INCLUDE_DIRS and IARMBUS_LIBRARIES
  so it should also try to find the IARMBus package first
* this build error does not occur with all builds configs.
  It probably depends on which other pluginx are also enabled.
  I noticed this on RPI reference image build.
* needed because ../helpers/utils.cpp always pulls in IARMBus


vijs:
I am closing this outdated PR. If this change set needs to be merged, please resubmit it to the latest sprint branch.

PR #583, opened 2020-12-14
Support async http requests

gladish:


gladish:
going to re-do on newer branch

PR #590, opened 2020-12-15
[DeviceInfo] Extended plugin to gather information about the STB from the config file.

lukiwa:


DeviceInfo/DeviceInfoImplementation.cpp
```diff
@@ -0,0 +1,123 @@
+#include "DeviceInfoImplementation.h"
```
pwielders, line 1: As this is part of the interface definition, can we move the mapping to text also to the definition in the ThunderInterface/definitions section ?
That way all users of this interface will have the same textual mapping. Avoids confusion :-)

DeviceInfo/DeviceInfoJsonRpc.cpp
```diff
@@ -33,13 +33,15 @@ namespace Plugin {
         Property<SysteminfoData>(_T("systeminfo"), &DeviceInfo::get_systeminfo, nullptr, this);
         Property<Core::JSON::ArrayType<AddressesData>>(_T("addresses"), &DeviceInfo::get_addresses, nullptr, this);
         Property<SocketinfoData>(_T("socketinfo"), &DeviceInfo::get_socketinfo, nullptr, this);
+        Property<StbcapabilitiesData>(_T("stbcapabilities"), &DeviceInfo::get_stbcapabilities, nullptr, this);
```
pwielders, line 36: Lets just call this capabilities, avoid alliteration :-) The fact that it is on the plugin Device, already explains that these are "device" -> STB capabilities and not Display capabilities since these can be found on the DisplayInfo :-)

DeviceInfo/doc/DeviceInfoPlugin.md
```diff
@@ -73,21 +73,92 @@ The table below lists configuration options of the plugin.
...
 # Properties
 
-The following properties are provided by the DeviceInfo plugin:
+The following properties are provided by the DeviceInfo interface:
 
 DeviceInfo interface properties:
 
 | Property | Description |
 | :-------- | :-------- |
+| [stbcapabilities](#property.stbcapabilities) <sup>RO</sup> | Capabilities of the STB |
```
pwielders, line 97: If the proeprty is stbcapabilities, could we change it to capabilities, method already reflect that tjis is on the DeviceInfo and not on the DisplayInfo

pwielders (changes requested):

lukiwa:
Created proper PR here: https://github.com/WebPlatformForEmbedded/ThunderNanoServicesRDK/pull/15

PR #605, opened 2020-12-17
Modified WebKitBrowser plugin - GLib API based code to add support fo…

neeraj-deshpande:
…r console log


emutavchi (approved):
2010 sprint is closed, so this should also be cherry-picket to 2101

pwielders (approved):

PR #616, opened 2020-12-22
RDK-30451: set default display for rdkshell surface mode

madanagopalt:


mfiess (commented):
RDKShell's PR 38 needs to be merged first.

mfiess (approved):

PR #618, opened 2020-12-22
RDKTV-96 - Fix sending early onSuspended event

madanagopalt:


RDKShell/RDKShell.cpp
```diff
@@ -198,6 +199,17 @@ namespace WPEFramework {
...
+                        std::map<std::string, PluginStateChangeData*>::iterator pluginStateChangeEntry = gPluginsEventListener.find(service->Callsign());
+                        if (pluginStateChangeEntry != gPluginsEventListener.end())
+                        {
+                            PluginStateChangeData* stateChangeData = pluginStateChangeEntry->second;
+                            if (nullptr != stateChangeData)
+                            {
+                                delete stateChangeData;
+                            }
+                            pluginStateChangeEntry->second = nullptr;
+                            gPluginsEventListener.erase(pluginStateChangeEntry);
```
mfiess, line 211: Is there a risk of multiple threads accessing `gPluginsEventListener` at the same time?  Same concern about `gActivePlugins`.  Maybe it's best to protect with a new mutex to prevent a crash.

mfiess (changes requested):

PR #637, opened 2021-01-04
XRE-16757: Make LocalStorage size configurable

nrajan002c:


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -551,6 +553,7 @@ static GSourceFuncs _handlerIntervention =
             Core::JSON::String CookieStorage;
             Core::JSON::String LocalStorage;
             Core::JSON::Boolean LocalStorageEnabled;
+            Core::JSON::String LocalStorageSize;
```
MFransen69, line 556: Is there a reason why not one of the Core::JSON number types were used instead of a string? That way you would have the advantage of  using the standard parser validation and conversion

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -1631,6 +1634,13 @@ static GSourceFuncs _handlerIntervention =
             g_free(wpeStoragePath);
             WKContextConfigurationSetLocalStorageDirectory(contextConfiguration, storageDirectory);
 
+            if (_config.LocalStorageSize.IsSet() == true && _config.LocalStorageSize.Value().empty() == false) {
+                uint32_t val = std::stoul(_config.LocalStorageSize.Value().c_str());
+                uint32_t gLocalStorageDatabaseQuotaInBytes = val * 1024;
+                TRACE_L1("Configured LocalStorage Quota  %u bytes", gLocalStorageDatabaseQuotaInBytes);
```
MFransen69, line 1640: Although I know there are still more places where TRACE_L1 is used in this file that is deprecated and it would be better to use the TRACE(Trace::Information, construction (also already used here). That way you would have the traces also available in a release build and configurable.

MFransen69 (changes requested):

WebKitBrowser/WebKitImplementation.cpp
nrajan002c: @MFransen69 I modified the type as per your suggestion . Could you please check it.

MFransen69 (approved):

emutavchi (approved):

PR #642, opened 2021-01-04
RDK-29466 : Added getTimeZones method

sborushevsky:
RDK-29466 : Added logging for getTimeZones

### DIFF
--- a/SystemServices/SystemServices.cpp
+++ b/SystemServices/SystemServices.cpp
@@ -62,6 +62,8 @@ using namespace std;
 #define SYSSRV_MAJOR_VERSION 1
 #define SYSSRV_MINOR_VERSION 0
 
+#define ZONEINFO_DIR "/usr/share/zoneinfo"
+
 /**
  * @struct firmwareUpdate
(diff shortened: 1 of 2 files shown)


SystemServices/SystemServices.cpp
```diff
@@ -243,6 +245,8 @@ namespace WPEFramework {
             : AbstractPlugin()
               , m_cacheService(SYSTEM_SERVICE_SETTINGS_FILE)
         {
+            Core::JSONRPC::Handler& systemVersion_2 = JSONRPC::CreateHandler({ 2 }, *this);
```
anand-ky, line 248: Pls use the new registerMethod() API with versioning support and pass in the versions. FYI this is the change done in 2010 sprint - https://github.com/rdkcentral/rdkservices/pull/535/files

anand-ky (changes requested):

SystemServices/SystemServices.cpp
sborushevsky: Used new new registerMethod().

anand-ky (approved):

PR #688, opened 2021-01-12
HISENSE-198 : Fix UsbAccess flaws

npoltorapavlo:
Reason for change: Do not rely on lighttpd document-root,
and don't parse lighttpd config which is unreliable.
Instead, create and use a specific lighttpd alias.
Return error "file exists" for createLink if link exists.
Do not hardcode mount path but find it using system files.
Test Procedure: Test UsbAccess as usually
but with multiple USB drives having single or
multiple partitions.
Risks: Low

### DIFF
--- a/UsbAccess/UsbAccess.cpp
+++ b/UsbAccess/UsbAccess.cpp
@@ -4,6 +4,7 @@
 #include <sys/stat.h>
 #include <dirent.h>
 #include <unistd.h>
+#include <mntent.h>
 
 const short WPEFramework::Plugin::UsbAccess::API_VERSION_NUMBER_MAJOR = 1;
 const short WPEFramework::Plugin::UsbAccess::API_VERSION_NUMBER_MINOR = 0;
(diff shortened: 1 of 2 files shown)


UsbAccess/UsbAccess.cpp
```diff
@@ -4,16 +4,17 @@
...
 const short WPEFramework::Plugin::UsbAccess::API_VERSION_NUMBER_MAJOR = 1;
 const short WPEFramework::Plugin::UsbAccess::API_VERSION_NUMBER_MINOR = 0;
 const string WPEFramework::Plugin::UsbAccess::SERVICE_NAME = "org.rdk.UsbAccess";
 const string WPEFramework::Plugin::UsbAccess::METHOD_GET_FILE_LIST = "getFileList";
 const string WPEFramework::Plugin::UsbAccess::METHOD_CREATE_LINK = "createLink";
 const string WPEFramework::Plugin::UsbAccess::METHOD_CLEAR_LINK = "clearLink";
-const string WPEFramework::Plugin::UsbAccess::USB_MOUNT_PATH = "/run/media/sda"; // platco
-const string WPEFramework::Plugin::UsbAccess::LINK_URL_HTTP = "http://localhost:50050/usbdrive";
-const string WPEFramework::Plugin::UsbAccess::LIGHTTPD_CONF_PATH = "/etc/lighttpd/lighttpd.conf";
+const string WPEFramework::Plugin::UsbAccess::LINK_URL_HTTP = "http://127.0.0.1:50050/usbdrive";
```
bobseamon, line 15: Is there a risk using 127.0.0.1 instead of localhost?

npoltorapavlo: reverted to "localhost"

bobseamon (approved):

PR #696, opened 2021-01-13
HISENSE-198 : Fix UsbAccess flaws

npoltorapavlo:
Reason for change: Do not rely on lighttpd document-root,
and don't parse lighttpd config which is unreliable.
Instead, create and use a specific lighttpd alias.
Return error "file exists" for createLink if link exists.
Do not hardcode mount path but find it using system files.
Test Procedure: Test UsbAccess as usually
but with multiple USB drives having single or
multiple partitions.
Risks: Low

### DIFF
--- a/UsbAccess/UsbAccess.cpp
+++ b/UsbAccess/UsbAccess.cpp
@@ -4,6 +4,7 @@
 #include <sys/stat.h>
 #include <dirent.h>
 #include <unistd.h>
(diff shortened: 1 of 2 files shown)


npoltorapavlo:
@bobseamon, cherry-picked #688 to 2010

bobseamon (approved):

PR #774, opened 2021-02-01
RDKTV-747 : Hidden passphrase in the log.

sborushevsky:


WifiManager/impl/WifiManagerConnect.cpp
```diff
@@ -29,7 +29,16 @@ using namespace WPEFramework::Plugin;
 
 uint32_t WifiManagerConnect::connect(const JsonObject &parameters, JsonObject &response)
 {
-    LOGINFOMETHOD();
+    if (params.HasLabel("passphrase"))
```
anand-ky, line 32: Did u mean parameters here instead of params?We need to pass the actual passphrase to the underlying layer and just suppress the logging. Pls verify that it builds and works.

WifiManager/WifiManager.cpp
```diff
@@ -182,7 +182,16 @@ namespace WPEFramework
 
         uint32_t WifiManager::connect(const JsonObject &parameters, JsonObject &response)
         {
-            LOGINFOMETHOD();
+            if (params.HasLabel("passphrase"))
```
anand-ky, line 185: Did u mean parameters here instead of params? We need to pass the actual passphrase to the underlying layer and just suppress the logging. Pls verify that it builds and works.

anand-ky (changes requested):

WifiManager/impl/WifiManagerConnect.cpp
sborushevsky: Forgot to add one line.
Added commit to fix this.

WifiManager/WifiManager.cpp
sborushevsky: Forgot to add one line.
Added commit to fix this.

anand-ky (approved):

PR #845, opened 2021-02-18
RDKTV-1785:[Platco]Audio ports getting enabled even in boot to Standb…

tpaul627:
…y mode

Reason for change: Even though pwrMgr disables the audio port during boot to standby,
Audio ports were getting enabled from DisplaySetting thunder plugin activation.
Audioports will not get enabled from DisplaySetting plugin activation but from power state
transition to ON.
Test Procedure: Verify all audio ports work after powerstate transition to ON.
Risks: Low
Signed-off-by: Tony Paul <Tony_Paul@comcast.com>

### DIFF
--- a/DisplaySettings/DisplaySettings.cpp
+++ b/DisplaySettings/DisplaySettings.cpp
@@ -289,7 +289,31 @@ namespace WPEFramework {
             LOGINFO("Starting the timer");
             m_timer.start(RECONNECTION_TIME_IN_MILLISECONDS);
 
-            InitAudioPorts();
+            JsonObject joGetParams;
+            JsonObject joGetResult;
+            joGetParams["params"] = JsonObject();
+            std::string getPowerStateInvoke = "org.rdk.System.1.getPowerState";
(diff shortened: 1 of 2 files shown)


DisplaySettings/DisplaySettings.cpp
```diff
@@ -289,7 +289,31 @@ namespace WPEFramework {
             LOGINFO("Starting the timer");
             m_timer.start(RECONNECTION_TIME_IN_MILLISECONDS);
 
-            InitAudioPorts();
+            JsonObject joGetParams;
+            JsonObject joGetResult;
+            joGetParams["params"] = JsonObject();
+            std::string getPowerStateInvoke = "org.rdk.System.1.getPowerState";
+            auto thunderController = getThunderControllerClient();
```
anand-ky, line 296: Rename getThunderControllerClient to something like getSystemPlugin. This would be similar to getHdmiCecSinkPlugin().

anand-ky (changes requested):

PR #869, opened 2021-02-23
LLAMA-936: Notify all handlers

npoltorapavlo:
Reason for change: By default, Notify in
PluginHost::JSONRPC notifies only the first handler.
Test Procedure: Use code sample from ticket
Risks: Low


anand-ky (approved):
looks good. Pls make the same change in cpc rdkservices (https://gerrit.teamccp.com/plugins/gitiles/rdk/components/cpc/rdkservices/generic)

PR #916, opened 2021-03-09
RDKTV-2522 - Conditional factory app launches on first resident app a…

kannanmadanagopalt:
…ctivation

### DIFF
--- a/RDKShell/RDKShell.cpp
+++ b/RDKShell/RDKShell.cpp
@@ -127,6 +127,7 @@ bool gRdkShellSurfaceModeEnabled = false;
 static std::string sThunderSecurityToken;
 std::shared_ptr<WPEFramework::JSONRPC::LinkType<WPEFramework::Core::JSON::IElement>> gSystemServiceConnection;
 bool gSystemServiceEventsSubscribed = false;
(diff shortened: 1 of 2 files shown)


mfiess (approved):

RDKShell/RDKShell.cpp
```diff
@@ -236,6 +238,16 @@ namespace WPEFramework {
...
                    gSystemServiceConnection = getThunderControllerClient(serviceCallsign);
                 }
+                else if (currentState == PluginHost::IShell::ACTIVATED && service->Callsign() == RESIDENTAPP_CALLSIGN && sResidentAppFirstLaunch)
+                {
+                    sResidentAppFirstLaunch = false;
+                    bool launchFactoryApp = mShell.checkForBootupFactoryAppLaunch();
+                    if (launchFactoryApp)
+                    {
+                      JsonObject request, response;
+                      uint32_t status = getThunderControllerClient("org.rdk.RDKShell.1")->Invoke(0, "launchFactoryApp", request, response);
```
mfiess, line 248: We should rework this to avoid a thunder api call.  for example something like: `mShell.launchFactoryApp()`

PR #919, opened 2021-03-10
[WebKitBrowser] Check for config line parse errors

emutavchi:


goruklu (approved):

HaseenaSainul (dismissed):
LGTM

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -1442,7 +1442,16 @@ static GSourceFuncs _handlerIntervention =
             _service = service;
 
             _dataPath = service->DataPath();
-            _config.FromString(service->ConfigLine());
+
+            string configLine = service->ConfigLine();
+            Core::OptionalType<Core::JSON::Error> error;
+            if (_config.FromString(configLine, error) == false) {
+                SYSLOG(Trace::Error,
```
MFransen69, line 1449: There is as special logging category for parsing errors, please see here: https://github.com/rdkcentral/Thunder/blob/d9feb8b7c0c953f0a9a384ad63cc5dfceca12f57/Source/plugins/Shell.cpp#L125

I think better to use that one here as well

MFransen69 (changes requested):

WebKitBrowser/WebKitImplementation.cpp
emutavchi: @MFransen69 thanks for the suggestion. I pushed one more commit to address this.

MFransen69 (approved):

PR #935, opened 2021-03-12
RDK-30161 : Implemented ScreenCapture for Realtek platform.

sborushevsky:
RDK-30161 : Fixed ScreenCapture colors for Realtek.

### DIFF
--- a/ScreenCapture/ScreenCapture.cpp
+++ b/ScreenCapture/ScreenCapture.cpp
@@ -406,6 +406,7 @@ namespace WPEFramework
             if (32 != pf->bitsPerPixel)
             {
                 LOGERR("Unsupported bits per pixel: %d", pf->bitsPerPixel);
+                fbDestroy(context);
                 return false;
             }
 
@@ -436,21 +437,22 @@ namespace WPEFramework
                 if(!saveToPng(buffer, w, h, png_out_data))
                 {
                     LOGERR("could not convert Nexus screenshot to png");
+                    fbDestroy(context);
                     return false;
                 }
-                else
-                    return true;
-
-
                 LOGINFO("[Done]");
+
             } else {
                 LOGERR("fbGetFramebuffer=null");
+                fbDestroy(context);
+                return false;
             }
 
             err = fbDestroy(context);
             if (err != ErrNone)
                 LOGERR("fbDestroy fail");
 
+            return true;
         }
 #endif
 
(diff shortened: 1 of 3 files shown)


ScreenCapture/ScreenCapture.cpp
```diff
@@ -334,6 +345,115 @@ namespace WPEFramework
...
+                if(!saveToPng(buffer, w, h, png_out_data))
+                {
+                    LOGERR("could not convert Nexus screenshot to png");
+                    return false;
+                }
+                else
+                    return true;
+
+
+                LOGINFO("[Done]");
```
mlyszczek, line 445: This will never be called.

ScreenCapture/ScreenCapture.cpp
```diff
@@ -334,6 +345,115 @@ namespace WPEFramework
...
+                    }
+                }
+
+                if(!saveToPng(buffer, w, h, png_out_data))
+                {
+                    LOGERR("could not convert Nexus screenshot to png");
+                    return false;
+                }
+                else
+                    return true;
```
mlyszczek, line 442: I don't have documentation fo fbInit(), but since context is a simple pointer, and pointer to context is passed to fbInit() I assume it must be malloc()ing some memory. So each time you successfully savetopng() you will leak memory. You should add fbDestroy(context) here (and in all places of early return after fbInit()).

mlyszczek (changes requested):
Please, address my memory leak concerns.

ScreenCapture/ScreenCapture.cpp
sborushevsky: Fixed memory leak.

ScreenCapture/ScreenCapture.cpp
```diff
@@ -334,6 +345,117 @@ namespace WPEFramework
...
+            FBContext *context = NULL;
+            err = fbCreate(&context);
+            if (err != ErrNone) {
+                LOGERR("fbCreate fail");
+                return false;
+            }
+
+            err = fbInit(context, &api, NULL);
+            if (err != ErrNone) {
+                LOGERR("fbInit fail");
```
mlyszczek, line 387: One more possible memory leak left here, in case fbInit() fails. It probably will be super rare, but still.

sborushevsky: Fixed this one as well.

mlyszczek:
looks good to me

anand-ky (approved):

PR #937, opened 2021-03-14
ActivityMonitor: remove delay when stopping monitor

mlyszczek:
Activity monitor has unconditional sleep in its thread at the very
end of while(1) loop. When activity monitor is started, caller
passes inverval in seconds, which defines how often monitoring loop
shall be triggered.

When monitor is being stopped, code will set flag to notify loop to
stop, but this flag will be read only after inverval (or what is
left of it) seconds of time passed, which can lead to high delay
when inverval is high enough.

This patch implements stopping on conditional variables. Monitor
will sleep in conditional variable wait_for() instead of usleep(),
and stop function will notify monitoring thread, thus forcing it to
stop immediately without waiting for whole inverval.

### DIFF
--- a/ActivityMonitor/ActivityMonitor.cpp
+++ b/ActivityMonitor/ActivityMonitor.cpp
@@ -204,17 +204,7 @@ namespace WPEFramework
         {
             LOGINFO();
 
-            {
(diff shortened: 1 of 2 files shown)


anand-ky (approved):
looks good.

PR #950, opened 2021-03-16
[WebKitBrowser] fix visibility getter

emutavchi:


MFransen69 (approved):

HaseenaSainul (commented):
There are many places we used _browser-> directly (without cast), so why the changes is required for this instance alone ?

emutavchi:
> There are many places we used _browser-> directly (without cast), so why the changes is required for this instance alone ?

the same cast is done in other 'getters'. the interface expects const object for getter, otherwise it acts as a setter: 

```
        virtual uint32_t Visible(bool& visible /* @out */) const = 0;
        virtual uint32_t Visible(const bool visible) = 0;
```

PR #953, opened 2021-03-17
RDK-31813: App control friendly name

apatel859:


bobseamon (approved):
Note: Need to move code from destructor to deinitialize.  Will be done in a separate ticket.  Amit will create.

PR #954, opened 2021-03-17
RDK-31813: App control friendly name

apatel859:


apatel859:
duplicate with 953

PR #958, opened 2021-03-17
RDK-31823: platco hisense READ EDID thunder API

acheri988:
Reason for change:
platco hisense READ EDID thunder API
IARM modifications
Test Procedure: None
Risks: Low

Change-Id: I121ac4f3b4168c542a9693ceaf8d728556c5801b

### DIFF
--- a/HdmiInput/HdmiInput.cpp
+++ b/HdmiInput/HdmiInput.cpp
@@ -26,6 +26,9 @@
 #include "dsError.h"
 #include "dsMgr.h"
 
+#include <vector>
+#include <algorithm>
+
 #define HDMI_HOT_PLUG_EVENT_CONNECTED 0
 #define HDMI_HOT_PLUG_EVENT_DISCONNECTED 1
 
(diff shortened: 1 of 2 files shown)


HdmiInput/HdmiInput.cpp
```diff
@@ -250,10 +255,25 @@ namespace WPEFramework
         uint32_t HdmiInput::readEDIDWrapper(const JsonObject& parameters, JsonObject& response)
         {
             LOGINFOMETHOD();
+            returnIfParamNotFound(parameters, "portId");
```
anand-ky, line 258: What is portId? Is it same as deviceId in writeEDID? Can we then use the same names? Also we should retain existing behavior and return EDID of default port when portId is not specified.

anand-ky (changes requested):

acheri988:
Old one. New one created: https://github.com/rdkcentral/rdkservices/pull/959

PR #960, opened 2021-03-18
Plugin template

fasilkv:
RDKCOM-2000: plugin template 

Reason for change: add custom plugin template 
Test Procedure: none 
Risks: Low

Signed-off-by: fasil kv <fasil_KV@comcast.com>


anand-ky:
Checking if we need this plugin template. We can refer to any of the existing thunder plugins to check how to create a plugin.

natfarleydev:
This is definitely needed, I've been attempting to make a plugin from examples alone but they all have little things which are undocumented making it unclear what's _actually_ needed for a plugin. Documentation like this makes it much easier for first time plugin creators.

Firdose2109:

PR #972, opened 2021-03-22
XIONE-3215: getFirmwareDownloadPercent sometimes returns 100 early

rdamod:
Reason for change: Updates the percentage only for firmware and not to redirect response codes before.
Risks:Low
Test Procedure: Check API working.

Signed-off-by: Rohith Damodaran <Rohith_Damodaran@comcast.com>


rdamod:
Verified on Jenkins.

anand-ky (approved):

PR #982, opened 2021-03-23
RDK-29971: Thunder Messenger Restrictions

npoltorapavlo:
Reason for change: Add method "create" and modify the flow,
now if security enabled only apps that match the regex
will be allowed to join the room.
Test Procedure: See ticket
Risks: Medium


anand-ky:
Changes look good. We need to wait for Apps to update their code to call create before join().

npoltorapavlo:
#1121

PR #1011, opened 2021-03-29
RDK-30925 : added getLastWakeupKey call.

sborushevsky:


sborushevsky:
> Let's avoid IARM code in RDKShell. I'll add a new RDKShell library call today that will allow the RDKShell thunder plugin to query the last key press. Then we can update the RDKShell thunder plugin to use the new library API to query the last key press.

Got rid of IARM code, used CompositorController::getLastKeyPress call instead.

mfiess (approved):

PR #1033, opened 2021-04-06
RDK-32035 : Modify USB Access Plugin

npoltorapavlo:
Reason for change: getFileList should return
folders, and files that have media extensions.
Test Procedure: See parent ticket
Risks: Low


npoltorapavlo:
brings #987

anand-ky (approved):

PR #1038, opened 2021-04-07
https://jira.rdkcentral.com/jira/projects/HP40A/issues/HP40A-995?filt…

VenkateshUmmadiSetty:
…er=allopenissues    changes made in the const string Network::Initialize(PluginHost::IShell* /* service */) function in Networn.cpp file    
eventHandler function in the Network.cpp file is not getting invoked so modified the code in the initialization function in Network.cpp file.

### DIFF
--- a/Network/Network.cpp
+++ b/Network/Network.cpp
@@ -175,10 +175,10 @@ namespace WPEFramework
             if (Utils::IARM::init())
             {
                 IARM_Result_t res;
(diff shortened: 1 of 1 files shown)


arun-madhavan-013:
@vijs ,
Could you please review/merge?
This change is required for https://jira.rdkcentral.com/jira/browse/VIP7802-250

PR #1046, opened 2021-04-08
DELIA-42792 SystemServices: use strings in firmwareUpdateState

michal-lyszczek-red:
Right now firmwareUpdateState returns update state as int (ranging
from 0 to 6), this patch changes it so it returns state as string
like "DOWNLOADING_COMPLETE" or "PREPARING_TO_REBOOT".


anand-ky:
Old PR. Closing it since change is not required

PR #1093, opened 2021-04-16
RDKTV-4713 : Update Amlogic chipset name

ababu250:
RDKTV-4713 : Update Amlogic chipset name

Reason for change: Remove white space from chip set name

Test Procedure: Test DeviceIdentification plugin
Risks: None


anand-ky (approved):
The value is getting hardcoded but it should be fine since this chipset value should not change in the future as per amlogic.

PR #1110, opened 2021-04-20
RDK-31873 : Changed getDeviceInfo to return all available info if no parameters are passed.

sborushevsky:


anand-ky (approved):
looks good

PR #1112, opened 2021-04-21
Revert "Revert "RDK-32142: UsbAccess plugin update per spec""

npoltorapavlo:
Reverts rdkcentral/rdkservices#1111


anand-ky (approved):
looks good

PR #1121, opened 2021-04-22
RDK-29971: Thunder Messenger security

npoltorapavlo:
Reason for change: Introduce secure rooms based on thunder security.
Test Procedure: To create a secure room, pass "secure":true and
"acl":[....] to "join". To join a secure room, pass "secure":true and
no/empty acl. To specify which app can set acl use "acl" method
in thunder acl config.
Risks: Low


anand-ky (commented):
@pwielders - Please review when you get a chance.

anand-ky:
Needs more changes based on last review from Pierre.

anand-ky (commented):
looks good. 
@pwielders - Pls review and approve.

Messenger/MessengerJsonRpc.cpp
```diff
@@ -75,7 +75,7 @@ namespace Plugin {
         const string& room = params.Room.Value();
 
         if (!user.empty() && !room.empty()) {
-            string roomId = JoinRoom(room, user);
+            string roomId = JoinRoom((room + "_" + string(params.Secure.Data())), user);
```
pwielders, line 78: The roomId is still postfixed with an '_' and called "insecure" or "secure". This, in my opinion is causing a very confusing impleentation (what if users start using the '_' and start using insecure/secure as postfix ? I thought we discussed to make this part of the map in which it is stored. Will indicate later where.
As our customers do not make use of this feature will mark this as a minor which means it is good to merge, although it might be a good idea to later on update this to an "explicit" flag that is *not* partof the room id.

pwielders (approved):
As the interface is not leaking the implementation details I approve the implementation (seems it is due for some deadline :-) )
However the implementation is not very explicit and error prone. As discussed I was assuming that the implementation would update the first template paremeters here:
https://github.com/rdkcentral/rdkservices/blob/48426a2313c0e010303df5cfa4dd566c4229fef7/Messenger/Messenger.h#L191-L192
to contain a class. This class would supply a roomid (as string, the old, default room id) but also some properties of the room, for example the type of the room (enum value, currently only supporting [secure|insecure] and whatever may happen in the future.
By attaching this information explicitly to the key is will make the code easier to maintina and more explicit with respect to its intentions.
On the interface Join() I would have expected a new parameter of the enum type to be passed to the actual implementation of the real room.
At the end the metrological customers will not be using it and on the interface they will not face any changes but this construction:

https://github.com/rdkcentral/rdkservices/blob/48426a2313c0e010303df5cfa4dd566c4229fef7/Messenger/MessengerJsonRpc.cpp#L122-L123

Would be much easier to understand and much less resource consuming if the enum would have been modelled directly into the this new class. For R3, I will update the rooms (using the same interface) but applying explicit code.

anand-ky:
sounds good, thanks Pierre. I will merge the current changes and can look at bringing your recommendations in R2 if needed.

PR #1132, opened 2021-04-26
RDKTV-4203 :  Fix for checking valid client name for opacity,scale an…

prani270:
…d generateKey

### DIFF
--- a/RDKShell/RDKShell.cpp
+++ b/RDKShell/RDKShell.cpp
@@ -4758,8 +4758,22 @@ namespace WPEFramework {
                   keyClient = keyInputInfo.HasLabel("callsign")? keyInputInfo["callsign"].String(): "";
                 }
                 lockRdkShellMutex();
-                ret = CompositorController::generateKey(keyClient, keyCode, flags, virtualKey);
-                gRdkShellMutex.unlock();
+                std::vector<std::string> clientList;
+                 CompositorController::getClients(clientList);
+                bool targetFound = false;
+                for (size_t i=0; i<clientList.size(); i++)
+               {
+                if (strcasecmp(clientList[i].c_str(),keyClient.c_str()) == 0)
+                {
+                    targetFound = true;
+                    break;
+                }
+                }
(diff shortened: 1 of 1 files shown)


RDKShell/RDKShell.cpp
```diff
@@ -4758,8 +4758,22 @@ namespace WPEFramework {
...
+                 CompositorController::getClients(clientList);
+                bool targetFound = false;
+                for (size_t i=0; i<clientList.size(); i++)
+               {
+                if (strcasecmp(clientList[i].c_str(),keyClient.c_str()) == 0)
+                {
+                    targetFound = true;
+                    break;
+                }
+                }
```
mfiess, line 4771: Let's use `std::find(clientList.begin(), clientList.end(), keyClient) != clientList.end()` instead.  Please ensure the indent spacing count matches the style of the method as well.

RDKShell/RDKShell.cpp
```diff
@@ -4951,7 +4965,21 @@ namespace WPEFramework {
...
+            CompositorController::getClients(clientList);
+            bool targetFound = false;
+            for (size_t i=0; i<clientList.size(); i++)
+            {
+                if (strcasecmp(clientList[i].c_str(),client.c_str()) == 0)
+                {
+                    targetFound = true;
+                    break;
+                }
+            }
```
mfiess, line 4978: Let's use `std::find(clientList.begin(), clientList.end(), client) != clientList.end()` instead.  Please ensure the indent spacing count matches the style of the method as well.

RDKShell/RDKShell.cpp
```diff
@@ -4969,7 +4997,21 @@ namespace WPEFramework {
...
```
mfiess, line 5010: Let's use `std::find(clientList.begin(), clientList.end(), client) != clientList.end()` instead.  Please ensure the indent spacing count matches the style of the method as well.

mfiess (changes requested):

PR #1135, opened 2021-04-27
DELIA-49986: Handle status and abort flag

livinsunny519:
Handle all the status, abort flags. Update docs.


MaintenanceManager/MaintenanceManager.h
```diff
@@ -83,10 +96,13 @@ namespace WPEFramework {
 
                 static string g_currentMode;
```
anand-ky, line 97: Please replace these static variables with members. There will be one one instance of the plugin running in thunder so we don't need to maintain state using globals or static members.

anand-ky (approved):

MaintenanceManager/MaintenanceManager.h
anand-ky: Approving for now due to DVT2 deadlines. Please submit follow-up change to get this fixed.

PR #1159, opened 2021-05-05
Adding fireboltmediaplayer

sverkoye:


justin-ware-red:
I've reviewed this and it looks fine (unmodified since sprint/2101 and includes fixes done there). The only that concerns me is should this micro-service be auto-starting? I would have thought not.

@goruklu - do you want to add me as a reviewer?

sverkoye:
@justin-ware-red , AFAIK it is not auto starting:
`set (autostart false)`

That's from one of the commits in the list: https://github.com/rdkcentral/rdkservices/pull/1159/commits/04484d88e256cf8a2ad45cc0704e971664d88074

justin-ware-red:
Apologies @sverkoye you're quite right. I should have looked closer.

PR #1178, opened 2021-05-11
DELIA-47718 : fixed setCollectionFrequency of FrameRate and DELIA-47719 - Checking the minimum frequency value

MManja792:


FrameRate/FrameRate.cpp
```diff
@@ -85,15 +85,24 @@ namespace WPEFramework
...
+			if( fpsFrequencyInMilliseconds > = 100) // make sure min freq is 100 and not less than that.
+			{
+        	    		setCollectionFrequency(fpsFrequencyInMilliseconds);
+            			returnResponse(true);
+			}
+            	}
+		}
+	     catch(...)
+	     {
+		returnResponse(false);
```
anand-ky, line 102: Pls catch the exception and add some logging and return error indicating that the value is invalid.

FrameRate/FrameRate.cpp
```diff
@@ -85,15 +85,24 @@ namespace WPEFramework
...
 
             LOGINFOMETHOD();
+            try{
+            	if (parameters.HasLabel("frequency"))
+            	{
+            		int fpsFrequencyInMilliseconds = DEFAULT_FPS_COLLECTION_TIME_IN_MILLISECONDS;
+	                fpsFrequencyInMilliseconds = std::stod(parameters["frequency"].String());
+			if( fpsFrequencyInMilliseconds > = 100) // make sure min freq is 100 and not less than that.
+			{
+        	    		setCollectionFrequency(fpsFrequencyInMilliseconds);
```
anand-ky, line 95: Add logging to indicate that the value is less than 100.

anand-ky (changes requested):

MManja792:
Added the logs on exception, invalid parameter and invalid value.

FrameRate/FrameRate.cpp
MManja792: Added the logs

anand-ky (approved):

PR #1182, opened 2021-05-11
RDKTV-6232 : setBlink does not stop on other LED APIs

manojmani1394:
Reason for change: To fix setBlink not stopping on other LED APIs
Test Procedure: setBlink should stop when setLED, setBrightness, power APIs are called
Risks: Create None
Signed-off-by: manoj mani manoj.mani@sky.uk

### DIFF
--- a/helpers/frontpanel.cpp
+++ b/helpers/frontpanel.cpp
@@ -115,6 +115,7 @@ namespace WPEFramework
 
         CFrontPanel::CFrontPanel()
         : m_blinkTimer(this)
+        , m_isBlinking(false)
         , mFrontPanelHelper(new FrontPanelHelper())
         {
         }
@@ -305,6 +306,7 @@ namespace WPEFramework
 
         bool CFrontPanel::powerOnLed(frontPanelIndicator fp_indicator)
         {
+            stopBlinkTimer();
             try
             {
                 if (powerStatus)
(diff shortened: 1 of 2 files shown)


helpers/frontpanel.cpp
```diff
@@ -624,18 +627,21 @@ namespace WPEFramework
...
             m_numberOfBlinks = 0;
+            m_isBlinking = true;
             m_maxNumberOfBlinkRepeats = numberOfBlinkRepeats;
             m_currentBlinkListIndex = 0;
             if (m_blinkList.size() > 0)
             {
                 FrontPanelBlinkInfo blinkInfo = m_blinkList.at(0);
                 setBlinkLed(blinkInfo);
-                blinkTimer.Schedule(Core::Time::Now().Add(blinkInfo.durationInMs), m_blinkTimer);
+                if (m_isBlinking)
```
anand-ky, line 637: Checking how will this work if we setBlink on different led indicators (power, data, record). Should we have different blink timers for each of these leds?

manojmani1394: Hi Anand,
Having multiple blink timers might fix issues while handling multiple LEDs, but it will still have this race condition between Revoke and Schedule functions. This fix avoids the overrides the race condition.

anand-ky: ok sounds good. Will take this change and can look at adding multiple timers to handle other LEDs.

anand-ky (approved):

PR #1198, opened 2021-05-13
Sprint/2103

ddevad:


apatel859 (approved):

DisplaySettings/DisplaySettings.cpp
```diff
@@ -139,6 +139,7 @@ namespace WPEFramework {
         {
             LOGINFO("ctor");
             DisplaySettings::_instance = this;
+	    Core::JSONRPC::Handler& DisplaySettings_v2 = JSONRPC::CreateHandler({ 2 }, *this);
```
anand-ky, line 142: We don't need to create this handler. You can just use the overloaded registerMethod() and pass the versions to be supported. Pls refer to System Plugin for example.

anand-ky (changes requested):

ddevad:
will add new one

PR #1207, opened 2021-05-14
RDKCOM-2087, RDKCOM-1978, RDKCOM-2090: WiFiManager added new getSavedSSID method for retrieving saved SSID, Introducing Amlogic STB specific flag for Network plugin fixes.

arun-madhavan-013:
There is no get/set pair for SaveSSID method in WiFiManager plugin.
Bringing in a method for retrieving getSavedSSID.
Logic is: Utilize PersistentStore plugin to save and read saved credentials.


vijs (changes requested):
Caching the SSID details in the PersistentStore won't work as it will get outdated when WiFi credentials are changed through the means including WPS.

arun-madhavan-013:
Discarding this change as it increases security vulnerabilities and also Comcast's LostAndFound feature does not require this. But community does not have that.
Since same logic can be implemented in UI; if needed UI can do.

PR #1214, opened 2021-05-17
RDK-32574:Create all new compositors in the same thread

madanagopalt:


RDKShell/RDKShell.cpp
```diff
@@ -397,6 +442,34 @@ namespace WPEFramework {
             }*/
         }
 
+        static bool isClientExists(std::string client)
+        {
+            bool exist = false;
+            for (int i=0; i<gCreateDisplayRequests.size(); i++)
```
mfiess, line 448: Let's add a mutex around `gCreateDisplayRequests` to ensure this list is not modified at the same time by another thread.

mfiess (changes requested):

PR #1229, opened 2021-05-19
RDKTV-4742 : Saved key, used to power on the device for getLastWakeup…

sborushevsky:
…Key call.

### DIFF
--- a/RDKShell/RDKShell.cpp
+++ b/RDKShell/RDKShell.cpp
@@ -161,6 +161,8 @@ static uint32_t gWillDestroyEventWaitTime = RDKSHELL_WILLDESTROY_EVENT_WAITTIME;
 #define RESIDENTAPP_CALLSIGN "ResidentApp"
 #define PERSISTENT_STORE_CALLSIGN "org.rdk.PersistentStore"
 
+
+#define RECONNECTION_INITIAL_DELAY_MILLISECONDS 2000
 #define RECONNECTION_TIME_IN_MILLISECONDS 10000
 
 enum FactoryAppLaunchStatus
@@ -642,7 +644,7 @@ namespace WPEFramework {
         }
 
         RDKShell::RDKShell()
-                : AbstractPlugin(API_VERSION_NUMBER_MAJOR), mClientsMonitor(Core::Service<MonitorClients>::Create<MonitorClients>(this)), mEnableUserInactivityNotification(true), mCurrentService(nullptr), mLastWakeupKeyTimestamp(0)
+                : AbstractPlugin(API_VERSION_NUMBER_MAJOR), mClientsMonitor(Core::Service<MonitorClients>::Create<MonitorClients>(this)), mEnableUserInactivityNotification(true), mCurrentService(nullptr), mLastWakeupKeyCode(0), mLastWakeupKeyModifiers(0), mLastWakeupKeyTimestamp(0)
         {
             LOGINFO("ctor");
             RDKShell::_instance = this;
@@ -994,11 +996,9 @@ namespace WPEFramework {
                 gWillDestroyEventWaitTime = atoi(willDestroyWaitTimeValue); 
             }
 
-            if (Core::ERROR_NONE != subscribeForSystemEvent("onSystemPowerStateChanged"))
-            {
-                m_timer.start(RECONNECTION_TIME_IN_MILLISECONDS);
-                std::cout << "Started SystemServices connection timer" << std::endl;
-            }
+            m_timer.start(RECONNECTION_INITIAL_DELAY_MILLISECONDS);
+            m_timer.setInterval(RECONNECTION_TIME_IN_MILLISECONDS);
+            std::cout << "Started SystemServices connection timer" << std::endl;
 
             return "";
         }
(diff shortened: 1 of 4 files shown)


RDKShell/RDKShell.cpp
```diff
@@ -646,7 +642,7 @@ namespace WPEFramework {
         }
 
         RDKShell::RDKShell()
-                : AbstractPlugin(API_VERSION_NUMBER_MAJOR), mClientsMonitor(Core::Service<MonitorClients>::Create<MonitorClients>(this)), mEnableUserInactivityNotification(true), mCurrentService(nullptr)
+                : AbstractPlugin(API_VERSION_NUMBER_MAJOR), mClientsMonitor(Core::Service<MonitorClients>::Create<MonitorClients>(this)), mEnableUserInactivityNotification(true), mCurrentService(nullptr), mLastWakeupKeyTimestamp(0)
```
madanagopalt, line 645: initialize all newly added member variables

RDKShell/RDKShell.cpp
```diff
@@ -995,6 +993,13 @@ namespace WPEFramework {
             {
                 gWillDestroyEventWaitTime = atoi(willDestroyWaitTimeValue); 
             }
+
+            if (Core::ERROR_NONE != subscribeForSystemEvent("onSystemPowerStateChanged"))
```
madanagopalt, line 997: avoid doing subscription call here as, if it goes to activation, it may delay rdkshell initialization. Isntead during initialization, start the timer with minimal delay

RDKShell/RDKShell.cpp
```diff
@@ -4555,39 +4562,27 @@ namespace WPEFramework {
...
-                Utils::activatePlugin(SYSTEM_SERVICE_CALLSIGN);
-                std::cout << "Activated SystemService" << std::endl;
-            }
-
-            if (nullptr != gSystemServiceConnection)
+            
+            if (0 != mLastWakeupKeyTimestamp)
             {
                 JsonObject req, res;
+                gRdkShellMutex.lock();
```
madanagopalt, line 4569: We may not need a lock here, as if thunder call blocked, it may block ui thread

RDKShell/RDKShell.cpp
```diff
@@ -5616,6 +5611,79 @@ namespace WPEFramework {
...
+                    m_timer.stop();
+                    std::cout << "Stopped SystemServices connection timer" << std::endl;
+                }
+            }
+        }
+
+        int32_t RDKShell::subscribeForSystemEvent(std::string event)
+        {
+            int32_t status = Core::ERROR_GENERAL;
+            gRdkShellMutex.lock();
```
madanagopalt, line 5637: we don't want a lock here

RDKShell/RDKShell.cpp
sborushevsky: Initialized.

RDKShell/RDKShell.cpp
sborushevsky: Added 2 seconds delay.

RDKShell/RDKShell.cpp
sborushevsky: Removed.

RDKShell/RDKShell.cpp
sborushevsky: Removed.

RDKShell/RDKShell.cpp
```diff
@@ -5616,6 +5608,76 @@ namespace WPEFramework {
...
+                    std::cout << "Stopped SystemServices connection timer" << std::endl;
+                }
+            }
+        }
+
+        int32_t RDKShell::subscribeForSystemEvent(std::string event)
+        {
+            int32_t status = Core::ERROR_GENERAL;
+
+            if (!Utils::isPluginActivated(SYSTEM_SERVICE_CALLSIGN))
```
madanagopalt, line 5635: please keep it as a single call and store a result and reuse in below conditions is plugin aleady activated

RDKShell/RDKShell.cpp
```diff
@@ -5616,6 +5608,76 @@ namespace WPEFramework {
...
+            {
+                Utils::activatePlugin(SYSTEM_SERVICE_CALLSIGN);
+                std::cout << "called activatePlugin for SystemService" << std::endl;
```
madanagopalt, line 5638: set systemserviceplugin reference to null here

RDKShell/RDKShell.cpp
```diff
@@ -995,6 +995,11 @@ namespace WPEFramework {
             {
                 gWillDestroyEventWaitTime = atoi(willDestroyWaitTimeValue); 
             }
+
+            m_timer.start(RECONNECTION_INITIAL_DELAY_MILLISECONDS);
```
madanagopalt, line 999: Can we make it to start immediately with timeout of 0

RDKShell/RDKShell.cpp
```diff
@@ -995,6 +995,11 @@ namespace WPEFramework {
...
+            m_timer.setInterval(RECONNECTION_TIME_IN_MILLISECONDS);
```
madanagopalt, line 1000: Also, is system service not getting activated before rdkshell and we need to activate system service from rdkshell?. please clarify

RDKShell/RDKShell.cpp
sborushevsky: On my Xi5 it takes less that a second.

RDKShell/RDKShell.cpp
sborushevsky: Set.

RDKShell/RDKShell.cpp
sborushevsky: Changed delay to 0.

RDKShell/RDKShell.cpp
sborushevsky: I has "autostart" = false. I can't be sure that it's activated by UI of something else.

mfiess (approved):

PR #1246, opened 2021-05-21
Revert "DELIA-50464:Script not populating last fw details."

anand-ky:
Reverts rdkcentral/rdkservices#1233


anand-ky:
Need to make a fix to append to /opt/logs/swupdate.log instead of overwriting it.

PR #1247, opened 2021-05-21
RDKTV-1694: HDMI do not display resolution

ddevad:
Reason for change: Added support for videoStreamInfoUpdate
event
Test Procedure: Check thunder events are fired
whenever video mode changes
Risks: None

Signed-off-by: Deekshit Devadas <deekshit.devadasy@sky.uk>


ddevad:
will add a new one.  Abandon

PR #1270, opened 2021-05-25
RDK-31097 DeviceDiagnostics: fix multiple issues after testing

mlyszczek:
- change event response from video to videoDecoderStatusChange,
    same for audio
- send event on every decoder status upate, not only for decoders
    most recently read via getAudioDecoderStatus
- add some info logs, decoder updates from gstreamer are not too
    frequent, so this will not pollute logs too much.

### DIFF
--- a/DeviceDiagnostics/DeviceDiagnostics.cpp
+++ b/DeviceDiagnostics/DeviceDiagnostics.cpp
@@ -137,35 +137,27 @@ namespace WPEFramework
         }
         /* Searches m_{video,audio}DecoderStatus for most active decoder.
          * Most active status is "Active" followed by "Paused" and then
-         * "Idle". Key of decoder, that status was returned for, is stored
-         * in m_mast{Video,Audio}DecoderStatus.
+         * "Idle".
          *
          * decoderName can be "video" or "audio"
          *
          * When there are no decoders in map, IDLE state will be returned.
          */
-        std::string DeviceDiagnostics::getMostActiveDecoderStatus(const std::string &decoderName)
+        DeviceDiagnostics::DecoderStatus DeviceDiagnostics::getMostActiveDecoderStatus(const std::string &decoderName)
         {
             std::unordered_map<std::string, DecoderStatusInfo> *decoderStatus;
-            std::string *lastDecoder;
             DecoderStatus mostActiveStatus = DECODER_STATUS_IDLE;
 
-            if (decoderName == "video") {
+            if (decoderName == "video")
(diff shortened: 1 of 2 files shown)


DeviceDiagnostics/DeviceDiagnostics.cpp
```diff
@@ -264,14 +259,14 @@ namespace WPEFramework
         uint32_t DeviceDiagnostics::getVideoDecoderStatus(const JsonObject& parameters, JsonObject& response)
         {
             LOGINFOMETHOD();
-            response["video"] = getMostActiveDecoderStatus("video");
+            response["videoDecoderStatusChange"] = getMostActiveDecoderStatus("video");
```
anand-ky, line 262: the param should be "videoDecoderStatus". The param in the event should be "videoDecoderStatusChange".

DeviceDiagnostics/DeviceDiagnostics.cpp
```diff
@@ -264,14 +259,14 @@ namespace WPEFramework
...
             returnResponse(true);
         }
 
         uint32_t DeviceDiagnostics::getAudioDecoderStatus(const JsonObject& parameters, JsonObject& response)
         {
             LOGINFOMETHOD();
-            response["audio"] = getMostActiveDecoderStatus("audio");
+            response["audioDecoderStatusChange"] = getMostActiveDecoderStatus("audio");
```
anand-ky, line 269: the param should be "audioDecoderStatus". The param in the event should be "audioDecoderStatusChange".

DeviceDiagnostics/DeviceDiagnostics.cpp
```diff
@@ -241,12 +238,10 @@ namespace WPEFramework
...
             }
 
+            /* add/update map with new state and notify any listener */
             (*decoderStatus)[std::string(eventData->pipeline_id)] = decoderInfo;
-
-            /* this pipeline has been most recently read, send event
-             * to inform caller status has changed */
-            if (strcmp(eventData->pipeline_id, lastDecoder->c_str()) == 0)
-                t->onDecoderStatusChange(eventData->decoder, eventData->status);
+            t->onDecoderStatusChange(eventData->decoder, eventData->status);
```
anand-ky, line 243: The onVideoDecoderStatusChanged and onAudioDecoderStatusChanged events should also aggregate (call getMostActiveDecoderStatus) and then fire the event if the aggregate has changed.

anand-ky (changes requested):

anand-ky (approved):

PR #1288, opened 2021-05-28
[RDKTV-3218] As part of Cobalt-PreCertification updated the Speech Da…

anjalirajan:
…ta Storage based on the log levels.

As part of speech text logging

- Changed the log level of httpsrc in GST_DEBUG environment variable.

- Corrected the code to reflect the log level based on the TTS_DEFAULT_LOG_LEVEL environment variable. 

- As part of event notification, log the speech text for TRACE level, otherwise only speech id is logged.


anand-ky:

PR #1296, opened 2021-05-28
RDK-32357 : Added Telemetry plugin

sborushevsky:


Telemetry/Telemetry.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+                while ((de = readdir(d)))
+                {
+                    if (0 == de->d_name[0])
+                        continue;
+
+                    if (0 == strcmp(de->d_name, ".") || 0 == strcmp(de->d_name, ".."))
+                        continue;
+
+                    profiles.Add(de->d_name);
```
anand-ky, line 164: Pls add json object here {"name":"profile_1"} instead of just names. We can then expand it in the future to add more information (key-value pairs) about the profiles if needed.

Telemetry/Telemetry.cpp
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            returnResponse(true);
+        }
+
+        void Telemetry::getProfiles(JsonArray &profiles)
+        {
+            profiles.Clear();
+
+            DIR *d = opendir("/opt/.t2persistentfolder/");
```
anand-ky, line 150: Where are we reading the default FTUE report profile?

anand-ky (changes requested):

Telemetry/Telemetry.cpp
sborushevsky: Changed result of getAvailableReportProfiles(v1) to an array of objects, though Solution Overview in RDK-29174 says it should be array of strings.

Telemetry/Telemetry.cpp
sborushevsky: Looks like there is no such thing as default profile. Should I hardcode "FTUE" in the case of empty /opt/.t2persistentfolder/ ?

anand-ky (approved):

Telemetry/Telemetry.cpp
anand-ky: No, we shouldn't hardcode. Let's check with Mahee on where to read the default report profile. Approving these current changes.

PR #1331, opened 2021-06-04
XIONE-4741: HdmiCec_2 init not happening properly

acheri988:
Reason for change:
HdmiCec_2 init not happening properly
Test Procedure: None
Risks: Low

Change-Id: If67408d42236347e14ebfd741da9996acb0d2bea

### DIFF
--- a/HdmiCec_2/HdmiCec_2.cpp
+++ b/HdmiCec_2/HdmiCec_2.cpp
@@ -330,7 +330,8 @@ namespace WPEFramework
            HdmiCec_2::_instance = this;
            smConnection = NULL;
            InitializeIARM();
-
+           //Initialize cecEnableStatus to false in ctor
+           cecEnableStatus = false;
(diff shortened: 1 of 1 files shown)


HdmiCec_2/HdmiCec_2.cpp
```diff
@@ -330,7 +330,8 @@ namespace WPEFramework
            HdmiCec_2::_instance = this;
            smConnection = NULL;
            InitializeIARM();
```
anand-ky, line 332: Pls move the InitializeIARM and other HAL initialization to the Initialize() API and return empty string to indicate that it's initiated correctly. Any initialization in Initialize() should be cleaned up in DeInitialize() - https://github.com/rdkcentral/Thunder/blob/master/Source/plugins/IPlugin.h#L70

acheri988:
Ne one created: https://github.com/rdkcentral/rdkservices/pull/1332

PR #1336, opened 2021-06-07
DELIA-50319: wpeframework service crash

justin-ware-red:
The in-process portion of the FireboltMediaPlayer cannot handle the
    out-of-process portion stopped unexpectedly; either killed or crashing.
    This is because it is assuming the error codes that are returned rather
    than handing unexpected ones, such as ERROR_CONNECTION_CLOSED.

    Reason for change: explicitly killing the FireboltMediaPlayer
    out-of-process micro-service or having it crashes cause wpeframwork
    to crash.
    Risks: Low
    Test Procedure: Launch FireboltMediaPlayer and use kill -9 on its
    process.


justin-ware-red:
@anand-ky or @vijs - could you review please?

goruklu (approved):

PR #1342, opened 2021-06-08
RDKTV-6150: New TV gets ENT-41010 in activation

MarkVandenbriele:
Reason for change: Prevent reboot cycle on initial update.
Test Procedure: Update without /opt/fwdnldstatus.txt.
Risks: Low

Signed-off-by: Mark Vandenbriele <mark.vandenbriele@consult.red>

### DIFF
--- a/SystemServices/SystemServices.cpp
+++ b/SystemServices/SystemServices.cpp
@@ -1790,11 +1790,14 @@ namespace WPEFramework {
             std::vector<string> lines;
 
 	    if (!Utils::fileExists(FWDNLDSTATUS_FILE_NAME)) {
-		    populateResponseWithError(SysSrv_FileNotPresent, response);
-		    returnResponse(retStat);
+                //If firmware download file doesn't exist we can still return the current version
+                response["currentFWVersion"] = getStbVersionString();
+                response["downloadedFWVersion"] = downloadedFWVersion;
+                response["downloadedFWLocation"] = downloadedFWLocation;
+                response["isRebootDeferred"] = isRebootDeferred;
+                retStat = true;
 	    }
-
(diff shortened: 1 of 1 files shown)


SystemServices/SystemServices.cpp
```diff
@@ -1790,11 +1790,14 @@ namespace WPEFramework {
             std::vector<string> lines;
 
 	    if (!Utils::fileExists(FWDNLDSTATUS_FILE_NAME)) {
-		    populateResponseWithError(SysSrv_FileNotPresent, response);
-		    returnResponse(retStat);
+                //If firmware download file doesn't exist we can still return the current version
+                response["currentFWVersion"] = getStbVersionString();
```
anand-ky, line 1794: Pls check if getStbVersionString() returns "unknown" in which case return empty ("") and then return success as false.

anand-ky (changes requested):

SystemServices/SystemServices.cpp
```diff
@@ -1790,11 +1790,23 @@ namespace WPEFramework {
...
-		    returnResponse(retStat);
+                //If firmware download file doesn't exist we can still return the current version
+                response["downloadedFWVersion"] = downloadedFWVersion;
+                response["downloadedFWLocation"] = downloadedFWLocation;
+                response["isRebootDeferred"] = isRebootDeferred;
+                retStat = true;
+                string ver =  getStbVersionString();
+                if(ver == "unknown")
+                {
+                    response["currentFWVersion"] = "";
```
anand-ky, line 1801: retStat needs to be false here.

anand-ky (approved):

PR #1352, opened 2021-06-09
RDKTV-7412 firmware state file should be cleared after the reboot

michaelamal:
Reason for change:global variable is used to update state
Risks: No

Signed-off-by: Michael Anand Michael_AmalAnand@comcast.com

### DIFF
--- a/SystemServices/SystemServices.h
+++ b/SystemServices/SystemServices.h
@@ -99,7 +99,7 @@ namespace WPEFramework {
                 static JsonObject _systemParams;
                 static const string MODEL_NAME;
                 static const string HARDWARE_ID;
-
+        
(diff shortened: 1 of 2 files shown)


SystemServices/SystemServices.h
```diff
@@ -134,8 +134,11 @@ namespace WPEFramework {
                 void InitializeIARM();
                 void DeinitializeIARM();
 #endif /* defined(USE_IARMBUS) || defined(USE_IARM_BUS) */
-
+                
+                
                 /* Events : Begin */
+               
+                static int FwUpdateState_LatestEvent;
```
anand-ky, line 141: This doesn't need to be static. There can only be on instance of a plugin running in thunder. 
Pls change it to non-static private data member. Also pls change the other static members to be non-static.

static std::string m_currentMode;
static cTimer m_operatingModeTimer;
static int m_remainingDuration;

anand-ky (changes requested):

anand-ky:

PR #1355, opened 2021-06-09
RDK-31097 DeviceDiagnostics: replace IARM with ERM api

mlyszczek:
This patch, replaces IARM api with ERM. Logic about most active
decoder is in ERM library, so thunder only reads it and returns
value as string to the caller. Since ERM does not support event
new thread is spawned to poll ERM for status change, and notify
clients once decoder status changes.

To avoid hoggin CPU, ERM is polled once every 30 seconds, and
because of that, once decoder status changes, event with that
information may be delayed for up to 30 seconds.

getAVDecoderStatus api will always return up to date value from
ERM library.

### DIFF
--- a/DeviceDiagnostics/DeviceDiagnostics.cpp
+++ b/DeviceDiagnostics/DeviceDiagnostics.cpp
@@ -21,21 +21,14 @@
 
 #include <curl/curl.h>
 #include <libIBus.h>
+#include <time.h>
 
 #include "utils.h"
 
 #define DEVICE_DIAGNOSTICS_METHOD_NAME_GET_CONFIGURATION  "getConfiguration"
-#define DEVICE_DIAGNOSTICS_METHOD_GET_VIDEO_DECODER_STATUS "getVideoDecoderStatus"
-#define DEVICE_DIAGNOSTICS_METHOD_GET_AUDIO_DECODER_STATUS "getAudioDecoderStatus"
+#define DEVICE_DIAGNOSTICS_METHOD_GET_AV_DECODER_STATUS "getAVDecoderStatus"
 
-#define DEVICE_DIAGNOSTICS_EVT_ON_VIDEO_DECODER_STATUS_CHANGED "onVideoDecoderStatusChanged"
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -19,6 +19,10 @@ cmake_minimum_required(VERSION 3.3)
 
 find_package(WPEFramework)
 
+if(HAVE_ERM)
```
anand-ky, line 22: Pls move this to services.cmake. Also pls rename it to something like BUILD_ENABLE_ERM to match similar build flags.

DeviceDiagnostics/DeviceDiagnostics.cpp
```diff
@@ -77,28 +69,51 @@ namespace WPEFramework
...
+            ret = pthread_mutex_init(&m_AVDecoderStatusLock, NULL);
+            if (ret > 0)
+            {
+                LOGERR("pthred_mutex_init(m_AVDecoderStatusLock) failed: %s", strerror(ret));
+                EssRMgrDestroy(m_EssRMgr);
+                return "pthread_mutex_init(m_AVDecoderStatusLock) failed";
+            }
+
+            m_pollThreadRun = 1;
+            ret = pthread_create(&m_AVPollThread, NULL, AVPollThread, NULL);
```
anand-ky, line 90: Pls switch to std::thread instead of using pthread.

DeviceDiagnostics/DeviceDiagnostics.cpp
```diff
@@ -77,28 +69,51 @@ namespace WPEFramework
...
+#ifdef CONFIG_HAVE_ERM
+            int ret;
+
+            if ((m_EssRMgr = EssRMgrCreate()) == NULL)
+            {
+                LOGERR("EssRMgrCreate() failed");
+                return "EssRMgrCreate() failed";
+            }
+
+            ret = pthread_mutex_init(&m_AVDecoderStatusLock, NULL);
```
anand-ky, line 81: Pls switch to std::mutex instead of pthread mutex.

anand-ky (changes requested):

anand-ky (approved):

PR #1400, opened 2021-06-17
[SecurityAgent] TokenDispatcher now uses RPC::Communicator instead of…

sramani-metro:
… IPCServer - allowing multiple connections


anand-ky:

PR #1433, opened 2021-06-22
LLAMA-2068: Remove IARM Disconnect and Term

livinsunny519:
Reason for change: IARM_Bus_Disconnect() and IARM_Bus_Term())  are no longer required
during deinitiaize.
Test Procedure: Activate/Deactivate the plugin.
Risk: High

### DIFF
--- a/MaintenanceManager/MaintenanceManager.cpp
+++ b/MaintenanceManager/MaintenanceManager.cpp
@@ -310,9 +310,7 @@ namespace WPEFramework {
         void MaintenanceManager::InitializeIARM()
         {
             if (Utils::IARM::init()) {
-                LOGINFO();
                 IARM_Result_t res;
-                IARM_CHECK(IARM_Bus_Connect());
                 // Register for the Maintenance Notification Events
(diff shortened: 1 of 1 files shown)


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -310,7 +310,6 @@ namespace WPEFramework {
         void MaintenanceManager::InitializeIARM()
         {
             if (Utils::IARM::init()) {
-                LOGINFO();
                 IARM_Result_t res;
                 IARM_CHECK(IARM_Bus_Connect());
```
anand-ky, line 314: Pls remove this IARM_CHECK(IARM_Bus_Connect());
This is done within Utils::IARM::init().

anand-ky (changes requested):

MaintenanceManager/MaintenanceManager.cpp
livinsunny519: Removed and updated

anand-ky (approved):

PR #1454, opened 2021-06-25
RDKTV-8238 Get the last wakeup key from IR remote

sujeeshtp:
Reason for change: Get the last wakeup key on deep sleep wakeup

Test Procedure: Put the TV in deep sleep and wakeup using app/power key
and see in wpeframework log last key pressed key code is returning

Risks: Low


sujeeshtp:
resubmitting

PR #1464, opened 2021-06-28
RDKTV-8323: HDMI ARC/eARC Power State Handling

ddevad:
Reason for change:1) Added a separate cec ARC routing
thread for co-ordinated CEC ARC requests & events
handling
2) Added support for Audio Device Power On HdmiCecsink
method invocation in Displaysettings
3) Remove ARC initiation status check from
HdmiCecSink. Will be managed by Displaysettings plugin
4) No need to start timer if HdmICecSink is activated
and subscribtion for cec events & ARC init can be done
directly in InitAudioPorts. This will reduce time during bootup
& standby/on transitions
5) Some ARC devices send ARC intiation message
even when they are in standby state. Always send audio device
power on to hdmicecsink to wake up the device before routing audio
in cases when CEC ARC initiation handshake is completed
6) Request ARC Initiation from TV side if connected ARC Device
only sends systemAudioMode ON and doesn't initate ARC from it's
side
7) Cleanup ARC state when TV goes to Standby and start fresh
on TV ON. To recover from cases when ARC device doesn't send
systemAudioMode OFF or Terminate ARC when going to standby on
getting standby message from TV side.

Test Procedure: Refer Ticket
Risks: None

Signed-off-by: Deekshit Devadas deekshit.devadasy@sky.uk

### DIFF
--- a/DisplaySettings/DisplaySettings.cpp
+++ b/DisplaySettings/DisplaySettings.cpp
@@ -220,6 +220,10 @@ namespace WPEFramework {
 
 	    m_subscribed = false; //HdmiCecSink event subscription
 	    m_hdmiInAudioDeviceConnected = false;
+        m_arcAudioEnabled = false;
+	    m_currentArcRoutingState = ARC_STATE_ARC_TERMINATED;
+	    m_cecArcRoutingThreadRun = false;
+	    m_arcRoutingThread = std::thread(cecArcRoutingThread);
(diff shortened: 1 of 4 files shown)


apatel859 (approved):

vijs (approved):
Please do a full regression testing and also make sure the plugin is not trying to activate "org.rdk.HdmiInput" in devices where it is not available.

PR #1475, opened 2021-06-29
ONEM-20048: removed unregister

mikolaj-staworzynski-red:
It causes problems (crashes) in another scenario:
(1) Activate FMP
(2) Start playback
(3) Deactivate FMP through controller


jameslofthouse-RED (approved):
I no longer see the crash with these changes removed

mikolaj-staworzynski-red:
Latest version of FireboltMediaPlayer.cpp on sprint/2105 already contain (or better say does not contain problematic Unregister operations, that was a topic of that pull request: reference: https://github.com/rdkcentral/rdkservices/commits/sprint/2105/FireboltMediaPlayer/FireboltMediaPlayer.cpp

PR #1485, opened 2021-06-30
RDKTV-8205 Splash screen not showing after bootloader screen when lau…

binuinbaraj:
…nching Factory app


binuinbaraj:
Will create a new one

PR #1532, opened 2021-07-08
DELIA-51041: Added error checks in Network Plugin Init

tabbas651:
DELIA-51050:Disabled NetSrvMgr check for XG1V3 and XI3 during Network Plugin activation
DELIA-51208:Disabled NetSrvMgr check for XG1 and XG2 during Network Plugin activation


vijs:
Changes are cherry-picked to main.

PR #1543, opened 2021-07-13
Pass autostart and out of process from cmake

dawid-trendota-red:


DisplayInfo/DisplayInfo.config
```diff
@@ -1,10 +1,10 @@
-set(autostart true)
+set(autostart ${PLUGIN_DISPLAYINFO_AUTOSTART})
 set(preconditions Platform)
 
 map()
     key(root)
     map()
-      kv(mode "Off")
+      kv(outofprocess ${PLUGIN_DISPLAYINFO_OUTOFPROCESS})
```
michalkucharczyk, line 7: _mode_ key shall not be removed

michalkucharczyk (changes requested):

PR #1564, opened 2021-07-15
[SecurityAgent] TokenDispatcher now uses RPC::Communicator instead of…

npoltorapavlo:
… IPCServer - allowing multiple connections


npoltorapavlo:
merge PR #1400 into 2105

vijs (approved):

PR #1568, opened 2021-07-15
Liberty global onem 20606 plugins pass config from cmake

dawid-trendota-red:


TraceControl/TraceControl.config
```diff
@@ -1,7 +1,8 @@
-set(PLUGIN_TRACECONTROL_AUTOSTART true CACHE BOOL "Automatically start TraceControl plugin")
+set(PLUGIN_TRACECONTROL_AUTOSTART ${PLUGIN_TRACECONTROL_AUTOSTART} CACHE BOOL "Automatically start TraceControl plugin")
```
michalkucharczyk, line 1: This duplicates line 5. It is not needed.

PR #1569, opened 2021-07-15
Onem 20606 plugins pass config from cmake

dawid-trendota-red:


michalkucharczyk (dismissed):

DisplayInfo/DisplayInfo.config
```diff
@@ -1,10 +1,15 @@
-set(autostart true)
+set(PLUGIN_DISPLAYINFO_AUTOSTART true CACHE BOOL "Automatically start DisplayInfo plugin")
+set(PLUGIN_DISPLAYINFO_OUTOFPROCESS false CACHE BOOL "Start DisplayInfo plugin in out of process")
```
pwielders, line 2: outofprocess does not exist as a field name in the root object, only in the configuration object and is deprecated. The root object is its successor to give more information for out of process properties, like container location or host IP for running on different systems)
So please remove the PLUGIN_DISPLAYINFO_OUTOFPROCESS completeley here it has no added value and is confusing.

pwielders (changes requested):

DisplayInfo/DisplayInfo.config
dawid-trendota-red: It should be fixed with:

https://github.com/rdkcentral/rdkservices/pull/1569/commits/d8a973059e58af38ac3833f5ca3e5c29391352b3

pwielders (approved):

vijs (approved):

PR #1586, opened 2021-07-19
ONEM-20194 - don't force westeros sink. Allow to predefine it with env

lukasz-wasylyk-red:
AAMP_ENABLE_WESTEROS_SINK may be set gracefully, not forcing in case enviroment has it already predefined. This way we can set it to 0 and use aamp properly.


michalkucharczyk (approved):

jameslofthouse-RED (approved):
This looks ok to me

vijs (approved):

PR #1589, opened 2021-07-19
ONEM-18263: OCDM plugin startup as container

lukasz-wasylyk-red:
kv container is added conditionally to config


OpenCDMi/OCDM.config
```diff
@@ -8,6 +8,9 @@ endif()
 
 map()
     kv(outofprocess ${PLUGIN_OPENCDMI_OOP})
+    if(PLUGIN_OPENCDMI_CONTAINER)
```
MFransen69, line 11: Although it might work better to not have the outofprocess from the line above together with the newer "mode" keyword this could be confusing, mode is the newer version of outofprcocess with more options (e.g. outofprocess:false equals "mode":"off"). So having same in one config file might be confusing. So I would also use the if above to not include outofprocess.

MFransen69 (changes requested):

OpenCDMi/OCDM.config
lukasz-wasylyk-red: Yes, you are correct, please see correction from LGI side.

MFransen69 (approved):

OpenCDMi/OCDM.config
pwielders: Maybe , if you look at pull -request #1569, it is usefull to use the same semantics. Allow from the outside to define the mode, like: kv(mode ${PLUGIN_OPENCDMI_MODE})

pwielders (changes requested):
Maybe , if you look at pull -request #1569, it is usefull to use the same semantics. Allow from the outside to define the mode, like: kv(mode ${PLUGIN_OPENCDMI_MODE}) and specify "Container" in the PLUGIN_OPENCDMI_MODE variable

OpenCDMi/OCDM.config
mikolaj-staworzynski-red: Updated. Please see corrections.

pwielders (approved):

pwielders (approved):

PR #1591, opened 2021-07-20
RDK-32545, RDK-33171, RDK-33172: Bluetooth Audio Output Mute & Volume…

ssirig714:
… Control

Reason for change: Thunder apis for the volume up/down and mute/unmute control
     of the paired BT audio out device.
Test Procedure: tested with curl commands passing new methods
Risks: Low
Signed-off-by: Sudheer Reddy Sirigireddy <SudheerReddy_Sirigireddy@cable.comcast.com>

### DIFF
--- a/Bluetooth/Bluetooth.cpp
+++ b/Bluetooth/Bluetooth.cpp
@@ -56,6 +56,8 @@ const string WPEFramework::Plugin::Bluetooth::METHOD_SET_EVENT_RESPONSE = "respo
 const string WPEFramework::Plugin::Bluetooth::METHOD_GET_DEVICE_INFO = "getDeviceInfo";
 const string WPEFramework::Plugin::Bluetooth::METHOD_GET_AUDIO_INFO = "getAudioInfo";
 const string WPEFramework::Plugin::Bluetooth::METHOD_GET_API_VERSION_NUMBER = "getApiVersionNumber";
+const string WPEFramework::Plugin::Bluetooth::METHOD_GET_DEVICE_VOLUME_MUTE_INFO = "getDeviceVolumeMuteInfo";
+const string WPEFramework::Plugin::Bluetooth::METHOD_SET_DEVICE_VOLUME_MUTE_INFO = "setDeviceVolumeMuteInfo";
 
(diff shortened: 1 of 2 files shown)


Bluetooth/Bluetooth.cpp
```diff
@@ -653,6 +662,83 @@ namespace WPEFramework
...
+                      Utils::String::contains(deviceProfile, "JOYSTICK")) {
+                      lenDevOpDiscType = BTRMGR_DEVICE_OP_TYPE_HID;
+             }
+             else if (Utils::String::contains(deviceProfile, "LE TILE") ||
+                      Utils::String::contains(deviceProfile, "LE")) {
+                      lenDevOpDiscType = BTRMGR_DEVICE_OP_TYPE_LE;
+             }
+             else if (Utils::String::contains(deviceProfile, "DEFAULT")) {
+                      lenDevOpDiscType = BTRMGR_DEVICE_OP_TYPE_UNKNOWN;
+             }
```
gladish, line 728: Since this code is repeated, you should create a static function in this .cpp like BTRMGR_DeviceOperationType_FromString(const std::string & s);

Bluetooth/Bluetooth.cpp
```diff
@@ -653,6 +662,83 @@ namespace WPEFramework
...
+
+             rc = BTRMGR_GetDeviceVolumeMute (0, deviceHandle, lenDevOpDiscType, &ui8volume, &mute);
+             if (BTRMGR_RESULT_SUCCESS != rc) {
+                 LOGERR("Failed to get the volume info");
```
gladish, line 732: Log the value of rc.

gladish (changes requested):

Bluetooth/Bluetooth.cpp
ssirig714: thanks for review . Addressed in latest diffs.

Bluetooth/Bluetooth.cpp
ssirig714: Addressed in latest diffs.

gladish (approved):

PR #1608, opened 2021-07-23
XIONE-1555: ContinueWatching API fails

MarkVandenbriele:
Reason for change: Return token correctly
Test Procedure: Run curl commands from ticket.
Risks: Low

Signed-off-by: Mark Vandenbriele <mark.vandenbriele@consult.red>


MarkVandenbriele:
Moving changes to separate branches.

PR #1611, opened 2021-07-23
LLAMA-2652 Device is getting stuck in stanby after start Maintenance …

michaelamal:
…start

Reason for change: Removing Request Reboot call as no action on it.
Risks: None

Signed-off-by: Michael Anand Michael_AmalAnand@comcast.com


michaelamal:
Abandon

PR #1614, opened 2021-07-23
LLAMA-2652 Device is not getting out of standby and when maintenance …

michaelamal:
…starts

Reason for change: remove  RequestReboot as it is not in use
Risks: No

Signed-off-by: Michael Amal Michael_AmalAnand@comcast.com

### DIFF
--- a/MaintenanceManager/MaintenanceManager.cpp
+++ b/MaintenanceManager/MaintenanceManager.cpp
@@ -340,7 +340,7 @@ namespace WPEFramework {
             /* on boot up we set these things */
             MaintenanceManager::g_currentMode = FOREGROUND_MODE;
 
-            MaintenanceManager::g_notify_status=MAINTENANCE_IDLE;
+            MaintenanceManager::m_notify_status=MAINTENANCE_IDLE;
             MaintenanceManager::g_epoch_time="";
 
             /* to know the maintenance is solicited or unsolicited */
@@ -352,7 +352,7 @@ namespace WPEFramework {
             MaintenanceManager::g_task_status=0;
 
             /* we post just to tell that we are in idle at this moment */
-            MaintenanceManager::_instance->onMaintenanceStatusChange(g_notify_status);
+            MaintenanceManager::_instance->onMaintenanceStatusChange(m_notify_status);
 
             int32_t exec_status=E_NOK;
 
@@ -374,6 +374,7 @@ namespace WPEFramework {
             /* we moved every thing to a thread */
             /* only when dcm is getting a DCM_SUCCESS/DCM_ERROR we say
              * Maintenance is started until then we say MAITENANCE_IDLE */
+
             m_thread = std::thread(&MaintenanceManager::task_execution_thread, _instance);
         }
 
@@ -389,7 +390,7 @@ namespace WPEFramework {
 
         void MaintenanceManager::iarmEventHandler(const char *owner, IARM_EventId_t eventId, void *data, size_t len)
         {
-            Maint_notify_status_t m_notify_status=MAINTENANCE_STARTED;
+            Maint_notify_status_t notify_status=MAINTENANCE_STARTED;
             IARM_Bus_MaintMGR_EventData_t *module_event_data=(IARM_Bus_MaintMGR_EventData_t*)data;
             IARM_Maint_module_status_t module_status;
             time_t successfulTime;
@@ -536,8 +537,6 @@ namespace WPEFramework {
                             /*will be set to false once COMEPLETE/ERROR received for LOGUPLOAD*/
                             LOGINFO(" LOGUPLOAD already IN PROGRESS -> setting m_task_map of LOGUPLOAD to true \n");
                             break;
-
-
                     }
                 }
                 else{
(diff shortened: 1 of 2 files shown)


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -813,6 +800,8 @@ namespace WPEFramework {
                     string abort_flag="";
 
                     /* only one maintenance at a time */
+                    /* Lock so that g_notify_status will not be updated  further */
+                    m_statusMutex.lock();
```
anand-ky, line 804: Why are we locking at 2 places? here and within onMaintenanceStatusChange() before updating g_notify_status. Once this lock is taken here, it's going to block the task_execution_thread from updating g_notify_status.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -585,23 +578,17 @@ namespace WPEFramework {
                             m_notify_status=MAINTENANCE_ERROR;
```
anand-ky, line 578: Pls change m_notify_status to notify_status to indicate it's a local variable. Also pls change g_notify_status to m_notify_status to indicate it's a member and not global.

anand-ky (changes requested):

MaintenanceManager/MaintenanceManager.cpp
michaelamal: fixed

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -550,7 +553,7 @@ namespace WPEFramework {
                 if ( (g_task_status & TASKS_COMPLETED ) == TASKS_COMPLETED ){
                     if ( (g_task_status & ALL_TASKS_SUCCESS) == ALL_TASKS_SUCCESS ){ // all tasks success
                         LOGINFO("DBG:Maintenance Successfully Completed!!");
-                        m_notify_status=MAINTENANCE_COMPLETE;
+                        notify_status=MAINTENANCE_COMPLETE;
```
mvalas827, line 556: You can update m_notify_status (class variable) here itself after holding on to status mutex.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -561,47 +564,31 @@ namespace WPEFramework {
...
-                            /* even though we end up in skipped task /error
-                             * check if we have the reboot required is recevied */
-                            requestSystemReboot();
-                        }
-                        else {
-                            LOGINFO("Reboot Not Required !!");
-                        }
                     }
 
+                    MaintenanceManager::_instance->onMaintenanceStatusChange(notify_status);
```
mvalas827, line 583: move notification after thread join.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -831,20 +820,28 @@ namespace WPEFramework {
...
+                    m_statusMutex.unlock();
                     returnResponse(result);
                 }
 
         void MaintenanceManager::onMaintenanceStatusChange(Maint_notify_status_t status) {
             JsonObject params;
             /* we store the updated value as well */
-            g_notify_status=status;
+            m_statusMutex.lock();
+            m_notify_status=status;
```
mvalas827, line 843: As commented before, you can remove this status update here as the class variable is updated and controlled by mutex.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -374,6 +374,11 @@ namespace WPEFramework {
             /* we moved every thing to a thread */
             /* only when dcm is getting a DCM_SUCCESS/DCM_ERROR we say
              * Maintenance is started until then we say MAITENANCE_IDLE */
+
```
mvalas827, line 377: I don't see a need for this. If the global/class variables are properly protected and notifications are sent appropriately.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -831,20 +820,28 @@ namespace WPEFramework {
                         /* we set this to false */
                         g_is_critical_maintenance="false";
 
+
+                        if(m_thread.joinable()){
```
mvalas827, line 824: I don't see a need for this. If the global/class variables are properly protected and notifications are sent appropriately.

MaintenanceManager/MaintenanceManager.cpp
mvalas827: Seems fine to me as the start method will return with in progress error. If start method was able to acquire lock then the previous execution should have been completed. Let me know if you still see some race condition here, we can discuss.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -561,47 +564,31 @@ namespace WPEFramework {
...
-                            LOGINFO("Reboot Not Required !!");
-                        }
                     }
 
+                    MaintenanceManager::_instance->onMaintenanceStatusChange(notify_status);
+                    LOGINFO("ENDING MAINTENANCE CYCLE"); 
                     if(m_thread.joinable()){
                         m_thread.join();
                     }
                 }
```
mvalas827, line 588: You can release mutex after thread join

mvalas827 (changes requested):

anand-ky (approved):

PR #1618, opened 2021-07-26
RDKTV-9273 Skip network dependent maintenance in offline

sujeeshtp:
Reason for change: When TV is offline mode some of the maintenance activity
is taking 2.5hours to complete the maintenance and blocking the TV to go to
deep sleep, this fails the energy certification test

Test Procedure: Keep the TV in offline mode and see TV is going to deepsleep
in 15min

Risks: Medium

### DIFF
--- a/MaintenanceManager/MaintenanceManager.cpp
+++ b/MaintenanceManager/MaintenanceManager.cpp
@@ -37,6 +37,7 @@
 #include <bits/stdc++.h>
 #include <algorithm>
 
+
 #include "MaintenanceManager.h"
 #include "utils.h"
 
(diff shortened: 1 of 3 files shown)


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -255,8 +258,11 @@ namespace WPEFramework {
             }
 
             LOGINFO("Reboot_Pending :%s",g_is_reboot_pending.c_str());
-
-            string cmd="";
+            /* Delay the network call to make sure the network plugins is activated
+	     * before making any call to network plugins
+	     */
+            sleep(30);
```
anand-ky, line 264: Why are we adding 30 seconds delay everytime? This is going to delay every maintenance activity thread.
Pls check if network plugin is activated before adding delay.

anand-ky (changes requested):

PR #1622, opened 2021-07-26
RDKTV-9273 Skip network dependent maintenance in offline

sujeeshtp:
Reason for change: When TV is offline mode some of the maintenance activity
is taking 2.5hours to complete the maintenance and blocking the TV to go to
deep sleep, this fails the energy certification test

Test Procedure: Keep the TV in offline mode and see TV is going to deepsleep
in 15min
Risks: Medium

(cherry picked from commit a4fa820e1de2e140c08737070bd669e55adb71e3)

### DIFF
--- a/MaintenanceManager/MaintenanceManager.h
+++ b/MaintenanceManager/MaintenanceManager.h
@@ -127,6 +127,10 @@ namespace WPEFramework {
                 std::condition_variable task_thread;
                 std::thread m_thread;
 
+                std::unordered_map<string, bool> m_task_map;    
+                        
+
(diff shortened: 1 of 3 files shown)


MaintenanceManager/MaintenanceManager.h
```diff
@@ -127,6 +127,10 @@ namespace WPEFramework {
                 std::condition_variable task_thread;
                 std::thread m_thread;
 
+                std::unordered_map<string, bool> m_task_map;
```
anand-ky, line 130: Do we need this m_task_map?

PR #1631, opened 2021-07-27
RDKTV-9417 : HDMI-CEC state does not persist

acheri988:
Reason for change:
HDMI-CEC_state_does_not_persist
Test Procedure: None
Risks: Low

Change-Id: I8374c97ed6a3ca901686d83a5e1115f8259a3baa

### DIFF
--- a/HdmiCec/HdmiCec.cpp
+++ b/HdmiCec/HdmiCec.cpp
@@ -322,7 +322,7 @@ namespace WPEFramework
             cecSetting.IElement::ToFile(file);
 
             file.Close();
-
+            Utils::syncPersistFile(CEC_SETTING_ENABLED_FILE);
             return;
         }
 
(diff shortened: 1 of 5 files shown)


HdmiCec/HdmiCec.cpp
```diff
@@ -322,7 +322,7 @@ namespace WPEFramework
             cecSetting.IElement::ToFile(file);
 
             file.Close();
-
+            Utils::syncPersistFile(CEC_SETTING_ENABLED_FILE);
```
anand-ky, line 325: Pls move the entire PersistSettings to Utils and name it something like PersistJsonSetting(file, Key, Val). Also don't need to create separate func for syncing. can be done within PersistJsonSetting.

acheri988:
New one created: https://github.com/rdkcentral/rdkservices/pull/1632

PR #1656, opened 2021-07-30
RDK-33287: Stop Maintenance API

livinsunny519:
Reason for change: Abort the maintenance task if an
api call to thunder plugin.
Test Procedure: Build and verify
Run all end to end scenarios.
Risk: High

### DIFF
--- a/MaintenanceManager/MaintenanceManager.cpp
+++ b/MaintenanceManager/MaintenanceManager.cpp
@@ -65,7 +65,10 @@ using namespace std;
 #define API_VERSION_NUMBER_MINOR 0
 #define SERVER_DETAILS  "127.0.0.1:9998"
 
+
+#define PROC_DIR "/proc"
 #define TR181_AUTOREBOOT_ENABLE "Device.DeviceInfo.X_RDKCENTRAL-COM_RFC.Feature.AutoReboot.Enable"
+#define TR181_STOP_MAINTENANCE  "Device.DeviceInfo.X_RDKCENTRAL-COM_RFC.Feature.StopMaintenance.Enable"
 
 string notifyStatusToString(Maint_notify_status_t &status)
 {
@@ -121,10 +124,10 @@ string moduleStatusToString(IARM_Maint_module_status_t &status)
             ret_status="MAINTENANCE_PINGTELEMETRY_ERROR";
             break;
         case MAINT_FWDOWNLOAD_COMPLETE:
-            ret_status="MAINTENANCE_FWDONLOAD_COMPLETE";
+            ret_status="MAINTENANCE_FWDOWNLOAD_COMPLETE";
             break;
         case MAINT_FWDOWNLOAD_ERROR:
-            ret_status="MAINTENANCE_FWDONLOAD_ERROR";
+            ret_status="MAINTENANCE_FWDOWNLOAD_ERROR";
             break;
         case MAINT_REBOOT_REQUIRED:
             ret_status="MAINTENANCE_REBOOT_REQUIRED";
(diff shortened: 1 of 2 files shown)


mhughesacn:
Hi @livinsunny519
MaintenanceManager/MaintenanceManager.cpp:findProcessPid is not original code.  I think the original source is http://proswdev.blogspot.com/2012/02/get-process-id-by-name-in-linux-using-c.html and this is a problem as there is no permission to use (open to correction).  Comcast Legal has prevented such code from being used before.  The code is a straight copy and must be rewritten.  However, this is a pretty standard way to read the proc file, so it is acknowledged that rewritten code will have to do similar actions.  But comments and variable names must not be identical, or layout as much as possible.
Thanks,
Martin Hughes RDK CMF Compliance Team

livinsunny519:
This is just for checking the code logic.. This will be removed.

mhughesacn:
Thanks Sunny

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -874,6 +888,163 @@ namespace WPEFramework {
...
+            int k_ret=EINVAL;
+            int i=0,record=-1;
+            int32_t exec_status=E_NOK;
+
+            bool task_status[4]={false};
+            bool result=false;
+            bool task_incomplete=false;
+
+            /* only unsolicited maintenance we honour this call & based on RFC */
+            if( checkAbortFlag() && (UNSOLICITED_MAINTENANCE == g_maintenance_type )){
```
livinsunny519, line 913: need to be removed. Only RFC check is fine.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -874,6 +888,163 @@ namespace WPEFramework {
...
+                            record = i;
+                            LOGINFO("Checking the PID\n");
+                            pid_num=getTaskPID(script_names[i].c_str());
+                            LOGINFO("PID of script_name [%d] = %s is %d \n", i,script_names[i].c_str(),pid_num);
+                            if( pid_num != -1){
+                                /* send the signal to task to terminate */
+                                k_ret=kill(pid_num,SIGTERM);
+                                if (k_ret == 0){
+                                    LOGINFO(" %s Termimated\n",script_names[i].c_str());
+                                    /*this means we killed the task currently running */
```
livinsunny519, line 945: break the loop if we are done with the PID. No need to continue.

livinsunny519 (commented):
addressed

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -874,6 +885,166 @@ namespace WPEFramework {
...
+                        LOGINFO("task status [%d]  = %s ScriptName %s",i,(task_status[i])? "true":"false",script_names[i].c_str());
+                    for (i=0;i<4;i++){
+                        if(task_status[i]){
+                            record = i;
+                            LOGINFO("Checking the Task PID\n");
+                            pid_num=getTaskPID(script_names[i].c_str());
+                            LOGINFO("PID of script_name [%d] = %s is %d \n", i,script_names[i].c_str(),pid_num);
+                            if( pid_num != -1){
+                                /* send the signal to task to terminate */
+                                k_ret=kill(pid_num,SIGABRT);
```
anand-ky, line 939: Should this be a SIGABRT or SIGTERM? Is this getting trapped by the scripts for graceful exit?

anand-ky (changes requested):

MaintenanceManager/MaintenanceManager.cpp
livinsunny519: SIGTERM is getting ignored in the case SOLICITED maintenance
Due to this issue, we moved to SIGABRT
all scripts have signal handlers and exiting gracefully.

anand-ky: SIGABRT handled as part of https://gerrit.teamccp.com/#/q/topic:RDK-33288_2105_sprint+(status:open+OR+status:merged)

anand-ky (approved):

PR #1674, opened 2021-08-04
StateObserver: fix memory leaks

michal-kopczynski-red:
Every cJSON_Parse() should have matching cJSON_Delete().


anand-ky:

PR #1728, opened 2021-08-20
RDKTV-10041: HDMI Compatibility Mode API is not behaving as expected …

dhivya-ilangovan:
RDKTV-10041: HDMI Compatibility Mode API is not behaving as expected on setEdid API Method calls

Reason for change: ThunderJS uses string version to identify the version of plugin. Which causing
setEdidVersion api is getting failed during integration.
Hence modifying the param name of setEdidVersion from version to edidVersion.
Test Procedure: Verified get/set curl commands
Risks: Low

### DIFF
--- a/HdmiInput/HdmiInput.cpp
+++ b/HdmiInput/HdmiInput.cpp
@@ -749,9 +749,9 @@ namespace WPEFramework
 
             LOGINFOMETHOD();
             returnIfParamNotFound(parameters, "portId");
-            returnIfParamNotFound(parameters, "version");
+            returnIfParamNotFound(parameters, "edidVersion");
             string sPortId = parameters["portId"].String();
-            string sVersion = parameters["version"].String();
+            string sVersion = parameters["edidVersion"].String();
             try {
                 portId = stoi(sPortId);
             }catch (const device::Exception& err) {


HdmiInput/HdmiInput.cpp
```diff
@@ -749,9 +749,9 @@ namespace WPEFramework
 
             LOGINFOMETHOD();
             returnIfParamNotFound(parameters, "portId");
-            returnIfParamNotFound(parameters, "version");
```
anand-ky, line 752: Pls don't c hange the existing parameter since this will break existing clients. You can add another parameter edidVersion with the same info.

anand-ky (changes requested):

HdmiInput/HdmiInput.cpp
anand-ky: Since this is newly added API, this change looks good. Pls update documentation for these APis as discussed.

anand-ky:
@apatel859 - Pls help update documentation by changing HDMIInput.json and generating https://github.com/rdkcentral/rdkservices/blob/sprint/2105/HdmiInput/doc/HdmiInputPlugin.md

anand-ky (approved):

PR #1749, opened 2021-08-27
Kill unresposinve suspended web process faster

emutavchi:
We want to recover faster from possible hangs.

### DIFF
--- a/DisplaySettings/DisplaySettings.cpp
+++ b/DisplaySettings/DisplaySettings.cpp
@@ -1228,7 +1228,6 @@ namespace WPEFramework {
                 if (aPort.isConnected())
                 {
                     mode = aPort.getStereoMode();
-
                     if (aPort.getType().getId() == device::AudioOutputPortType::kHDMI)
                     {
                         /* In DS5, "Surround" implies "Auto" */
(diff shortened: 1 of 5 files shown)


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2440,6 +2442,27 @@ static GSourceFuncs _handlerIntervention =
...
+            if (!isWebProcessResponsive && _state == PluginHost::IStateControl::SUSPENDED)
+            {
+                SYSLOG(Logging::Notification, (_T("Killing unresponsive suspended WebProcess, pid=%u, reply num=%d(max=%d), url=%s\n"),
+                                            webprocessPID, _unresponsiveReplyNum, kWebProcessUnresponsiveReplyDefaultLimit,
+                                            activeURL.c_str()));
+                if (_unresponsiveReplyNum <= kWebProcessUnresponsiveReplyDefaultLimit)
+                {
+                    _unresponsiveReplyNum = kWebProcessUnresponsiveReplyDefaultLimit;
+                    Logging::DumpSystemFiles(webprocessPID);
+                    if (syscall(__NR_tgkill, webprocessPID, webprocessPID, SIGFPE) == -1)
```
MFransen69, line 2454: Please use the core functionality for this. Create a ProcessInfo with the webprocessPID and call Dump on the object.

MFransen69 (changes requested):

WebKitBrowser/WebKitImplementation.cpp
emutavchi: @MFransen69 ProcessInfo::Dump doesn't work as desired in this case. It uses kill which sends signal to a process ground( any thread can handle that signal). The goal here is to send SIGFPE to a UI thread, hence the use of tgkill.

MFransen69 (approved):

HaseenaSainul (approved):

PR #1804, opened 2021-09-14
Broadcom implementation for DeviceIdentification plugin

rmaddali991:


mhughesacn:
Hi Rajesh: rdkservices is an opensourced component.  The file Broadcom.cpp contains a confidential Broadcom header and a confidential RDK header, neither of which is consistent with opensourcing.  You would need an RDK Apache header on the file for it to scan without errors.  However, before you remove the existing headers, you must be certain that the code is not confidential to Broadcom, and that they have granted full permission for it to be opensourced.

PR #1805, opened 2021-09-14
RDKTV-10975: Return updated Downloaded FW Version

anand-ky:
Return Downloaded Firmware Version based on IARM Firmware Update State instead of always reading from /opt/fwdnldstatus.txt.
If Firmware Update State is Downloading or above then return DnldVersion from /opt/fwdnldstatus.txt else return empty


sujeeshtp (approved):
Looks good.

PR #1824, opened 2021-09-21
Monitor: Syncup with Metro plugin changes

HaseenaSainul:
1. CONFIG variable updates based on proper grouping
2. Added backward compatibility for older configs
3. Warnings removed


anand-ky:

PR #1827, opened 2021-09-21
RDK-30226 : Updated doc for onFirmwareDownloadProgress event

sborushevsky:


anand-ky (changes requested):
No, we cannot change the doc file directly. You have to change the json file and generate doc file

PR #1861, opened 2021-10-05
Extend OCIContainer to support crypted containers.

marcin-hajkowski-epam:
Gives possibility to link omi component which decrypts, mounts and unmounts encrypted bundle according to ocicontainer requests. If omi library has not been found local stub files are being used (no impact on existing functionality).


mhughesacn:
Hi Marcin,
This component is assigned to RDK, so the copyright/header should be RDK Apache on the two files omi_proxy.hpp and i_omi_proxy.hpp.
Thanks,

PR #1878, opened 2021-10-11
ONEM-22659: OCDM new Deinitialize flow.

mikolaj-staworzynski-red:
(cherry picked from commit 9c908eb3b440ea986565bd99a0eb52277de5167a)

### DIFF
--- a/OpenCDMi/OCDM.cpp
+++ b/OpenCDMi/OCDM.cpp
@@ -144,23 +144,8 @@ namespace Plugin {
         _memory->Release();
 
         _opencdmi->Deinitialize(service);
-
-        if (_opencdmi->Release() != Core::ERROR_DESTRUCTION_SUCCEEDED) {
-
-            ASSERT(_connectionId != 0);
-
-            TRACE(Trace::Information, (_T("OCDM Plugin is not properly destructed. %d"), _connectionId));
-
-            RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
-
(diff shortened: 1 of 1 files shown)


OpenCDMi/OCDM.cpp
```diff
@@ -171,6 +156,11 @@ namespace Plugin {
             subSystem->Set(PluginHost::ISubSystem::NOT_DECRYPTION, nullptr);
             subSystem->Release();
         }
+        RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
```
pwielders, line 159: Suggest to first get the RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId)); before you do the final Release() as the release might give you a race condition, whteher or not you can get the interface and thus, if you do not get it, stop of the container is still being called but not in the context of this thread.. So suggest:
`
    RPC::IRemoteConnection*` connection(_service->RemoteConnection(_connectionId));

    uint32_t result = _opencdmi->Release();
    ASSERT(result == Core::ERROR_DESTRUCTION_SUCCEEDED);

    PluginHost::ISubSystem* subSystem = service->SubSystems();

    subSystem->Set(PluginHost::ISubSystem::NOT_DECRYPTION, nullptr);
            subSystem->Release();
    }
    if (connection != nullptr) {
            connection->Terminate();
            connection->Release();
    }
`

pwielders (changes requested):
Change the order of aquiring, as this might potentially be a race conditin.

PR #1911, opened 2021-10-18
RDK-31480: RDM/Packager Plugin Support for Thunder Plugins

PaviP98:
Packager plugin should be capable of downloading and installing a thunder plugin using RDM framework.
Default startup state of Plugin depends on Packager Plugin will be Unavailable.
Packager plugin should be capable of changing the state of downloaded thunder plugin (Unavailable -> Deactivated) after download.


tdeva14 (commented):
Changes look good

Packager/PackagerImplementation.cpp
```diff
@@ -26,6 +26,10 @@
 #endif
 #include <opkg_download.h>
 
+#include <cjson/cJSON.h>
```
pwielders, line 29: For clear visibility on external dependencies, Thunder development rule, put external header files (encompassed by the '<>' characters,) in the Module.h file. This way dependencies of a plugin can be easilly tracked.

helpers/utils.cpp
```diff
@@ -321,6 +321,32 @@ bool Utils::isPluginActivated(const char* callSign)
...
+            LOGWARN("Getting status for callSign %s, status: %d", callSign, status);
+        }
+
+        if(!pluginUnavailable) {
+            LOGWARN("Plugin %s is not in unavailable state", callSign);
+        } else {
+            LOGINFO("Plugin %s is in unavailable state", callSign);
+        }
+        return pluginUnavailable;
+    }
```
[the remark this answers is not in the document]
pwielders: Using an empty string as callsign (I guess callsign = nullptr) in the QueryIinterfaceByCallsign() will give you the IShell* of the Controller.

helpers/utils.h
```diff
@@ -346,6 +346,8 @@ namespace Utils
 
     bool isPluginActivated(const char* callSign);
 
+    bool isPluginUnavailable(const char* callSign);
```
pwielders, line 349: Interface (method definitions) are so important for developpers who consume it. 
This definition to me implies 2 things. 
1) If it is a static method, feels like legacy C code, there you expect static methods. Also what would happen if I feed it a nullptr, whats the use case for that. Since it does not say that it is expecteing an array of characters but a pointer to characters, I assume the nullptr has a meaning, but what ?
2) If it is part of an object, it feels as C++ code. But than why is this function not const? Is there instance object changes if I call this method? Or did they just forget the const at the end of the method. Would the signature actually be: bool IsPluginUnavailable(const TCHAR callsign[]) const; ?

pwielders (dismissed):

karuna2git (commented):
Please follow the coding guidelines regarding the alignment..

anand-ky (approved):

PR #1973, opened 2021-11-08
DELIA-40282 : Cec_devlist return NULL when daemon not running

acheri988:
Reason for change: DELIA-53130
Cec_devlist return NULL when daemon not running
Test Procedure: None
Risks: Low

Change-Id: If7d8f62d27323158b3cc1f24ee02698c3cef9d8b
Signed-off-by: Anooj Cheriyan <Anooj_Cheriyan@comcast.com>

### DIFF
--- a/HdmiCec/HdmiCec.cpp
+++ b/HdmiCec/HdmiCec.cpp
@@ -785,6 +785,13 @@ namespace WPEFramework
 					_instance->smConnection, _instance->logicalAddress);
 			return isConnected;
 		}
+		RFC_ParamData_t param;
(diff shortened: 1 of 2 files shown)


HdmiCec/HdmiCec.cpp
```diff
@@ -785,6 +785,13 @@ namespace WPEFramework
 					_instance->smConnection, _instance->logicalAddress);
 			return isConnected;
 		}
+		RFC_ParamData_t param;
```
apatel859, line 788: Should not we check if CEC enabled instead of RFC flag

apatel859 (changes requested):

acheri988:
New one created: https://github.com/rdkcentral/rdkservices/pull/1976

PR #1996, opened 2021-11-16
RDKTV-11877: sync timeZoneDST and TZ

npoltorapavlo:
Reason for change: TZ should be
timeZoneDST (if exists), otherwise real time zone.
setTimeZoneDST should set TZ.
Test Procedure: Refer to the ticket.
Risks: Low

### DIFF
--- a/LocationSync/LocationSync.cpp
+++ b/LocationSync/LocationSync.cpp
@@ -28,7 +28,7 @@ namespace Plugin {
     static Core::ProxyPoolType<Web::JSONBodyType<LocationSync::Data>> jsonResponseFactory(4);
 
     namespace {
-        string readTextFile(const string &filename) {
+        string ReadTextFile(const string &filename) {
             string result;
 
             Core::File file(filename);
@@ -72,7 +72,7 @@ namespace Plugin {
         string version = service->Version();
 
         if (config.TimeZoneOverrideFile.IsSet() == true) {
-            auto timeZone = readTextFile(config.TimeZoneOverrideFile.Value());
+            auto timeZone = ReadTextFile(config.TimeZoneOverrideFile.Value());
   
             if (timeZone.empty() == false) {
                 Core::SystemInfo::SetEnvironment(_T("TZ"), timeZone);
--- a/LocationSync/LocationSync.cpp
+++ b/LocationSync/LocationSync.cpp
@@ -27,6 +27,25 @@ namespace Plugin {
     static Core::ProxyPoolType<Web::Response> responseFactory(4);
     static Core::ProxyPoolType<Web::JSONBodyType<LocationSync::Data>> jsonResponseFactory(4);
 
+    namespace {
+        const string TZ_FILE = "/opt/persistent/timeZoneDST";
+
+        string readTextFile(const string &filename) {
+            string result;
+
+            Core::File file(filename);
+
+            if (file.Open(true)) {
+                Core::JSON::String str(false);
+                if (str.IElement::FromFile(file)) {
+                    result = str.Value();
+                }
+            }
+
+            return result;
+        }
+    }
+
 #ifdef __WINDOWS__
 #pragma warning(disable : 4355)
 #endif
@@ -53,6 +72,12 @@ namespace Plugin {
         Config config;
         config.FromString(service->ConfigLine());
         string version = service->Version();
+
+        auto timeZone = readTextFile(TZ_FILE);
+
+        if (timeZone.empty() == false) {
+            Core::SystemInfo::SetEnvironment(_T("TZ"), timeZone);
+        }
 
         if (LocationService::IsSupported(config.Source.Value()) == Core::ERROR_NONE) {
             _skipURL = static_cast<uint16_t>(service->WebPrefix().length());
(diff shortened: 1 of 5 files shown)


anand-ky:
@pwielders - Pls check and review LocationSync changes. 
Basically changes done to honor the timezone set by the Resident App (through System Plugin setTimeZoneDST())  and not auto set TZ within LocationSync. This will ensure that the timezone set by user is honored by WPE browser.

LocationSync/LocationSync.cpp
```diff
@@ -27,6 +27,25 @@ namespace Plugin {
     static Core::ProxyPoolType<Web::Response> responseFactory(4);
     static Core::ProxyPoolType<Web::JSONBodyType<LocationSync::Data>> jsonResponseFactory(4);
 
+    namespace {
+        const string TZ_FILE = "/opt/persistent/timeZoneDST";
```
pwielders, line 31: This looks like a magicals. Its good programming practice to avoid magicals and I think to achieve backwards compatibility, you can easily avoid it if you put it in the plugin config, see next remark.

LocationSync/LocationSync.cpp
```diff
@@ -27,6 +27,25 @@ namespace Plugin {
...
+
+        string readTextFile(const string &filename) {
```
pwielders, line 33: As you might have experienced, the Metro style is using PascalCase method naming, so this should be called ReadTextFile

LocationSync/LocationSync.cpp
```diff
@@ -54,6 +73,12 @@ namespace Plugin {
         config.FromString(service->ConfigLine());
         string version = service->Version();
 
+        auto timeZone = readTextFile(TZ_FILE);
```
pwielders, line 76: Let please preserve backwards compatibility. If I read the code I think what you would like to do is overrule the TZ environment variable if it is stored in a file. 
So to make such a feature backwards compatible I suggest adding a config parameter here: https://github.com/rdkcentral/rdkservices/blob/sprint/2107/LocationSync/LocationSync.h#L157 and call it override. In the override you can specify a path (which helps you get ride of the magical, see comments above). 
Now suppose that the config parameter Override.IsSet() and it contains a real filename, only than you try to load that file and you set the environment variable here. This way existing customers will not face havoc if they would place a file at the hardcoded location "/opt/persistent/timeZoneDST"

LocationSync/LocationSync.cpp
```diff
@@ -149,7 +174,12 @@ namespace Plugin {
             subSystem->Release();
 
             if ((_sink.Location() != nullptr) && (_sink.Location()->TimeZone().empty() == false)) {
-                Core::SystemInfo::SetEnvironment(_T("TZ"), _sink.Location()->TimeZone());
+                auto timeZone = readTextFile(TZ_FILE);
```
pwielders, line 177: Here you could evaluate the the config parameter again and only if it is not set, set the Core::SystemInfo::SetEnvironment(_T("TZ"), _sink.Location()->TimeZone());
Currently it might be a bit confusing. Seems if the file is, at the moment of running, empty, the timezone retrieved is set, if it was not empty, the set one is set. THis makes the functionality very time depending and thus hard to predict. Suggest to not set if the override is set to create a consistent behaviour over time.

pwielders (changes requested):
Overal: I think the feature to override the TZ set by the Location sync could better be implemented by setting a override TimeZone in the config. Not a file location in the config, but the required timezone in the config to override. Setting configs for a plugin is a generic feature and already available in the Controller, so no development work needed and probably already used for several plugins. This way the Resident App could directly (or the SystemSetting plugin) set the override TimeZone in the LocationSync info and you are done. without using a persitent storage location (flash wear out ?)
Bottomline, prefer that at least the filename needs to be set from the config (can be set build time) so that this feature does not break our customers deployments  but suggest you consider the timezone taking from the config i.s.o. the file and set it directly from ResidentApp/SystemSettings plugin.

npoltorapavlo:
addressed the review comments. Also tried the suggested solution to store the time zone value in config but observed the value doesn't persist after reboot.

pwielders:
"addressed the review comments. Also tried the suggested solution to store the time zone value in config but observed the value doesn't persist after reboot (and the plugin should be deactivated to set config)." Do not know how I can respond to a remark, so copied you're request.
By default Thunder or its plugins do not enforce a behaviour that "might" be unwanted. So changing the config and persisting it is a two step approach. Changing the config is the first step. If the domain logic want it to persist, you have to tell the controller to persist (storeconfig) the config. This way changes can be implemented as volatile (do not store) or as persitent (do store).

So if you like the plugin to persist the changes, please call storeconfig (https://github.com/rdkcentral/Thunder/blob/master/Source/WPEFramework/doc/ControllerPlugin.md#method.storeconfig)

The reason the plugin has to be "deactivated/activated" is since the pre-requisites (the configuration wich is an crucial part of the workings of the plugin) have changed, so this requires a restart so all logic for the plugin is re-evaluated with the new settings.

If there would be no such re-evaluation required, another option would be to "enhance" the LocationSync interface (like we did for the TimeSync. The TimeSync, has a similar feature, if the time should be coming from the TOT (DVB table supplying the time). We added to the TimeSync plugin a possibility to "overrule" the set time from the DVB, once this time is available. We do not specify that it is a DVB time, we just add an interface to the TImeSync plugin to set the time externally. I think the TimeZOne setting you add here is pretty much similar, althouh you probably want to persist it once it gets set from the outside (which would be possible as well, a lot of plugins persist local plugin information).
See: https://github.com/rdkcentral/ThunderNanoServices/blob/master/TimeSync/doc/TimeSyncPlugin.md#property.time

LocationSync/LocationSync.h
```diff
@@ -151,10 +151,12 @@ namespace Plugin {
                 : Interval(30)
                 , Retries(8)
                 , Source()
+                , TimeZoneOverrideFile()
             {
                 Add(_T("interval"), &Interval);
                 Add(_T("retries"), &Retries);
                 Add(_T("source"), &Source);
+                Add(_T("timeZoneOverrideFile"), &TimeZoneOverrideFile);
```
pwielders, line 159: field names in JSON for the config are typically all lowercase. suggest calling this timezonefile ?

PR #1998, opened 2021-11-17
Enable Intersection Observer object by default

nrajan002c:
Signed-off-by: nganes961 <nambirajan_ganesan@comcast.com>

### DIFF
--- a/WebKitBrowser/WebKitImplementation.cpp
+++ b/WebKitBrowser/WebKitImplementation.cpp
@@ -2275,6 +2275,9 @@ static GSourceFuncs _handlerIntervention =
             // Turn on/off local storage
             WKPreferencesSetLocalStorageEnabled(preferences, _localStorageEnabled);
 
+            // Turn on Intersection Observer object
+            WKPreferencesSetIntersectionObserverEnabled(preferences, true);
+
             // Media Content Types Requiring Hardware Support
             if (_config.MediaContentTypesRequiringHardwareSupport.IsSet() == true
                 && _config.MediaContentTypesRequiringHardwareSupport.Value().empty() == false) {


emutavchi (commented):
LGTM, need someone from Metro to review as well

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2275,6 +2275,9 @@ static GSourceFuncs _handlerIntervention =
             // Turn on/off local storage
             WKPreferencesSetLocalStorageEnabled(preferences, _localStorageEnabled);
 
+            // Turn on Intersection Observer object
```
MFransen69, line 2278: Could you please only enable this for WebKit 2.28 or later (#ifdef)?

Indeed there seems to be no side effect to turn it on, so it does not have to be configurable, but it is only implemented in 2.28 and as the plugin also needs to work with older webkit versions (it seems to be implemented empty in 2.22 but might not be available at all in older versions) better to have it only enabled on 2.28 or later.

MFransen69 (changes requested):

WebKitBrowser/WebKitImplementation.cpp
nrajan002c: Hi @MFransen69 
We are currently requiring this support in 2.22 for one of the premium app. 
We had similar issue with other observer  (ResizeObserver) support was not there in 2.22. We had brought to this branch. 
Could you please share the IntersectionObserver changes, we can add this to 2.22 to make it clear.

MFransen69: Could you please reach out to Özgür Deveci for that (@modeveci)? Think then still you would need to conditionally add the call here for older WebKit versions (like 2017) but that you can also ask him. Thanks!

PR #2015, opened 2021-11-25
Updated SecurityAgent documents as per code

pbhatt655:
Updated the SecurityAgent.json and SecurityAgentPlugin.md documents as per the latest code

Jira : https://ccp.sys.comcast.net/browse/RDK-34513

### DIFF
--- a/SecurityAgent/SecurityAgent.json
+++ b/SecurityAgent/SecurityAgent.json
@@ -216,7 +216,7 @@
     },
     "methods": {
         "createtoken": {
-            "summary": "Creates a signed JsonWeb token",
+            "summary": "Creates a signed JsonWeb token. On success, returns Signed JsonWeb token and on failure, returns error message and error code as mentioned in below Errors table.",
             "params": {
                 "type": "object",
                 "properties": {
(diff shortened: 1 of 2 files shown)


SecurityAgent/SecurityAgent.json
```diff
@@ -247,7 +246,7 @@
                 "type": "object",
                 "properties": {
                     "token": {
-                        "description": "Signed JsonWeb token",
+                        "description": "On success, returns Signed JsonWeb token. On failure, returns error message and error code from the following table.",
```
npoltorapavlo, line 249: this "description" field is for the "token" property of "result". Neither this property nor "result" return "error message and error code". On error, "error" object is set instead of "result". Explanation of what the method returns should probably be in the method description.

npoltorapavlo (changes requested):

SecurityAgent/SecurityAgent.json
pbhatt655: Modified the document as per your suggestion

anand-ky (approved):

PR #2020, opened 2021-11-29
Fix typos

jruzzi:
There were two event names in the API docs that did not match their implementation. The API docs were updated to match the implementation.

Please include these changes in main as well.


DisplaySettings/doc/DisplaySettingsPlugin.md
```diff
@@ -3408,58 +3408,8 @@ Possible values:
 }
 ```
 
-<a name="method.setSoundMode"></a>
```
anand-ky, line 3411: Was this intended to be removed? There is no setSoundMode description now with the new changes.

jruzzi: good catch. This commit was not intended to be part of this pull request. I'm closing the pull request and have opened #2029 which removes the stale commit.

PR #2039, opened 2021-12-07
RDKTV-12808: WebKitBrowser Preconditions

TeknoVenus:
Allow configuring preconditions with CMake flags.

This is to prevent containers from failing to start due to OCDM socket not being present by adding `Decryption` precondition on some platforms. This ensures we wait until OCDM is ready so that containers will start without error.


emutavchi (commented):
I don't see why is this required. If you need to add more to preconditions, let's discuss. I'd prefer preconditions to be explicitly set in the config file instead of having a configuration option.

anand-ky:
Closing older PRs. Please rebase and submit on latest sprint branch.

PR #2044, opened 2021-12-09
RDK-34677: Thunder implementation for getMfgSerialNumber

dhivya-ilangovan:
Reason for change: added new API getMfgSerialNumber
Test Procedure: Verified get curl command
curl command: curl -H "Authorization: Bearer $TOKEN_tmp" -d '{"jsonrpc":"2.0","id":"3","method": "org.rdk.System.1.getMfgSerialNumber","params":{}}' http://127.0.0.1:9998/jsonrpc
expected result: {"jsonrpc":"2.0","id":3,"result":{"mfgSerialNumber":"F000209C000177B10","success":true}}
Risks: Low

### DIFF
--- a/SystemServices/System.json
+++ b/SystemServices/System.json
@@ -484,6 +484,26 @@
                 ]
             }
         },
+        "getMfgSerialNumber": {
+            "summary": "Gets the Manufacturing Serial Number",
+            "result": {
+                "type": "object",
+                "properties": {
+                    "mfgSerialNumber": {
+                        "summary": "Manufacturing Serial Number",
+                        "type": "string",
+                        "example": "F00020CE000003"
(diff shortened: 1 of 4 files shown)


anand-ky:
@dhivya-ilangovan - Pls merge the HAL changes before merging this PR. Also is this value different from getSerialNumber that is currently supported? Can you help share both values from a llama device.

dhivya-ilangovan:
> 

Hi Anand, 
getSerialNumber is for device serial number which is different from Manufacture serial number. These are the two different values stored in two different locations of Flash.
Also This Manufature Serial number is perticularly for LAMA DE/IT and Alpaca DE/IT. Hence needed separate API

PR #2051, opened 2021-12-10
Manufacturing Serial Number support Sky Glass DE/IT

manjunathtsai:


manjunathtsai:
need to create it for sprin 2107

PR #2067, opened 2021-12-20
RDKTV-13125: WPEFramework timeout during shutdown

livinsunny519:
Reason for change:Due to maintenance running the task seems to be
blocked.
Test Procedure:Refer Ticket.
Risks: Low

### DIFF
--- a/MaintenanceManager/MaintenanceManager.cpp
+++ b/MaintenanceManager/MaintenanceManager.cpp
@@ -237,14 +237,16 @@ namespace WPEFramework {
             std::unique_lock<std::mutex> lck(m_callMutex);
             if (UNSOLICITED_MAINTENANCE == g_maintenance_type && internetConnectStatus){
                 LOGINFO("---------------UNSOLICITED_MAINTENANCE--------------");
-                for ( i=0;i< task_count ;i++ ){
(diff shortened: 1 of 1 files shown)


livinsunny519:
Developer Testing in progress. Do not merge.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -237,16 +237,16 @@ namespace WPEFramework {
...
+                    LOGINFO("DBG After Lock");
                     cmd=task_names_foreground[i].c_str();
                     cmd+=" &";
                     cmd+="\0";
                     m_task_map[task_names_foreground[i].c_str()]=true;
                     LOGINFO("Starting Script (USM) :  %s \n", cmd.c_str());
-                    if (!m_abort_flag){
-                        system(cmd.c_str());
-                    }
+                    system(cmd.c_str());
```
npoltorapavlo, line 249: is it expected that the script will start when m_abort_flag is true?

livinsunny519: I have modified it. Please check

npoltorapavlo: looks good

livinsunny519:
Created https://github.com/rdkcentral/rdkservices/pull/2077

PR #2096, opened 2022-01-10
LLAMA-4597:Bring back the Network Check

livinsunny519:
Reason for change:Periodic maintenance lasts 1'32''
when eth0 interface established, but router has not internet connection.
Test Procedure:Refer Ticket.
Risk: Medium

### DIFF
--- a/MaintenanceManager/MaintenanceManager.cpp
+++ b/MaintenanceManager/MaintenanceManager.cpp
@@ -222,11 +222,7 @@ namespace WPEFramework {
 
             LOGINFO("Reboot_Pending :%s",g_is_reboot_pending.c_str());
 
-#if defined (SKY_BUILD)
-            bool internetConnectStatus = true;
-#else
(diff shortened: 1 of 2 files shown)


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -222,11 +222,7 @@ namespace WPEFramework {
 
             LOGINFO("Reboot_Pending :%s",g_is_reboot_pending.c_str());
 
-#if defined (SKY_BUILD)
-            bool internetConnectStatus = true;
-#else
             bool internetConnectStatus = isDeviceOnline();
```
anand-ky, line 225: Pls increase timeout for isDeviceOnline to 2 min (4 checks every 30 seconds)

livinsunny519: added the retry logic.

anand-ky (approved):

PR #2098, opened 2022-01-10
RDK-34750: add getPublicIP method to Network plugin

tabbas651:
Reason for change: XRE wants to obtain the public facing wan ip
Also fixed a few compiler warnings.
Test Procedure: json-rpc curl commands from command line on both desktop and device.  later xre will test.
Risks: Low
Signed-off-by: Thamim Razith <ThamimRazith_AbbasAli@comcast.com>

### DIFF
--- a/Network/Network.cpp
+++ b/Network/Network.cpp
@@ -64,7 +64,7 @@ typedef struct _IARM_BUS_NetSrvMgr_Iface_EventData_t {
     union {
         char activeIface[INTERFACE_SIZE];
         char allNetworkInterfaces[INTERFACE_LIST];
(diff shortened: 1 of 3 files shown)


karuna2git (changes requested):
why "enableInterface" is changed to "SetInterface" ??
Also, this is not scope of this changeset?

tabbas651:
> why "enableInterface" is changed to "SetInterface" ?? Also, this is not scope of this changeset?
 yes karuna. it's revert back to enableInterface. and updated in second commit .please review it

tabbas651:
it is not required

PR #2105, opened 2022-01-12
Warehouse json & md files updated as per code

s-keerthi:
Documentation enhancement for Warehouse plugin
Jira link: https://ccp.sys.comcast.net/browse/RDK-35075

### DIFF
--- a/Warehouse/Warehouse.json
+++ b/Warehouse/Warehouse.json
@@ -12,11 +12,6 @@
             "type": "string",
             "example": ""
         },
-        "PARAM_SUCCESS": {
-            "summary": "Whether the parameter succeeded",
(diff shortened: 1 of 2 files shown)


npoltorapavlo:
@s-keerthi please keep the "error" property. It is part of the "result" object in case of failure so should not be removed from the docs.

s-keerthi:
> @s-keerthi please keep the "error" property. It is part of the "result" object in case of failure so should not be removed from the docs.

@npoltorapavlo Your suggestion has been addressed and "error" changes were reverted back. Thank you

npoltorapavlo:
looks good

anand-ky (approved):

PR #2121, opened 2022-01-13
ONEMPERS-287 XCast: retry reconnection till success

dawid-trendota-red:


michalkucharczyk (dismissed):

anand-ky:
@apatel859 - Pls review.

XCast/XCast.cpp
```diff
@@ -373,7 +373,10 @@ void XCast::onLocateCastTimer()
     int status = _rtConnector->connectToRemoteService();
     if(status != 0)
     {
-        locateCastObjectRetryCount++;
+        if(locateCastObjectRetryCount < 4)
```
apatel859, line 376: instead of making this change just update 

        if(locateCastObjectRetryCount == 4)
condition to 
        if(locateCastObjectRetryCount >= 4)
 
so in case if we print locateCastObjectRetryCount we know after how many attempt it got connection success

dawid-trendota-red: This approach might introduce an undefined behavior in a (very unlikely) scenario of integer overflow. Is it worth the possibility to track how many connection attempts were done?

dawid-trendota-red: @apatel859 could you answer whether you agree with such risk of integer overflow?

michalkucharczyk: @dawid-trendota-red 
I would lean towards this conditions (mentioned by @apatel859 ):
if(locateCastObjectRetryCount >= 4)

If you believe overflowing locateCastObjectRetryCount is a real problem (rather unlikely, as most modern platforms use 32bit int), please change the declarion of locateCastObjectRetryCount to uint64_t.

michalkucharczyk (approved):

apatel859 (approved):

apatel859 (approved):

anand-ky (approved):

PR #2165, opened 2022-01-25
XCast Json and md files updated as per code

pjayar213:
Documentation enhancement for XCast plugin
Jira ticket link: https://ccp.sys.comcast.net/browse/RDK-35237

### DIFF
--- a/XCast/XCast.json
+++ b/XCast/XCast.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "../../../Tools/JsonGenerator/schemas/interface.schema.json",
+    "$schema": "https://raw.githubusercontent.com/rdkcentral/Thunder/master/Tools/JsonGenerator/schemas/interface.schema.json",
     "jsonrpc": "2.0",
     "info": {
         "title": "XCast API", 
@@ -56,7 +56,7 @@
     },
     "methods": {
         "getApiVersionNumber":{
-            "summary": "Gets the API version",
+            "summary": "Gets the API version.\n  \n### Events \n\n No Events",
             "result": {
                 "type":"object",
                 "properties": {
@@ -76,7 +76,7 @@
             }
         },
         "getEnabled":{
-            "summary": "Reports whether xcast is enabled or disabled",
+            "summary": "Reports whether xcast is enabled or disabled.\n  \n### Events \n\n No Events",
             "result": {
                 "type":"object",
                 "properties": {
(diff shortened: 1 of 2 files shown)


XCast/XCast.json
```diff
@@ -161,7 +161,7 @@
             }
         },
         "registerApplications": {
-            "summary": "Registers an application",
+            "summary": "Registers an application.\n  \n### Events \n\n No Events",
```
apatel859, line 164: This API allows to whitelist the apps which support dial service. If app want to dynamically update the app list same API should be called with updated list.

XCast/XCast.json
```diff
@@ -197,7 +197,7 @@
             }
         },
         "setFriendlyName":{
-            "summary": "Sets a name for the implementation",
+            "summary": "Sets a name for the implementation.\n  \n### Events \n\n No Events",
```
apatel859, line 200: This API allows application to override the default friendly name value with the friendly name  passed as an argument.

XCast/XCast.json
```diff
@@ -94,7 +94,7 @@
             }
         },
         "getFriendlyName":{
-            "summary": "Gets the human readable name for the implementation",
+            "summary": "Gets the human readable name for the implementation.\n  \n### Events \n\n No Events",
```
apatel859, line 97: Returns the friendly name set by setFriendlyName API.

XCast/XCast.json
```diff
@@ -214,7 +214,7 @@
             }
         },
         "setStandbyBehavior":{
-            "summary": "Sets the expected xcast behavior in standby mode",
+            "summary": "Sets the expected xcast behavior in standby mode.\n  \n### Events \n\n No Events",
```
apatel859, line 217: This API allows application to override controls the xcast behavior in standby mode. Default behavior in STANDBY mode is disable so client device can not discover the server. When STANDBY behavior is active client device could discover the server.

XCast/XCast.json
pjayar213: Updated changes as per your review comments

XCast/XCast.json
pjayar213: Updated changes as per your review comments

XCast/XCast.json
pjayar213: Updated changes as per your review comments

XCast/XCast.json
pjayar213: Updated changes as per your review comments

apatel859 (approved):

anand-ky (approved):

PR #2170, opened 2022-01-25
LLAMA-5154:Panel fails to come out of standby

apatel859:
Reason for change: Added fix to handle exceptions.
Test Procedure:Refer JIRA
Risks: low

### DIFF
--- a/DisplaySettings/DisplaySettings.cpp
+++ b/DisplaySettings/DisplaySettings.cpp
@@ -75,6 +75,7 @@ using namespace std;
 
 static bool isCecArcRoutingThreadEnabled = false;
 static bool isCecEnabled = false;
+static std::thread audioPortInitThread;
 
 #ifdef USE_IARM
 namespace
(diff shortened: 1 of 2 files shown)


DisplaySettings/DisplaySettings.cpp
```diff
@@ -75,6 +75,7 @@ using namespace std;
 
 static bool isCecArcRoutingThreadEnabled = false;
 static bool isCecEnabled = false;
+static std::thread audioPortInitThread;
```
anand-ky, line 78: Does this need to be static. Can we make it a member (of type Utils::ThreadRAII)?

DisplaySettings/DisplaySettings.cpp
```diff
@@ -3801,7 +3807,20 @@ namespace WPEFramework {
                              eventData->data.state.curState, eventData->data.state.newState);
                 m_powerState = eventData->data.state.newState;
                 if (eventData->data.state.newState == IARM_BUS_PWRMGR_POWERSTATE_ON) {
-                    DisplaySettings::_instance->InitAudioPorts();
+	            try
+                    {
+		        LOGWARN("creating worker thread for initAudioPortsWorker");
+		        audioPortInitThread = std::thread(initAudioPortsWorker);
+			audioPortInitThread.detach();
```
anand-ky, line 3814: Should this be detached or can we look at joining it during deinitialize()?

anand-ky (changes requested):

DisplaySettings/DisplaySettings.cpp
apatel859: im not using threadRAII as it not serve the desire purpose. i can make audioPortInitThread to meember of displaySetting hhowever i choose this option as i want to use that in callback function which is static function so I have to pass as a data to thread funtion.

PR #2176, opened 2022-01-27
SystemServices json & md files updated as per code

s-keerthi:
Documentation enhancement for SystemServices plugin
Jira Link: https://ccp.sys.comcast.net/browse/RDK-35198

### DIFF
--- a/SystemServices/System.json
+++ b/SystemServices/System.json
@@ -34,8 +34,8 @@
         },
         "temperature": {
             "summary": "The temperature",
-            "type": "number",
-            "example": 48.000000
+            "type": "string",
+            "example": "48.000000"
         },
         "enabled": {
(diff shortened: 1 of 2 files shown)


sborushevsky:
Result of "getLastFirmwareFailureReason" method contains "failReason", not the "failureReason".

"getStoreDemoLink", "fileURL" should be removed from "required", since it depends on STORE_DEMO_FILE existence, otherwise there will be "error".

"getTimeZones", there is "zoneinfo" in the response, not "zoneInfo".

"deletePersistentPath", looks like it takes "callsign" or "type" parameters, but there is "callsign" in "required".

sborushevsky (dismissed):
Looks good to me.

anand-ky (approved):

PR #2186, opened 2022-01-28
RDK-35012: Suppress Maintenance

livinsunny519:
Reason for change: Add code change to suppress maintenace when
device is not activated.
Test Procedure: Refer ticket.
Risk:Medium

### DIFF
--- a/MaintenanceManager/MaintenanceManager.cpp
+++ b/MaintenanceManager/MaintenanceManager.cpp
@@ -170,13 +170,14 @@ namespace WPEFramework {
         MaintenanceManager* MaintenanceManager::_instance = nullptr;
 
         cSettings MaintenanceManager::m_setting(MAINTENANCE_MGR_RECORD_FILE);
-        //TODO  this need to moved to a seperate class and vector based.
 
         string task_names_foreground[]={
             "/lib/rdk/RFCbase.sh",
             "/lib/rdk/swupdate_utility.sh >> /opt/logs/swupdate.log",
             "/lib/rdk/Start_uploadSTBLogs.sh"
         };
+
+        vector<string> tasks;
 
         string script_names[]={
             "DCMscript_maintaince.sh",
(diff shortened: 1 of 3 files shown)


livinsunny519:
Do Not merge the code. This needs to be tested in LLAMA and Placto.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -216,67 +217,234 @@ namespace WPEFramework {
...
+
+            switch(result){
+                case ACTIVATED:
+                    ret_result = true;
+                    break;
+                case ACTIVATION_DISCONNECT:
+                    ret_result = false;
+                    break;
+                default:
+                    ret_result = true;
```
anand-ky, line 439: checking why is ret_result set to true in the default case. Shouldn't it be true only for ACTIVATED?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -216,67 +217,234 @@ namespace WPEFramework {
...
 
+        const string MaintenanceManager::getActivatedStatus()
+        {
+            JsonObject joGetParams;
+            JsonObject joGetResult;
+            std::string callsign = "org.rdk.AuthService.1";
+            std::string token;
+            uint8_t i = 0;
+            bool isAuthSerivcePluginActive = false;
+#if defined(BUILD_ENABLE_NOT_ACTIVATED)
```
anand-ky, line 331: Pls remove these default values. These are internal to getActivationStatus and might change in the future. If we cannot get the activation status after 4 retries, then can we switch to existing behavior and proceed with IsDeviceOnline check.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -216,67 +217,234 @@ namespace WPEFramework {
...
+            Utils::SecurityToken::getSecurityToken(token);
+
+            Core::SystemInfo::SetEnvironment(_T("THUNDER_ACCESS"), _T(SERVER_DETAILS));
+            auto thunder_client = make_shared<WPEFramework::JSONRPC::LinkType<WPEFramework::Core::JSON::IElement> >(callsign.c_str(), "");
+            if (thunder_client != nullptr) {
+                uint32_t status = thunder_client->Invoke<JsonObject, JsonObject>(5000, "getActivationStatus", joGetParams, joGetResult);
+                LOGINFO("Invoke status : %d",status);
+                if (status > 0) {
+                    LOGINFO("%s call failed %d", callsign.c_str(), status);
+#if defined(BUILD_ENABLE_NOT_ACTIVATED)
```
anand-ky, line 372: Pls remove these activation status defaults. These are internal to authService getActivationStatus API.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -216,67 +217,234 @@ namespace WPEFramework {
...
+#if defined(SUPPRESS_MAINTENANCE)
+            bool activationStatus=false;
+            bool skipFirmwareCheck=false;
 
-            MaintenanceManager::_instance->onMaintenanceStatusChange(MAINTENANCE_STARTED);
-            /*  In an unsolicited maintenance we make sure only after
-             *  after DCM task activities are started.
-             */
+            /* Activation check */
+            activationStatus = isDeviceActivated(skipFirmwareCheck);
```
anand-ky, line 235: Pls change isDeviceActivated to getActivationStatus to better reflect the purpose.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -216,67 +217,234 @@ namespace WPEFramework {
...
+                /* Network check */
+                internetConnectStatus = isDeviceOnline();
+            }
+#else
+            internetConnectStatus = isDeviceOnline();
+#endif
+            LOGINFO("Reboot_Pending :%s",g_is_reboot_pending.c_str());
+
+            MaintenanceManager::_instance->onMaintenanceStatusChange(MAINTENANCE_STARTED);
+#if defined(SUPPRESS_MAINTENANCE)
```
anand-ky, line 250: Pls remove this SUPPRESS_MAINTENANCE build flag to simplify the code. If we cannot read activation status from authservice then we can default to existing behavior.

amlogic.cmake
```diff
@@ -91,6 +91,11 @@ if (SKY_BUILD)
     add_definitions (-DSKY_BUILD)
 endif()
 
+if (SUPPRESS_MAINTENANCE)
```
anand-ky, line 94: Pls remove this SUPPRESS_MAINTENANCE build flag to simplify the code. If we cannot read activation status from authservice then we can default to existing behavior.

anand-ky (changes requested):

MaintenanceManager/MaintenanceManager.cpp
livinsunny519: This is specific to Platco & Hisense, so we are restricting this to these platforms based on the distro feature. Hence activation check is only for the aforesaid platforms, rest of the platforms we don't enable it.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -216,67 +217,234 @@ namespace WPEFramework {
...
+                result = ACTIVATION_CONNECT;
+                LOGINFO("result: activation-connect");
+#endif
+            }
+
+            switch(result){
+                case ACTIVATED:
+                    ret_result = true;
+                    break;
+                case ACTIVATION_DISCONNECT:
```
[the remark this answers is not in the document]
livinsunny519: modified.

MaintenanceManager/MaintenanceManager.cpp
livinsunny519: modified.

MaintenanceManager/MaintenanceManager.cpp
livinsunny519: removed.

amlogic.cmake
livinsunny519: This is specific to Platco & Hisense, so we are restricting this to these platforms based on the distro feature. Hence activation check is only for the aforesaid platforms, rest of the platforms we don't enable it.

MaintenanceManager/MaintenanceManager.cpp
livinsunny519: removed.

anand-ky (changes requested):

anand-ky (approved):

PR #2220, opened 2022-02-11
RDK-30296 : VoiceControl Thunder Plugin to OpenSource

jschmi201:
Reason for change: Needs to move from CPC to open source
Test Procedure: Verify with voiceControl test app.
Risks: None.
Signed-off-by: jschmidt <Jeff_Schmidt@cable.comcast>

### DIFF
--- a/VoiceControl/VoiceControl.json
+++ b/VoiceControl/VoiceControl.json
@@ -1,0 +1,9 @@
+{
+ "locator":"libWPEFrameworkVoiceControl.so",
+ "classname":"VoiceControl",
+ "precondition":[
+  "Platform"
+ ],
+ "callsign":"org.rdk.VoiceControl",
+ "autostart":false
+}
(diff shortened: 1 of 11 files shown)


VoiceControl/VoiceControl.json
```diff
@@ -0,0 +1,9 @@
+{
```
anand-ky, line 1: This looks to be the generated plugin config json file (generated from VoiceControl.config) which is incorrect. The VoiceControl.json needs to define the interface for this plugin. Pls refer to https://github.com/rdkcentral/rdkservices/blob/sprint/2201/DisplaySettings/DisplaySettings.json. Also the documentation (VoiceControl.md) needs to be generated from the interface file. Pls refer to https://github.com/rdkcentral/RDKServices#documentation

anand-ky (changes requested):

PR #2224, opened 2022-02-15
DELIA-54683: refactoring

npoltorapavlo:
Reason for change: Simplify writing unit tests
for DataCapture, DeviceDiagnostics and FrameRate
by removing the unnecessary includes.
Test Procedure: It should build.
Risks: Low

### DIFF
--- a/DataCapture/DataCapture.h
+++ b/DataCapture/DataCapture.h
@@ -20,16 +20,13 @@
 #pragma once
 
 #include "Module.h"
-#include "utils.h"
-#include "AbstractPlugin.h"
-#include "libIBus.h"
-//#include "irMgr.h"
+#include "libIARM.h"
 
 class socket_adaptor;
 
 namespace WPEFramework {
     namespace Plugin {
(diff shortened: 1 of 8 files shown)


DataCapture/DataCapture.h
```diff
@@ -20,16 +20,13 @@
...
-#include "libIBus.h"
-//#include "irMgr.h"
+#include "libIARM.h"
 
 class socket_adaptor;
 
 namespace WPEFramework {
     namespace Plugin {
-        class DataCapture : public AbstractPlugin {
+        class DataCapture : public PluginHost::IPlugin, public PluginHost::JSONRPC {
```
anand-ky, line 29: Let's retain AbstractPlugin base class and remove defining the Interface map. Thus plugins can still derive from it and add these own interface map. I would like to retain the helper APIs for registering different versions since the current APIs from thunder are not userfriendly and don't scale up for multiple versions.

anand-ky (changes requested):

PR #2230, opened 2022-02-17
ONEMPERS-306: HdcpProfile: returning local string c_str() corrected

michalkucharczyk:


anand-ky:
@apatel859 - Can you pls check where this getHdcpReasonStr() is being used. Don't see it within RDKServices.

HdcpProfile/HdcpProfile.cpp
```diff
@@ -44,7 +44,7 @@ namespace WPEFramework
     namespace Plugin
     {
         SERVICE_REGISTRATION(HdcpProfile, 1, 0);
-        static const char* getHdcpReasonStr (int eHDCPEnabledStatus);
+        static std::string getHdcpReasonStr (int eHDCPEnabledStatus);
```
apatel859, line 47: not used so remove it everwhere

anand-ky:
@michalkucharczyk - good catch on returning local string. This API is not used. Can you help remove this as part of this PR.

anand-ky (approved):

pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-2859

PR #2233, opened 2022-02-17
RDKTV-14636: WiFi Connectivity lost after reboot

zrash:
Reason for change: Mitigate Llama/Hisense startup issue
by polling NetSrvMgr availability. Also fixes LLAMA-2532.
Test Procedure: Refer to ticket for details
Risks: Medium
Signed-off-by: Zameerun Rasheed M S <zmamoo711@cable.comcast.com>


Network/Network.h
```diff
@@ -54,6 +54,10 @@ namespace WPEFramework {
             Network(const Network&) = delete;
             Network& operator=(const Network&) = delete;
 
+            //Private variables
+            bool isNetworkPluginInitialized;
```
karuna2git, line 58: Can you rename the variable with "m_" prepend..
Something like "m_isPluginReady"

Network/Network.cpp
```diff
@@ -199,7 +200,15 @@ namespace WPEFramework
 
 #ifndef NET_DISABLE_NETSRVMGR_CHECK
                 char c;
-                retVal = IARM_Bus_Call_with_IPCTimeout(IARM_BUS_NM_SRV_MGR_NAME, IARM_BUS_NETSRVMGR_API_isAvailable, (void *)&c, sizeof(c), 1000);
+                uint32_t retry = 0;
+                do{
+                    retVal = IARM_Bus_Call_with_IPCTimeout(IARM_BUS_NM_SRV_MGR_NAME, IARM_BUS_NETSRVMGR_API_isAvailable, (void *)&c, sizeof(c), (1000*10));
```
karuna2git, line 205: IARM_Bus_Call() by default has 5s timeout. Since you are getting into loop, I would recommend just to go with default timeout instead of 10s.

Network/Network.cpp
```diff
@@ -213,12 +222,14 @@ namespace WPEFramework
...
                     IARM_CHECK( IARM_Bus_RegisterEventHandler(IARM_BUS_NM_SRV_MGR_NAME, IARM_BUS_NETWORK_MANAGER_EVENT_DEFAULT_INTERFACE, eventHandler) );
                     LOGINFO("Successfully activated Network Plugin");
+                    isNetworkPluginInitialized = true;
                 }
             }
             else
             {
                 msg = "IARM bus is not available";
                 LOGERR("IARM bus is not available. Failed to activate Network Plugin");
+                retryIarmEventRegistration();
```
karuna2git, line 232: This is WRONG PLACE... 
It should be called when (retVal != IARM_RESULT_SUCCESS) case. 

ie Line#217

Network/Network.cpp
```diff
@@ -199,7 +200,15 @@ namespace WPEFramework
...
+                    retVal = IARM_Bus_Call_with_IPCTimeout(IARM_BUS_NM_SRV_MGR_NAME, IARM_BUS_NETSRVMGR_API_isAvailable, (void *)&c, sizeof(c), (1000*10));
+                    if(retVal != IARM_RESULT_SUCCESS){
+                        LOGERR("NetSrvMgr is not available. Failed to activate Network Plugin");
+                        usleep(500*1000);
+                        retry++;
+                    }
+                }while((retVal != IARM_RESULT_SUCCESS) && (retry < 20));
 #endif
 
                 if(retVal != IARM_RESULT_SUCCESS)
```
karuna2git, line 214: retryIarmEventRegistration() should be called here

karuna2git: Also change the log line here at 268.. Add something like

LOGWARN ("NETWORK_NOT_READY: The netsrvmgr Component is not available. Retrying in separate thread");

Network/Network.cpp
```diff
@@ -310,6 +321,50 @@ namespace WPEFramework
...
+            IARM_Result_t res = IARM_RESULT_SUCCESS;
+            while(isNetworkPluginInitialized != true)
+            {
+                IARM_Result_t retVal = IARM_RESULT_SUCCESS;
+
+#ifndef NET_DISABLE_NETSRVMGR_CHECK
+                char c;
+                uint32_t retry = 0;
+                do{
+                    retVal = IARM_Bus_Call_with_IPCTimeout(IARM_BUS_NM_SRV_MGR_NAME, IARM_BUS_NETSRVMGR_API_isAvailable, (void *)&c, sizeof(c), (1000*10));
```
karuna2git, line 341: Same comment as above.. Use default function

Network/Network.cpp
```diff
@@ -310,6 +321,50 @@ namespace WPEFramework
...
+                        LOGERR("retryIarmEventRegistration: NetSrvMgr is not available. Failed to activate Network Plugin, retrying");
+                        usleep(500*1000);
+                        retry++;
+                    }
+                }while((retVal != IARM_RESULT_SUCCESS) && (retry < 20));
+#endif
+
+                if(retVal != IARM_RESULT_SUCCESS)
+                {
+                    LOGERR("retryIarmEventRegistration: NetSrvMgr is not available. Failed to activate Network Plugin, retrying");
```
karuna2git, line 352: Change the logging.. But do not add "TELEMETRY" here..

Network/Network.cpp
```diff
@@ -310,6 +321,50 @@ namespace WPEFramework
...
+                }
+                else
+                {
+                    IARM_CHECK( IARM_Bus_RegisterEventHandler(IARM_BUS_NM_SRV_MGR_NAME, IARM_BUS_NETWORK_MANAGER_EVENT_INTERFACE_ENABLED_STATUS, eventHandler) );
+                    IARM_CHECK( IARM_Bus_RegisterEventHandler(IARM_BUS_NM_SRV_MGR_NAME, IARM_BUS_NETWORK_MANAGER_EVENT_INTERFACE_CONNECTION_STATUS, eventHandler) );
+                    IARM_CHECK( IARM_Bus_RegisterEventHandler(IARM_BUS_NM_SRV_MGR_NAME, IARM_BUS_NETWORK_MANAGER_EVENT_INTERFACE_IPADDRESS, eventHandler) );
+                    IARM_CHECK( IARM_Bus_RegisterEventHandler(IARM_BUS_NM_SRV_MGR_NAME, IARM_BUS_NETWORK_MANAGER_EVENT_DEFAULT_INTERFACE, eventHandler) );
+                    LOGINFO("retryIarmEventRegistration: successfully subcribed to IARM event for Network Plugin");
```
karuna2git, line 360: Change the logging. Add a log telemetry header as

LOGINFO
("NETWORK_AVAILABILITY_RETRY_SUCCESS: xxxxxxxxx");

karuna2git (changes requested):

PR #2238, opened 2022-02-18
RDK-35348: Update the getPublicIP() for phase2

tabbas651:
[RDK-35348: Update the getPublicIP() for phase2]

Reason for change: updated the getPublicIP api to allow only interface and ipv6 parameter and also document changes 
Test Procedure: please referred from RDK-35348
Risks: Low
Signed-off-by: Thamim Razith <ThamimRazith_AbbasAli@comcast.com>

### DIFF
--- a/Network/Network.cpp
+++ b/Network/Network.cpp
@@ -52,6 +52,12 @@ using namespace std;
 #define IARM_BUS_NETSRVMGR_API_isAvailable "isAvailable"
 #define IARM_BUS_NETSRVMGR_API_getPublicIP "getPublicIP"
 
+string gEndPoint = "stun.l.google.com";
+uint16_t gStunPort = 19302;
+uint16_t gStunBindTimeout = 30;
+uint16_t gStunCacheTimeout = 0;
+bool gStunSync = true;
+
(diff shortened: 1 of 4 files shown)


karuna2git (approved):

Network/Network.cpp
```diff
@@ -52,6 +52,12 @@ using namespace std;
 #define IARM_BUS_NETSRVMGR_API_isAvailable "isAvailable"
 #define IARM_BUS_NETSRVMGR_API_getPublicIP "getPublicIP"
 
+string gEndPoint = "stun.l.google.com";
```
karuna2git, line 55: Can you move these global variables as class private variables.

Network/Network.cpp
```diff
@@ -52,6 +52,12 @@ using namespace std;
...
+uint16_t gStunPort = 19302;
+uint16_t gStunBindTimeout = 30;
+uint16_t gStunCacheTimeout = 0;
+bool gStunSync = true;
```
karuna2git, line 59: same for all 5 global variables.

karuna2git (changes requested):

PR #2239, opened 2022-02-18
RDK-35348: Update the getPublicIP() for phase2

tabbas651:
Reason for change: updated the getPublicIP api to allow only interface and ipv6 parameter and also update document changes
Test Procedure: please referred from RDK-35348
Risks: Low
Signed-off-by: Thamim Razith ThamimRazith_AbbasAli@comcast.com


karuna2git:
1. Do not initialize the gStunServer, etc in header, Init it in constructor.
2. change these gStunServer, gStunPort, etc as m_stunServer, m_stunPort, etc
3. Do not call returnResponse() from getPublic IP as it is going to print the public IP into logs.. 
Instead call,

{code}
 response["success"] = result; 
 return (Core::ERROR_NONE);
{code}

karuna2git (changes requested):
Mentioned in the previous comment

PR #2279, opened 2022-03-04
Delia 54683 2203

npoltorapavlo:


DataCapture/Utils.h
```diff
@@ -0,0 +1,104 @@
+#pragma once
+
```
anand-ky, line 2: Can this move to a common header file outside of plugins so that it can be used by all?

anand-ky (changes requested):

anand-ky (approved):

PR #2320, opened 2022-03-10
 RDKDEV-254:WebKitBrowser: Set additional environmental variables needed for MTLS support

nkkrishnan2512:
This commit adds additional config options and set environmental variables
 needed for MTLS integration.This mainly sets two environment variables
 such as G_TLS_OPENSSL_CLIENT_CERT_KEY_DECRYPT which indicates whether
 the private key decryption should be done or not,
 G_TLS_OPENSSL_CLIENT_CERT_KEY_TYPE which indicates the ssl file type.


emutavchi (changes requested):
The original idea for clientcert/key was that there would be a prepare-certs script that will do the needed to generate cert/key pair files in format suitable for the network library.  Could you please try that approach instead of introducing yet another set of env variables? see /lib/rdk/prepare-mauth-certs.sh and /lib/systemd/system/mauth-certs.service

rjpdasilva:
Hi @emutavchi,

We have use cases where the certificate and/or the private key are kept encrypted (e.g. using unique chip secrets).
In such cases, we have to avoid as much as possible having the data in the clear. Using such scripts, will leave the data in the clear for the whole lifetime of the boot (or at least the lifetime of the WebKitBrowser). We want to decrypt t only as late as possible and only during the exact time that is needed. For this, we do need to deal with encrypted cert/key within wpewebkit, hence the need for the new env vars.

Is there any alternative suggested approach for our use case?

Thanks

rjpdasilva:
@emutavchi,

Some other notes:
* The existing env vars `G_TLS_OPENSSL_CLIENT_CERT_PATH` and `G_TLS_OPENSSL_CLIENT_CERT_KEY_PATH` are not even used in the `wpewebkit` version we're using (tip of the `origin/wpe-20170728` branch). We have made use of them to implement client certificates for MTLS, supporting both DER and PEM formats. Then we've added these new env vars for allowing in-place decryption of the private key if so configured, but keeping always backwards compatibility to the "normal" cases.
* We also plan to upstream our changes in `wpewebkit` related to this, but we need first to do some refactoring to make the decryption part more abstract/generic/configurable (although we wont be allowed to upstream our decryption code, for obvious reasons).

emutavchi:
@rjpdasilva we have a change for glib-openssl to support mauth, if you have access here is the patch https://code.rdkcentral.com/r/plugins/gitiles/components/generic/rdk-oe/meta-rdk-ext/+/refs/heads/rdk-next/recipes-core/glib-openssl/files/0001-XRE-14265-request-client-cert-support.patch (if not, I can share with you privately). I believe you can improve the patch to load the client cert from the private storage to meet your requirements.

as for wpewebkit patches. The latest(upstream) version of wpewebkit already has some support for providing client cert: https://bugs.webkit.org/show_bug.cgi?id=200805 and there is an ongoing effort for backporting libsoup3 support to 2.28: https://github.com/WebPlatformForEmbedded/WPEWebKit/tree/wpe-2.28-soup3. so, it might be better to upstream your changes to libsoup

rjpdasilva:
@emutavchi,

> we have a change for glib-openssl to support mauth, if you have access here is the patch https://code.rdkcentral.com/r/plugins/gitiles/components/generic/rdk-oe/meta-rdk-ext/+/refs/heads/rdk-next/recipes-core/glib-openssl/files/0001-XRE-14265-request-client-cert-support.patch (if not, I can share with you privately). I believe you can improve the patch to load the client cert from the private storage to meet your requirements.

We have access to the patch, thank you.
At a first look, this seems like something we could use, but we'll have to be undergoing some changes and tests to confirm. This could be a good approach for avoiding the patching of the `wpewebkit`, and instead have everything handled within `glib-openssl`. Nevertheless, we would still need the changes done in this PR for having access to the new env vars from `glib-openssl`.

> as for wpewebkit patches. The latest(upstream) version of wpewebkit already has some support for providing client cert: https://bugs.webkit.org/show_bug.cgi?id=200805 and there is an ongoing effort for backporting libsoup3 support to 2.28: https://github.com/WebPlatformForEmbedded/WPEWebKit/tree/wpe-2.28-soup3. so, it might be better to upstream your changes to libsoup

This will not be possible for us at the moment, because we have hard dependencies yet to `wpewebkit` from the `wpe-20170728` branch and, consequently, also dependencies on older versions of `libsoup` which still do not support client certificates this easily. Anyway, it seems we could drop the patches on `wpewebkit` and instead use the `glib-openssl` as you've suggested, but as mentioned, we would need to test.

Either way, we will still need to keep the changes from this PR.

rjpdasilva:
Hi @emutavchi,

Regarding this:
> @rjpdasilva we have a change for glib-openssl to support mauth, if you have access here is the patch https://code.rdkcentral.com/r/plugins/gitiles/components/generic/rdk-oe/meta-rdk-ext/+/refs/heads/rdk-next/recipes-core/glib-openssl/files/0001-XRE-14265-request-client-cert-support.patch (if not, I can share with you privately). I believe you can improve the patch to load the client cert from the private storage to meet your requirements.

We can confirm that we can use that patch and improve in order to achieve our goals.
Nevertheless, we still need this PR to get through, because we still need these new environment variables for the reasons I've explained before.

emutavchi:
@rjpdasilva  would you mind trying to encode the key type into the 'clientcert' param? i.e. something like securestorage://<locator>.

rjpdasilva:
We're adding a var for the type PEM (default) vs DER and a flag var for indicating if it is to be decrypted or not false (default) vs true.  
Is you suggestion to have this information encoded in the existing vars, like going from:
```
PLUGIN_WEBKITBROWSER_CLIENT_CERT="/path/to/cert"
PLUGIN_WEBKITBROWSER_CLIENT_CERT_KEY="/path/to/key"
```
to something like:
```
PLUGIN_WEBKITBROWSER_CLIENT_CERT="der:///path/to/cert"
PLUGIN_WEBKITBROWSER_CLIENT_CERT_KEY="der-securestorage:///path/to/key"
```
or:
```
PLUGIN_WEBKITBROWSER_CLIENT_CERT="pem-securestorage:///path/to/cert"
PLUGIN_WEBKITBROWSER_CLIENT_CERT_KEY="der-securestorage:///path/to/key"
```
Please note we need to encode both the format (can assume a PEM default if not specified) and the need to decrypt (can assume no need for it if not specified), for both the cert and the key.

Firdose2109:

emutavchi:
@Firdose2109 my preference is to re-use existing configuration/variables and make necessary changes in the patch for the networking library.

rjpdasilva:
@emutavchi, @Firdose2109,

We have already a working implementation based on top of https://code.rdkcentral.com/r/plugins/gitiles/components/generic/rdk-oe/meta-rdk-ext/+/refs/heads/rdk-next/recipes-core/glib-openssl/files/0001-XRE-14265-request-client-cert-support.patch that uses these new env vars to achieve our goals.

However, as per @emutavchi's suggestion, we're now trying to reuse the existing variables with the schemes suggested to see if we can avoid the new variables, so this can be on hold for now.

Will get back here when we have news.

anand-ky:
Closing older PRs. Pls resubmit on latest branches if required.

PR #2388, opened 2022-03-24
RDK-35521: Add Unit Tests for DataCapture using Google Test Framework

mateusz-adamski-red:
Signed-off-by: mateusz.adamski <mateusz.adamski2@sky.uk>

### DIFF
--- a/DataCapture/DataCapture.cpp
+++ b/DataCapture/DataCapture.cpp
@@ -85,7 +85,7 @@ namespace WPEFramework {
             Register(METHOD_ENABLE_AUDIO_CAPTURE, &DataCapture::enableAudioCaptureWrapper, this);
             Register(METHOD_GET_AUDIO_CLIP, &DataCapture::getAudioClipWrapper, this);
 
-            _sock_adaptor = new socket_adaptor();
+            _sock_adaptor.reset(new socket_adaptor());
         }
 
         DataCapture::~DataCapture()
@@ -104,7 +104,6 @@ namespace WPEFramework {
         void DataCapture::Deinitialize(PluginHost::IShell* /* service */)
         {
             DeinitializeIARM();
-            delete _sock_adaptor;
             DataCapture::_instance = nullptr;
         }
 
--- a/DataCapture/DataCapture.cpp
+++ b/DataCapture/DataCapture.cpp
@@ -20,7 +20,6 @@
 
 #include <algorithm>
 #include <regex>
-#include "audiocapturemgr_iarm.h"
 #undef LOG // we don't need LOG from audiocapturemgr_iarm as we are defining our own LOG
 #include "DataCapture.h"
 #include <curl/curl.h>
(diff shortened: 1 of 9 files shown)


npoltorapavlo:
@mateusz-adamski-red, please avoid changing the plugin's code unless it is a fix. There's no need to change socket_adaptor or DataCapture. By mocking socket_adaptor you skip testing socket_adaptor which is an important part of the plugin. So please remove SocketAdaptorMock and add a dummy websocket server. Pass its address as dataLocator. All IARM_Bus_Call-s should have expectations. There's no need to duplicate tests, for example, getAudioClip can be tested 1 time instead of 5 times.

npoltorapavlo:
The test is passing although it's not working correctly. For example, `POST / HTTP/1.1 200` isn't a valid HTTP response. So the log says `Upload failed: 1:'Unsupported protocol' (cURL error)`. Please make it work correctly. Also, add expectations for onAudioClipReady, similar to: https://github.com/rdkcentral/rdkservices/pull/2390/files#diff-c89082c11121049d83b0daafa144ca170429588534bde75070e75e41e33a2093 . ShouldTurnOnAudioCapture and ShouldUploadData invoke the same APIs. Do we need both of these tests?

mateusz-adamski-red:
> 

We have 2 tests, because there is 2 different expectations. First test check only if `enableAudioCapture` work. Second test check how events are handling. I think we should have 2 separate test. Follow rule 1 test - 1 behavior.

npoltorapavlo:
Please remove enableAudioCapture and getAudioClip from ShouldUploadData if they are not required for eventHandler to work. Also please write expectations for all IARM_Bus_Call-s. Your test is using uninitialized values. For example, max_buffer_duration is 78250038 so the check `_max_supported_duration < incoming_duration` works by chance. constructFormatString fails because the values are uninitialized. The test should not rely on uninitialized values. Please fix.

mateusz-adamski-red:
> Please remove enableAudioCapture and getAudioClip from ShouldUploadData if they are not required for eventHandler to work. Also please write expectations for all IARM_Bus_Call-s. Your test is using uninitialized values. For example, max_buffer_duration is 78250038 so the check `_max_supported_duration < incoming_duration` works by chance. constructFormatString fails because the values are uninitialized. The test should not rely on uninitialized values. Please fix.

Expecations added. We need  getAudioClip from ShouldUploadData in enableAudioCapture test.

npoltorapavlo:
with your test valgrind reports:
- 8 uninitialised
- 8 Invalid read
- FILE DESCRIPTORS: 10 open at exit
- definitely lost: 480 bytes in 5 blocks

without your test, it's:
- 2 uninitialised
- no Invalid read
- FILE DESCRIPTORS: 4 open at exit
- definitely lost: 24 bytes in 2 blocks

Please fix

npoltorapavlo:
looks good

anand-ky (approved):

PR #2431, opened 2022-04-05
RDKTV-15642: Avoid deadlock in Monitor

npoltorapavlo:
Reason for change: ThreadPool::Submit blocks if no slots.
If server locks were acquired (like for StateChange)
and all other threads depend on them then it's a deadlock.

Test Procedure: Test Monitor plugin.
Risks: Low


Monitor/Monitor.h
```diff
@@ -774,7 +774,7 @@ namespace Plugin {
                             // Moreover it's the only only which now becomes active. This means probing
                             // has to be activated as well since it was stopped at point the last observee
                             // turned inactive
-                            _job.Submit();
+                            _job.Schedule(Core::Time::Now());
```
MFransen69, line 777: I understand the deadlock situation. But this solution will no longer work in R3 as there if the schedule_time <= Now() it will internally submit instead of schedule.
I think an alternative and more consistent way to fix it is to Submit the job outside the lock as we do elsewhere in this file. I think that is the bug, that the submit is inside the lock, that should not have happened

MFransen69 (changes requested):

Monitor/Monitor.h
npoltorapavlo: @MFransen69, the lock with is causing the deadlock is `_notificationLock.Lock();` in ServiceMap::StateChange.

MFransen69: Hi @npoltorapavlo 
Thanks for explaining. The way the state changes are handled is redesigned in R3. Let me discuss with the person who did this and get back on this (he is not in at the moment)

MFransen69 (approved):
Hi @npoltorapavlo 

I discussed with the person who did the SateChange changes in R3 and that was indeed to also fix the threading issues there. Bringing these to R2 is not an option as it requires an interface change. So I propose to approve your changes in this PR and bring them to the RDKServices to fix the issue locally, they will be changed again whenever the new R3 version of the Monitor (together with the R3 Thunder) will make it to RDKServices.

npoltorapavlo:
@pwielders @anand-ky please merge

anand-ky:
@MFransen69 @pwielders - Checking if we are waiting for Pierre's review or can we merge with Marcel's +1.

MFransen69:
Hi @anand-ky yes you can merge just based on my approval

anand-ky:
thanks @MFransen69

PR #2446, opened 2022-04-08
DELIA-55604: Fix for terminate called without an active exception in HDMICEC

ksanth850:
DELIA-55604: Fix for terminate called without an active exception in HDMICEC

Reason for change: Thread join for created updated and poll thread
Test Procedure: Activate and deactivate HDMICEC plugin number of times
Risks: Medium


karuna2git (commented):
Im good with the change

apatel859 (approved):

apatel859:
> I don't see Initialize() for this plugin. Can we add one. Also CecDisable() is called while deactivating multiple plugins like HDMICec_2, HDMICecSink. We have had issues before where we deactivate one of the plugins and it affects CEC. Can we add some logic to ensure that CECDisable is called only once when all plugins are deinitialized. Also vice versa for CECEnable.

Anand, We dont expect multiple CEC plugin to be running on the same device. So coordination not required. 

Kumar,  Initialize is not there as it was done in constructor it self. So you can move init part from contractor to Initialize .

karuna2git:
Hey @apatel859  & @anand-ky, 
The fix that was contributed is to address the crash. Enhancing the plugin to have ::Initialize() is not intended here.

Also, I guess its upto the plugin owner to decide whether to have separate  ::Initialize()  function for their plugin.
If you wanted to maintain a consistency with other plugins, then below plugins also needs the change. May be we can address all 5 plugins in a separate PR; Not in this.

1. FrontPanel
2. HdcpProfile
3. HdmiCecSink
4. HdmiInput

Pl advise..

Thanks,
Karunakaran A

anand-ky:
sounds good. @apatel859 - Pls address creating Initialize() in a separate change.

anand-ky (approved):

PR #2452, opened 2022-04-11
RDK-36049: Get DHCP Server address of my network interface

ksanth850:
Reason for change: To get DHCP ServerIP Address even ipversion case sensitive test cases also.
Test Procedure: As RDK Developer, I must be able to get DHCP Server address of my network interface.
                This DHCP Server address will be used to pass it to DVB-CI+ implementation.
Risks: High


ksanth850:
Based on review comments going to create new one

PR #2468, opened 2022-04-15
RDK-36282: Add support for passing user scripts

Scony:
allows injecting user scripts via config:
```
"userscripts": ["/tmp/data/script.js", "script2.js", "nonexistent"]
```

### DIFF
--- a/WebKitBrowser/WebKitImplementation.cpp
+++ b/WebKitBrowser/WebKitImplementation.cpp
@@ -492,6 +492,7 @@ static GSourceFuncs _handlerIntervention =
                 , WatchDogCheckTimeoutInSeconds(0)
                 , WatchDogHangThresholdInSeconds(0)
                 , LoadBlankPageOnSuspendEnabled(false)
+                , UserScripts()
             {
                 Add(_T("useragent"), &UserAgent);
                 Add(_T("url"), &URL);
@@ -545,6 +546,7 @@ static GSourceFuncs _handlerIntervention =
                 Add(_T("watchdogchecktimeoutinseconds"), &WatchDogCheckTimeoutInSeconds);
                 Add(_T("watchdoghangthresholdtinseconds"), &WatchDogHangThresholdInSeconds);
                 Add(_T("loadblankpageonsuspendenabled"), &LoadBlankPageOnSuspendEnabled);
+                Add(_T("userscripts"), &UserScripts);
             }
             ~Config()
             {
@@ -603,6 +605,7 @@ static GSourceFuncs _handlerIntervention =
             Core::JSON::DecUInt16 WatchDogCheckTimeoutInSeconds;   // How often to check main event loop for responsiveness
             Core::JSON::DecUInt16 WatchDogHangThresholdInSeconds;  // The amount of time to give a process to recover before declaring a hang state
             Core::JSON::Boolean LoadBlankPageOnSuspendEnabled;
+            Core::JSON::ArrayType<Core::JSON::String> UserScripts;
         };
 
         class HangDetector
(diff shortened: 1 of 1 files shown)


emutavchi (approved):
looks good to me. let's wait for a review from Metrological team.

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2248,6 +2251,27 @@ static GSourceFuncs _handlerIntervention =
...
+                            return;
+                        }
+                        auto* script = webkit_user_script_new(scriptContent, WEBKIT_USER_CONTENT_INJECT_ALL_FRAMES, WEBKIT_USER_SCRIPT_INJECT_AT_DOCUMENT_START, nullptr, nullptr);
+                        webkit_user_content_manager_add_script(userContentManager, script);
+                        webkit_user_script_unref(script);
+                        g_free(scriptContent);
+                    };
+                for (unsigned userScriptIndex = 0; userScriptIndex < _config.UserScripts.Length(); userScriptIndex++) {
+                    const auto& scriptPath = _config.UserScripts[userScriptIndex];
+                    SYSLOG(Trace::Information, (_T("Loading user script '%s'"), scriptPath.Value().c_str()));
```
MFransen69, line 2270: Hi, for the error message above it is consistent with other error message in this file to use SYSLOG however for this line  where it is informational I would propose to use "TRACE(Trace::Information, ..." here so we do not pollute the syslog output with informational messages

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2248,6 +2251,27 @@ static GSourceFuncs _handlerIntervention =
                 reinterpret_cast<GCallback>(wpeNotifyWPEFrameworkMessageReceivedCallback), this);
             webkit_user_content_manager_register_script_message_handler(userContentManager, "wpeNotifyWPEFramework");
 
+            if (_config.UserScripts.IsSet()) {
+                auto loadScript =
+                    [&userContentManager](const std::string& path) {
+                        gchar* scriptContent;
+                        auto success = g_file_get_contents(path.c_str(), &scriptContent, nullptr, nullptr);
```
MFransen69, line 2258: I guess you do not always expect these scripts to be in the datapath? As that should be the case for a stack where datapath is filled could we (like for the extension path) make these relative to the datapath when datapath is set (or additionally check if the scriptpath is a relative path, does not start with a /, and only then append it the the datapath)?

MFransen69, line 2258: I guess you do not always expect these scripts to be in the datapath? As that should be the case for a stack where datapath is filled could we (like for the extension path) make these relative to the datapath when datapath is set (or additionally check if the scriptpath is a relative path, does not start with a /, and only then append it the the datapath)?

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2414,9 +2438,39 @@ static GSourceFuncs _handlerIntervention =
 
             WKPageGroupSetPreferences(pageGroup, preferences);
 
+            auto userContentController = WKUserContentControllerCreate();
             auto pageConfiguration = WKPageConfigurationCreate();
             WKPageConfigurationSetContext(pageConfiguration, wkContext);
             WKPageConfigurationSetPageGroup(pageConfiguration, pageGroup);
+            WKPageConfigurationSetUserContentController(pageConfiguration, userContentController);
+
+            if (_config.UserScripts.IsSet()) {
```
MFransen69, line 2447: As the code is similar to the above put it in a private function and call it in both places perhaps?

MFransen69, line 2447: As the code is similar to the above put it in a private function and call it in both places perhaps?

MFransen69 (changes requested):

Firdose2109:

Scony:
@MFransen69 could you please re-review?

PR #2531, opened 2022-04-28
DELIA-55550: Security fix in UsbAccess

npoltorapavlo:
Reason for change: The updateFirmware service does not sufficiently
check api input before passing it to a system() call, allowing
unescaped commands in the path portion of the "fileName" argument.
Test Procedure: Test org.rdk.UsbAccess.2.updateFirmware.
Risks: Low


Firdose2109:

anand-ky (approved):

PR #2532, opened 2022-04-28
RDKTV-16004 : setDisplayFrameRate method is not working properly

nsahu910:
Reason for change: adding checks for framerate value
Test Procedure:check in ticket
Risks: none
Signed-off-by: Neeraj Sahu <Neeraj_Sahu@comcast.com>

### DIFF
--- a/FrameRate/FrameRate.cpp
+++ b/FrameRate/FrameRate.cpp
@@ -265,16 +265,27 @@ namespace WPEFramework
 
             string sFramerate = parameters["framerate"].String();
 
-            bool success = true;
-            try
+            bool success = false;
+            if(sFramerate == "3840x2160x60"
+                            || sFramerate == "3840x2160x47"
+                            || sFramerate == "3840x2160x48"
+                            || sFramerate == "3840x2160x50"
+                            || sFramerate == "3840x2160x59")
+	    {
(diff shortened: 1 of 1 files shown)


Firdose2109:

FrameRate/FrameRate.cpp
```diff
@@ -265,16 +265,27 @@ namespace WPEFramework
...
 
-            bool success = true;
-            try
-            {
-                device::VideoDevice &device = device::Host::getInstance().getVideoDevices().at(0);
-                device.setDisplayframerate(sFramerate.c_str());
-            }
-            catch (const device::Exception& err)
+            bool success = false;
+            if(sFramerate == "3840x2160x60"
```
anand-ky, line 269: We cannot hardcode these values since this is a generic plugin that can be used on different devices with different supported display frame rates. The HAL should error out if the DisplayFrameRate is not getting set. Pls reach out to @apatel859 on details on the error returned and return suitable error in the jsonrpc response for setDisplayFrameRate

anand-ky (changes requested):

anand-ky:
Closing older PRs. Pls resubmit to latest 2205 sprint.

PR #2535, opened 2022-04-28
RDKTV-16030  Create class hierarchy framework and code refinement

tamilselvanak:


Firdose2109:

ControlSettings/ControlSettings.cpp
```diff
@@ -248,16 +248,29 @@ namespace Plugin {
                , m_isDisabledHdmiIn4KZoom (false)
     {
         LOGINFO("Entry\n");
-        devicePtr = new WPEFramework::Plugin::TV;
+#ifndef ENABLE_TV_SUPPORT
```
apatel859, line 251: in recipe file define something like  #define CONFIG_DEVICE_TYPE TV or STB
and use that variable here to create object of TV or STB class

ControlSettings/STB.cpp
```diff
@@ -25,11 +25,12 @@ namespace Plugin {
 
     STB::STB()
```
apatel859, line 26: Class name should be controlSettingBase. controlSettingTV controlSettingSTB

ControlSettings/ControlSettings.cpp
```diff
@@ -248,16 +248,29 @@ namespace Plugin {
...
+        devicePtr = new STB;
+#else
+        devicePtr = new TV;
+#endif
+
 	ControlSettings::_instance = this;
 	InitializeIARM();
 
+        if(devicePtr->isTvSupportEnabled())//TV Specific API
```
apatel859, line 260: This we diiscuss to move to TV and STB specifc class. so you should add virtial Register method

apatel859, line 260: I observe lots of  code in control setting which has reference to TV ...should be move to TV specific implemetnation

ControlSettings/ControlSettings.cpp
tamilselvanak: Thank you, Amit. "CONFIG DEVICE TYPE TV or CONFIG DEVICE TYPE STB" will be the new name.
Will change it accordingly.

ControlSettings/ControlSettings.cpp
tamilselvanak: We cant move out API registration to derive class because,

1)AS creating object for ControlSettings class which is independent of Device/STB/TV class. As a result, AS has access to an API that is only available in the ControlSettingsPlugin class scope.

2)Let's say you want to forward the API registration to the TV class.

registerMethod("getBacklight", &TVMgr::getBacklight, this, {2});----->API will be register for TV class which not available in AS scope.

If you want to register any API with the TV/STB/Device class, the second argument must be a function pointer, and the third argument must be "this" ptr.

In this scenario, AS will not call getBacklight since it is registered in the TV class, and AS can only view the controlSettingsplugin functions.

As a result, just the registration and init sequences were pushed into the Main ControlSettings class, and the functionality was moved to a TV/STB class.

tamilselvanak: As per my previous comment All registartion and Init sequence has been kept in "ControlSettings" Main class.
So only AS can invoke the corresponding API.

PR #2545, opened 2022-05-03
RDK-35488: Issues in CMF-RPi Dunfell Builds when rdkshell in surfacemode

madanagopalt:


Firdose2109:

PR #2585, opened 2022-05-10
RDKTV-13125: WPEFramework timeout during shutdown in Maintenance cleanup

PaviP98:
RDKTV-13125: WPEFramework timeout during shutdown in Maintenance cleanup

Graceful exit is not happening with maintenace manager when we stop wpeframework.
It got stuck with deInitialize method as it locks until a conditional variable is notified/
Hence unlocked the tasks if it is still waiting.

Verify maintenance manager is exiting gracefully when maintenance is in progress


tdeva14:
Changes looks good

mvalas827 (approved):
Looks good. Please merge.

mvalas827 (commented):
Looks good.

PR #2598, opened 2022-05-11
RDK-32095: Launch Netflix in Dobby container

goruklu:
cherry-pick Netflix container mode changes in RDKShell to element_svt branch


goruklu:
after discussing this internally, it seems like we don't need this in element_svt branch. It's already part of element_pp branch.

PR #2604, opened 2022-05-12
RDK-36297: Add api to get frame rate

zhilarasi:
Reason for change : changes for API to get frame rate

Test Procedure: TBD

Risks: None


binuinbaraj:
please update the commit with only changed lines

PR #2626, opened 2022-05-17
[RDKDEV-262] DeviceIdentification: syncup from metro version

pkumbh631:


pkumbh631:
@mhughesacn I have updated the NOTICE file as per your comment even though Copyright scan is failing.

mhughesacn:
Hi @pkumbh631: The team have pointed out that this is not the right way to use github/pull reqs.  They ask that you make the NOTICE change on the first pull request as a second commit (i.e. reopen pull req 2621) and close this one.  I will have to clear the copyright fail, which I'll do as soon as I see it. Thanks!

PR #2632, opened 2022-05-18
RDK-35564: Support Semantic Versioning for RDKServices

mateusz-adamski-red:


mateusz-adamski-red:
`update chanelog and api version` failed because I didn't modify CHANELOG.md here i add it. This web-hook check if file was modified not added.

PR #2636, opened 2022-05-20
RDKDEV-259 :Security agent : Bug fix in security agent plugin's configuration and url for token creation in injected bundle

johansaji:

This PR is the rebased version of **https://github.com/rdkcentral/rdkservices/pull/2357**

### DIFF
--- a/WebKitBrowser/InjectedBundle/SecurityAgent.cpp
+++ b/WebKitBrowser/InjectedBundle/SecurityAgent.cpp
@@ -46,6 +46,8 @@ namespace JavaScript {
                     uint8_t buffer[2 * 1024];
 
                     std::string url = WebKit::Utils::GetURL();
+                    //Formatting the url string to json for creating token
+                    url = "{\"url\":\"" + url + "\"}";
 
                     std::string tokenAsString;
                     if (url.length() < sizeof(buffer)) {
--- a/SecurityAgent/SecurityAgent.config
+++ b/SecurityAgent/SecurityAgent.config
@@ -1,6 +1,7 @@
 set (autostart true)
+set(PLUGIN_SECURITYAGENT_ACL_FILE_NAME "acl.json" CACHE STRING "SecurityAgent ACL file name")
 
 map()
-    map(acl acl.json)
+    kv(acl ${PLUGIN_SECURITYAGENT_ACL_FILE_NAME})
 end()
 ans(configuration)


WebKitBrowser/InjectedBundle/SecurityAgent.cpp
```diff
@@ -46,6 +46,8 @@ namespace JavaScript {
                     uint8_t buffer[2 * 1024];
 
                     std::string url = WebKit::Utils::GetURL();
+                    //Formatting the url string to json for creating token
```
anand-ky, line 49: Pls remove this change. The url is currently used as is and this will break other c hanges.

anand-ky (changes requested):

pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3132

WebKitBrowser/InjectedBundle/SecurityAgent.cpp
johansaji: @anand-ky wouldnt keeping the url as it create conflict in SecurityAgent Plugin here https://github.com/rdkcentral/rdkservices/blob/sprint/2207/SecurityAgent/SecurityContext.cpp#L61

rjpdasilva: This change is needed to be compliant with the `SecurityAgent` plugin from the same sprints.

Please check the code in https://github.com/rdkcentral/rdkservices/blob/sprint/2208/SecurityAgent/SecurityContext.cpp#L55-L64:
{code:cpp}
    SecurityContext::SecurityContext(const AccessControlList* acl, const uint16_t length, const uint8_t payload[])
        : _token(string(reinterpret_cast<const TCHAR*>(payload), length))
        , _accessControlList(nullptr)
    {
        _context.FromString(_token);

        if ( (_context.URL.IsSet() == true) && (acl != nullptr) ) {
            _accessControlList = acl->FilterMapFromURL(_context.URL.Value());
        }
    }
{code}

This is where the token is consumed by the `SecurityAgent` plugin.
Please note that the `_context.FromString(_token);` statement means that the code is assuming that `_token` is a JSON object that could have the `url`, `user` and `hash` keys, as defined by the `Payload` class:
* https://github.com/rdkcentral/rdkservices/blob/sprint/2208/SecurityAgent/SecurityContext.h#L82
{code:cpp}
        Payload _context;
{code}
* https://github.com/rdkcentral/rdkservices/blob/sprint/2208/SecurityAgent/SecurityContext.h#L30-L53
{code:cpp}
        class Payload : public Core::JSON::Container {
        public:
            Payload(const Payload&) = delete;
            Payload& operator=(const Payload&) = delete;

            Payload()
                : Core::JSON::Container()
                , URL()
                , User()
                , Hash()
            {
                Add(_T("url"), &URL);
                Add(_T("user"), &User);
                Add(_T("hash"), &Hash);
            }
            ~Payload()
            {
            }

        public:
            Core::JSON::String URL;
            Core::JSON::String User;
            Core::JSON::String Hash;
        };
{code}

So yes, even on `sprint/2208`, this should already be needed!
I'll retarget the PR to `sprint/2208`.

PR #2649, opened 2022-05-24
TTSManager::setConfiguration() - check voices.empty() before using first()

Gavriliuk:


Gavriliuk (commented):
This change should avoid of run-time crashes

The misconfiguration issue could be fixed separately

ramasat (approved):

PR #2652, opened 2022-05-24
Publish latest documentation

jruzzi:
This PR includes a review of the 2204 changes and fixes files breaking the build. It includes the latest generation script and manually fixes broken cross links to properties.

### DIFF
--- a/Messenger/common.json
+++ b/Messenger/common.json
@@ -1,0 +1,190 @@
+{
+    "errors": {
+      "general": {
+        "code": 1,
+        "message": "ERROR_GENERAL"
+      },
+      "unavailable": {
+        "code": 2,
+        "message": "ERROR_UNAVAILABLE"
+      },
+      "asyncfailed": {
+        "code": 3,
+        "message": "ERROR_ASYNC_FAILED"
+      },
+      "asyncaborted": {
+        "code": 4,
+        "message": "ERROR_ASYNC_ABORTED"
+      },
+      "illegalstate": {
+        "code": 5,
+        "message": "ERROR_ILLEGAL_STATE"
+      },
+      "openingfailed": {
+        "code": 6,
+        "message": "ERROR_OPENING_FAILED"
+      },
+      "acceptfailed": {
+        "code": 7,
+        "message": "ERROR_ACCEPT_FAILED"
+      },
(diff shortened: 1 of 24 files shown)


Messenger/common.json
```diff
@@ -0,0 +1,190 @@
+{
```
anand-ky, line 1: Can you remove this file. This was updated recently to reflect the latest interface for Messenger.

anand-ky (changes requested):

Messenger/common.json
anand-ky: @npoltorapavlo - Can you check if it's ok to add Common.json file here in Messenger.

docs/api/MessengerPlugin.md
```diff
@@ -1,5 +1,5 @@
 <!-- Generated automatically, DO NOT EDIT! -->
-<a name="head.Messenger_Plugin"></a>
+<a name="Messenger_Plugin"></a>
```
npoltorapavlo, line 2: "head." and other things are auto-generated. Why are we editing an auto-generated file?

Messenger/common.json
npoltorapavlo: I can't say any of this is "ok" as my understanding is interfaces should be in ThunderInterfaces and not in rdkservices.

jruzzi: @npoltorapavlo, that is not how rdkservices works at present and I'm not trying to get into a religious war here. The fact is that you added a reference to a file that did not exist and that broken link broke the documentation build. We either have to restore the file the way it was, or we have to add the referenced file (common.json) somewhere in our repository.

docs/api/MessengerPlugin.md
jruzzi: The generated files are post-processed using a script that fixes the non-standard generated anchors. These anchors lead to broken links in our documentation site.

anand-ky:
Let's go ahead with these changes. Should be fine to have common.json here in RDKServices.

jruzzi:
I will open a separate JIRA to move common.json to a central location so that other plugins can reference the error codes and we do not have to include the error code definitions in each of the JSON definitions.

PR #2735, opened 2022-06-09
RDK-35564: Support Semantic Versioning for RDKServices

nauka-programowania-MA:


nauka-programowania-MA:
This will be committed with: https://github.com/rdkcentral/Thunder/pull/979 So unittests will fail because we have a lack of implementations and the build failed.

PR #2779, opened 2022-06-20
Clean up generic utils

npoltorapavlo:


npoltorapavlo:
a link to successful gerrit verification https://gerrit.teamccp.com/#/c/653136/

anand-ky (approved):

anand-ky:
@npoltorapavlo - Pls add a note in the README to include these specific utils header files.

npoltorapavlo:
@anand-ky the point is to reduce dependencies so that when writing unit tests there's no need to cope with includes that aren't needed. From this perspective, generic utils should be removed. When it's done there's no need to instruct not to use them so I did not add anything to the README

PR #2789, opened 2022-06-21
RDK-34568: Expand AVInput RDK Service to include HDMIInput and Compos…

dhivya-ilangovan:
…iteInput APIs

Reason for change: Combine HdmiInput and CompositeInput Plugins into AVInput
Test Procedure: sanity testases on HDMI and COMPOSITE inputs
Risks: Low

### DIFF
--- a/AVInput/AVInput.cpp
+++ b/AVInput/AVInput.cpp
@@ -18,193 +18,1160 @@
 **/
 
 #include "AVInput.h"
-
 #include "dsMgr.h"
 #include "hdmiIn.hpp"
-
+#include "compositeIn.hpp"
 #include "UtilsJsonRpc.h"
 #include "UtilsIarm.h"
+#include "exception.hpp"
+#include "dsUtl.h"
+#include "dsError.h"
+#include <vector>
+#include <algorithm>
+
+#define HDMI 0
+#define COMPOSITE 1
+#define AV_HOT_PLUG_EVENT_CONNECTED 0
+#define AV_HOT_PLUG_EVENT_DISCONNECTED 1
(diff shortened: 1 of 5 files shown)


mhughesacn:
Hi Dhivya,
I will have to clear the Arris copyrights in the scans.  Three things:
- Please confirm that RDK has changed the two files from Arris.  If you have not, then there is no need for an RDK Apache header on top.
- Please update NOTICE at top level to give credit to Arris.  Append:
```
Copyright ARRIS Enterprises, Inc. 2015
```
- The file headers should say:
`If not stated otherwise in this file or this component's LICENSE file...`
i.e. "Licenses.txt" is not correct, although you will find it because we haven't fixed this everywhere.

Thank you,

mhughesacn:
Hi Dhiyva,
What has happened to the Arris copyrights?  Is all the Arris material removed?  Also, all code files should have proper RDK headers (or Arris etc.) as appropriate.

npoltorapavlo:
@dhivya-ilangovan - the purpose of unit tests is to block changes that break them. You shouldn't adjust the tests for the errors in the code to pass. `-Werror` is there because it isn't hard to fix warnings and keep the code clean. So instead of removing `-Werror` can you please fix the warnings. As you don't need to write tests for the new APIs, it's only required to not break the existing tests. This is almost trivial. Add into test/headers/ the new headers your code includes. It should be good to start with the blank files. Then add into the headers only the items which your code references. For this, you can check the real header but don't copy-paste everything. As an example you can check test/headers/hdmiIn.hpp - it isn't the same as hdmiIn.hpp in devicesettings but it has everything for the plugin to build. Please avoid changing the whitespaces when you push the changes on review.

PR #2797, opened 2022-06-22
WebKitBrowser - enable media stream support

emutavchi:


modeveci:
@emutavchi I am currently checking if it is needed to make MediaStream as configuration option or good to keep always enabled.

modeveci (approved):
LGTM

PR #2845, opened 2022-07-05
Add configurations options for HTML5 Indexed Databases

asurdej-comcast:
1) Configure IndexedDB from plugin configuration file:
 * IndexedDB JS api availability
 * storage path
 * disk usage
2) Disable IndexedDB by default
3) Change default storage dir to g_get_user_cache_dir()/wpe/databases/indexeddb
(WebKit default was g_get_user_data_dir()/databases/indexeddb)


emutavchi (approved):

MFransen69:
As this is more on the webkit behaviour, @modeveci could you please have a alook?

modeveci (commented):
Do we need to change in any case the path and size of indexeddb? Or is it good to link when the enable flag is set?

modeveci:
Got it! Thank you!

modeveci (approved):

PR #2881, opened 2022-07-13
RDKTV-18114,RDKTV-18107: WPEFramework Crash in std::terminate

PaviP98:
Moved stopMaintenance RFC check to stopMaintenance() from stopMaintenanceTasks()
Removed unnecessary thread joins in MM

There should not be any crash with Maintenance Manager.
Ensure there were no issues with MM during wpeframework shutdown


mvalas827 (approved):
Looks good.

anand-ky (approved):

PR #2907, opened 2022-07-19
RDKTV-17965: Fix LocationService hang

npoltorapavlo:
Reason for change: Do not rely on system time when scheduling.
Test Procedure: Unknown.
Risks: Low

### DIFF
--- a/LocationSync/LocationService.cpp
+++ b/LocationSync/LocationService.cpp
@@ -327,7 +327,7 @@ namespace Plugin {
         , _country()
         , _region()
         , _city()
-        , _activity(*this)
+        , _activity()
         , _infoCarrier()
         , _request(Core::ProxyType<Web::Request>::Create())
     {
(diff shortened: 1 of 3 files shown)


MFransen69:
Hi @npoltorapavlo 

Would it be possible to have a short call to discuss this? I would like to understand the issue behind this better. I read the RDKTV-17965 jira issue text and I want to make sure I understand what the root cause is.

Best, Marcel Fransen

npoltorapavlo:
@MFransen69 the issue is Internet precondition is sometimes never met. I can't tell you what is the root cause because I have not seen it. In logs where it was seen the system time changes -+15 days during probe retries. Schedule relies on Core::TimerType which calculates the time delta. With the system time change, the delta can be wrong so I'm looking for a solution without the Schedule.

MFransen69:
Hi @npoltorapavlo 
Indeed that was my guess after reading the issue text Then I have a feeling of what might be the bigger root cause of this issue. Then fixing it like this is working around the problem, and it might manifest as well in other places. I think it will be easier to go over this in a call so I can explain in detail. Are you available on slack?

PR #2918, opened 2022-07-21
 RDKTV-17958: After waking TV from standby Audio not routed to eArc.

shashankkumarc:
Reason for Change: eArc not enabled due to delayed HPD event
Test Procedure: As per Jira
Risks: Low

Signed-off-by: shashankkumarc <shashank.kumar@sky.uk>


shashankkumarc:
Creating new request from 2208 sprint

PR #2922, opened 2022-07-21
RDK-32992 - Add Ermgr blacklist APIs

rwarier:


anand-ky (approved):

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -1702,8 +1702,10 @@ static GSourceFuncs _handlerIntervention =
             if (_config.ClientIdentifier.IsSet() == true) {
                 string value(service->Callsign() + ',' + _config.ClientIdentifier.Value());
                 Core::SystemInfo::SetEnvironment(_T("CLIENT_IDENTIFIER"), value, !environmentOverride);
+                Core::SystemInfo::SetEnvironment(_T("ESSRMGR_APPID"), value, !environmentOverride);
```
emutavchi, line 1705: There was an agreement to use CLIENT_IDENTIFIER to identify app. Any reason to repeat same info with different env variable?

rwarier: Essos will read the variable CLIENT_IDENTIFIER if ESSRMGR_APPID is not set. So this setenv is not required.

PR #2927, opened 2022-07-22
RDK-34568: For testing

dhivya-ilangovan:
Expand AVInput RDK Service to include HDMIInput and CompositeInput


npoltorapavlo:
@dhivya-ilangovan since you are adding new APIs please follow the [JSON-RPC spec](https://www.jsonrpc.org/specification#response_object). This is not hard to do. Instead of returnResponse, return the [error code](https://github.com/rdkcentral/Thunder/blob/239dc48818083b684537553353715a11e793710a/Source/core/Portability.h#L670). For the documentation purposes reflect in the schema which failures the API can return, for example like [here](https://github.com/rdkcentral/rdkservices/blob/df8b598e1943e0f35d7af8cf77943e034b278a1d/Messenger/Messenger.json#L76-L77). Also a good practice is to use JSON stubs. You have implemented the schema AVInputPlugin.json. The repo for these schemas is https://github.com/rdkcentral/ThunderInterfaces/tree/R2/jsonrpc. If your schema is valid, ThunderInterfaces will auto-generate a header file JsonData_AVInput.h. It is preferable to use C++ structures from that generated file., instead of hand-writing a JsonObject response.

dhivya-ilangovan:
Closing the current PR as having conflicts with latest base
https://github.com/rdkcentral/rdkservices/pull/2961

PR #2938, opened 2022-07-26
REFPLTV-1319: Fixing RDK Services getDefaultInterface

anilrdk:
Reason for change: Resetting defaultInterface value at change in Interface event

Test Procedure: Set the default interface to ETHERNET and check if getDefaultInterface gets correct value.

Risks: Low

Signed-off-by: avadla226 <Anil_Vadla@comcast.com>


Network/Network.cpp
```diff
@@ -1420,6 +1420,8 @@ namespace WPEFramework
             m_useIpv4EthCache = false;
             m_useIpv6EthCache = false;
             m_defIpversionCache = "";
+            m_defaultInterface = ""; /* REFPLTV-1319 : Resetting when there is switch in interface, to get new value in getDefaultInterface() */
```
karuna2git, line 1423: why do we need this?
The `m_defaultInterface` and `m_gatewayInterface` will be updated when you call getDefaultInterface() based on the cache value or fresh GET based on `m_useDefInterfaceCache` which is being reset right above Line#1417.

Are there any wrong data returned at any point?

anilrdk: The m_defaultInterface and m_gatewayInterface will be updated only for the first time. When there is change in Interface, we get onDefaultInterfaceChanged() notification. Now when we call getDefaultInterface(), it just gives old interface values and not the updated one. This is because of the case "if (m_defaultInterface.length() == 0)". This length is always > 0, because it had old values. And hence it does not enter that loop to get new interface values. With this fix, m_defaultInterface.length will become 0 and then we get new interface values.

karuna2git: `if (m_defaultInterface.length() == 0)` being executed only under HYBRID section.
In RDK terminology, the HYBRID Device Type means its GATEWAY device. ie It has DOCSIS modem as default interface and such device the default interface cannot be changed.

You must be using mediaclient as devicetype if not using the DOCSIS gateway.

This change is good; I can approve it. But this change unearths some questions on how the RDK Services and/or RDK Middleware is configured in your product. Let me know how I can help in that.

Thanks,

anilrdk: Yes, I understand that part of code executes only under HYBRID. And i am using HYBRID build on RPi platform.
For mc devicetype, it goes through IARM_Bus_Call() and we dont have issue there. 
With this fix, HYBRID devices also work fine.

karuna2git: >With this fix, HYBRID devices also work fine.

Yes. That's what I mentioned above. I can merge it. But in Hybrid device by theory, you are not allowed to change the default interface.

explain me how you are setting the default interface in rPi.? By calling setDefaultInterface() API.? It will  notify to Network Service Manger to swap the default interface. This `NetSrvMgr` is only present in `mediaclient` devices.  Hybrid devices will fail.

anilrdk: Yes, by calling setDefaultInterface(), the interface is changed, we get onDefaultInterfaceChanged() notification also. The IP is also changed, can be confirmed thru ifconfig connected to Eth0 or wlan0.
So is this failure expected in HYBRID devices.

karuna2git: Sounds like some mix n match is being done in rPI.
Since this change is helping you. No harm in merging this.

Im merging it.

karuna2git (approved):

pradeeptakdas:
Internal JIRA for tracking: https://ccp.sys.comcast.net/browse/RDKCOM-3234

PR #2940, opened 2022-07-26
RDKTV-18414: Voice Guidance not announcing activate URL properly

AravindMp22:
Reason for change : The replaceIfIsolated() will replace "/" with "or"
if the text have a punctuation character before or after it. Due to this
the URL scheme "://" will change into ":or/" which affecting the output of TTS.
To solve the issue a new function isUrlScheme() is created. if
the return value of this function is true then / will not replace by or.


varumugam123 (dismissed):

varumugam123:
@AravindMp22 Should we make this change in recent sprint 2208?

anand-ky:

PR #2944, opened 2022-07-27
Update refs to match 2207_sprint

npoltorapavlo:


npoltorapavlo:
in Thunder R2-v1.11 there's an issue with escaping characters:

before:
`"{\"jsonrpc\":\"2.0\",\"method\":\"org.rdk.AVInput.onAVInputActive\",\"params\":{\"url\":\"avin://input0\"}}"`
after:
`"{\"jsonrpc\":\"2.0\",\"method\":\"org.rdk.AVInput.onAVInputActive\",\"params\":{\"url\":\"avin:\\/\\/input0\"}}"`

due to this change:
https://github.com/rdkcentral/Thunder/pull/942

"/" does not need to be escaped

PR #2955, opened 2022-07-28
RDK-37454 : Correct setDeviceVolumeMuteInfo doc

PChandresh:
Reason for change: setDeviceVolumeMuteInfo uses 1 or 0 as arguments,
but currently documented as true or false. Fix the doc as
application developers are confused by this inconsistency

Test Procedure: Ensure that generated documentation is consistent
with the implementation of Bluetooth Plugin
Platforms to test - None

Risks: Medium

Signed-off-by: Chandresh Pitty <Chandresh_Pitty@cable.comcast.com>

### DIFF
--- a/Bluetooth/Bluetooth.json
+++ b/Bluetooth/Bluetooth.json
@@ -221,9 +221,9 @@
             "example":"50"
         },
         "mute": {
-            "summary":"Mute value of the device is either 0 or 1",
+            "summary":"Mute value of the device is either true or false",
             "type": "boolean",
-            "example": "false"
+            "example": false
         }
     },
     "methods": {
@@ -853,7 +853,9 @@
                         "$ref": "#/definitions/volume"
                     },
                     "mute": {
-                        "$ref": "#/definitions/mute"
+                        "summary":"Mute value of the device is either 1 or 0",
+                        "type": "string",
+                        "example": "1"
                     }
                 },
                 "required": [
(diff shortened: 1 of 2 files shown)


PChandresh:
setDeviceVolumeMuteInfo is implemented to expect an integer value as argument for mute -  https://github.com/rdkcentral/rdkservices/blob/sprint/2207/Bluetooth/Bluetooth.cpp#L1639
but the documentation for setDeviceVolumeMuteInfo indicates mute as boolean with 0 or 1

getDeviceVolumeMuteInfo is implemented to return a boolean value for mute status in - https://github.com/rdkcentral/rdkservices/blob/sprint/2207/Bluetooth/Bluetooth.cpp#L745
but the documentation for getDeviceVolumeMuteInfo indicates mute as boolean with 0 or 1

Unfortunately it looks like this was a error performed as part of
https://github.com/rdkcentral/rdkservices/commit/c82eac7d23f9ba9b9a599253efc27497dfec585d

At this moment, I dont want to change the implementation as it might break compatibility with current RA/PLUI deployed in the field across XClass/HiSense/Element devices.
In addition, any change I make in the implementation to correct this consistency, may need a new OTT ticket to correct the mute set/get currently implemented in RA/PLUI.

Use JSON values, which is consistent with the current implementation

    integer with 0 or 1, for mute of setDeviceVolumeMuteInfo.
    boolean with false or true, for mute for getDeviceVolumeMuteInfo.

Documentation updated at this moment to be consistent with the implementation.

    documentation for setDeviceVolumeMuteInfo will be updated so that mute is integer with 0 or 1
    documentation for getDeviceVolumeMuteInfo will be updated so that mute is boolean with false or true

anand-ky (approved):

PR #2965, opened 2022-07-30
SysServices module enabled in git workflow

sergiygladkyy:
Change-Id: Iecc0744866bd8ce7e34d3e7269155bc25c58bcb2


helpers/utils.cpp
```diff
@@ -266,7 +267,8 @@ bool Utils::isPluginActivated(const char* callSign)
 
 bool Utils::getRFCConfig(char* paramName, RFC_ParamData_t& paramOutput)
 {
-    WDMP_STATUS wdmpStatus = getRFCParameter("RDKShell", paramName, &paramOutput);
+    const char* rfcKey = "RDKShell";
+    WDMP_STATUS wdmpStatus = getRFCParameter((char*)rfcKey, paramName, &paramOutput);
```
npoltorapavlo, line 271: Can pass nullptr?

helpers/UtilsLogging.h
```diff
@@ -2,6 +2,13 @@
 
 #include <syscall.h>
 
+#undef LOGINFO
```
npoltorapavlo, line 5: ?

Tests/headers/devicesettings/sleepMode.hpp
```diff
@@ -0,0 +1,26 @@
+
+#pragma once
+#include <vector>
+namespace device {
+
+    template <class T>
+    using List = std::vector<T>;
```
npoltorapavlo, line 7: please use list.hpp

Tests/headers/devicesettings/sleepMode.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+
+    class SleepMode /*: public DSConstant*/ {
+        public:
+            static const int kLightSleep;
+            static const int kDeepSleep;
```
npoltorapavlo, line 13: I think plugins don't use these constants.

Tests/headers/devicesettings/securityagent/SecurityTokenUtil.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
```
npoltorapavlo, line 1: SecurityTokenUtil.h is not used

Tests/headers/ccec/drivers/CecIARMBusMgr.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#ifdef __cplusplus
+extern "C"
```
npoltorapavlo, line 4: Do we need extern "C"?

SystemServices/SystemServices.cpp
```diff
@@ -1657,7 +1655,7 @@ namespace WPEFramework {
         {
             bool status = false;
             try {
-                const device::SleepMode &mode = device::Host::getInstance().getPreferredSleepMode();
+                device::SleepMode mode = device::Host::getInstance().getPreferredSleepMode();
```
npoltorapavlo, line 1658: The object should be mockable. For example, device::SleepMode can wrap a pointer to a mock. Another approach is std::reference_wrapper as polymorphism works for references but this changes the type which should probably be avoided.

SystemServices/SystemServices.cpp
```diff
@@ -39,7 +39,7 @@
 
 #include "SystemServices.h"
 #include "StateObserverHelper.h"
-#include "utils.h"
+//#include "utils.h"
```
npoltorapavlo, line 42: I think SystemServices.h should NOT have "utils.h" but here it is ok.

npoltorapavlo (changes requested):

PR #2976, opened 2022-08-03
RDKTV-18587: Observed stopMaintenance Crash with MM

PaviP98:
stopMaintenance Api is not returning anything when stopMaintenance
RFC is disabled which is causing this crash since it receives empty
output from server.


mvalas827 (approved):
QA should have caught this early even if dev misses it. Change looks good.

anand-ky (approved):

PR #2985, opened 2022-08-03
DELIA-57482: Cleanup Versioning for few RDKServices APIs to support Semantic Versioning

mateusz-adamski-red:


DisplaySettings/DisplaySettings.cpp
```diff
@@ -150,113 +168,114 @@ namespace WPEFramework {
...
-            registerMethod("readEDID", &DisplaySettings::readEDID, this);
-            registerMethod("readHostEDID", &DisplaySettings::readHostEDID, this);
-            registerMethod("getActiveInput", &DisplaySettings::getActiveInput, this);
-            registerMethod("getTvHDRSupport", &DisplaySettings::getTvHDRSupport, this);
-            registerMethod("getSettopHDRSupport", &DisplaySettings::getSettopHDRSupport, this);
-            registerMethod("setVideoPortStatusInStandby", &DisplaySettings::setVideoPortStatusInStandby, this);
-            registerMethod("getVideoPortStatusInStandby", &DisplaySettings::getVideoPortStatusInStandby, this);
-            registerMethod("getCurrentOutputSettings", &DisplaySettings::getCurrentOutputSettings, this);
-
+            RegisterMethod(this, {1,2}, "getConnectedVideoDisplays", &DisplaySettings::getConnectedVideoDisplays, this);
```
anand-ky, line 171: We don't need these changes. the default registerMethod() takes care of making this API available on all available versions. So only the APIs that are registered on versions 2 or above using GetHandler() need to be made available on version 1. So basically replace GetHandler(2)->registerMethod() with registerMethod().

SystemServices/SystemServices.cpp
```diff
@@ -317,117 +316,114 @@ namespace WPEFramework {
              * @brief Invoking Plugin API register to WPEFRAMEWORK.
              */
 #ifdef DEBUG
-            registerMethod("sampleSystemServiceAPI", &SystemServices::sampleAPI, this);
+            RegisterMethod(this, {1,2}, "sampleSystemServiceAPI", &SystemServices::sampleAPI, this);
 #endif /* DEBUG */
-            registerMethod("getDeviceInfo", &SystemServices::getDeviceInfo, this);
+            RegisterMethod(this, {1,2}, "getDeviceInfo", &SystemServices::getDeviceInfo, this);
```
anand-ky, line 321: We don't need these changes. the default registerMethod() takes care of making this API available on all available versions. So only the APIs that are registered on versions 2 or above using GetHandler() need to be made available on version 1. So basically replace GetHandler(2)->registerMethod() with registerMethod().

anand-ky (changes requested):

PR #2989, opened 2022-08-04
Wrap mocks to keep the types intact

npoltorapavlo:


sergiygladkyy (approved):
Confirming this fixed the ds problem when building SystemServices

anand-ky (approved):

PR #2995, opened 2022-08-05
RDKTV-18414: VG not announcing activate URL properly

AravindMp22:
Reason for change: The replaceIfIsolated() will replace "/" with "or"
if the text have a punction character before or after it. Due to this
the URL scheme "://" will change into ":or/" which affecting the output
of TTS.
To solve the issue if the text contains "://" then it replace by
" colon slash slash " using ReplaceIfIsolated() Fn. And also a bool variable
is added as the default parameter for ReplaceIfIsolated() Fn.

Test Procedure: check the output of TTS with diffrent URL.

Risks: LOW

Signed-off-by: Aravind MP <aravind.m@tataelxsi.co.in>

### DIFF
--- a/TextToSpeech/impl/TTSURLConstructer.cpp
+++ b/TextToSpeech/impl/TTSURLConstructer.cpp
@@ -123,13 +123,13 @@ std::string TTSURLConstructer::constructURL(TTSConfiguration &config,std::string
     return tts_request;
 }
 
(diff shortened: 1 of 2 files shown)


varumugam123 (approved):

varumugam123 (commented):
Approving based on Aravind's observation that any '/' in the rest of the URL would be properly announced as "slash". So this change specifically replaces only the "://" part to be announced properly

ramasat (approved):

PR #3002, opened 2022-08-08
DELIA-57482: Cleanup Versioning for few RDKServices APIs to support S…

mateusz-adamski-red:
…emantic Versioning


anand-ky (approved):

anand-ky:
These changes are done to ensure that APIs registered in JSONRPC handlers 2 and above are also available in version 1 as we move to Semantic Versioning support in RDKServices. Going forward any new APIs can be added on the default JSONRPC version handler (version 1) but we need to update the plugin version as explained in https://github.com/rdkcentral/RDKServices/tree/sprint/2207#versioning. 

Also APIs can now be called without requiring the version to be passed (org.rdk.System.getDeviceInfo instead of org.rdk.System.1.getDeviceInfo).

npoltorapavlo:
This pr adds the "easy functions" which I hope will never be used. Someone may wrongly interpret it and put non-1 things into the "easy" RegisterMethodVersions which would be sad. Some comment would be good.

anand-ky:
Yes, going forward we should just use the default Register() for new APIs. 
@mateusz-adamski-red - Pls add some comments in UtilsVersions.h on when RegisterMethodVersions should be used.

PR #3008, opened 2022-08-09
LLAMA-6991: Fix network service

mateusz-adamski-red:


Network/Network.cpp
```diff
@@ -486,15 +486,15 @@ namespace WPEFramework
                     string interface = "";
                     bool persist = false;
 
-                    getStringParameter("interface", interface)
+                    getDefaultStringParameter("interface", interface, "")
```
anand-ky, line 489: This is already being done outside. I understand it's good to do it within this getDefaultStringParameter API but checking how does this help with the crash.

Network/Network.cpp
```diff
@@ -486,15 +486,15 @@ namespace WPEFramework
...
 
                         if (!(strcmp (interface.c_str(), "ETHERNET") == 0 || strcmp (interface.c_str(), "WIFI") == 0))
                         {
                             LOGERR ("Call for %s failed due to invalid interface [%s]", IARM_BUS_NETSRVMGR_API_setDefaultInterface, interface.c_str());
                             returnResponse (result)
                         }
 
-                    getBoolParameter("persist", persist)
+                    getDefaultBoolParameter("persist", persist, false)
```
anand-ky, line 497: This is already being done outside. I understand it's good to do it within this getDefaultBoolParameter API but checking how does this help with the crash.

anand-ky (changes requested):
Some of these checks are valid but some are redundant since it's being done outside.

PR #3015, opened 2022-08-10
RDKTV-18587: Observed stopMaintenance Crash with MM

PaviP98:
stopMaintenance Api is not returning anything when stopMaintenance
RFC is disabled which is causing this crash since it receives empty
output from server.


anand-ky (approved):

anand-ky:
@PaviP98 - We have enabled semantic versioning for RDKServices. Please update the patch version and also CHANGELOG.md. You can refer to https://github.com/rdkcentral/rdkservices/pull/3028/files that was done for RDKShell change.

More details on Semantic Versioning at https://github.com/rdkcentral/rdkservices#versioning

anand-ky (changes requested):
Needs Changes to update versioning as mentioned above. Pls do the same for OTA_5.

anand-ky (approved):

PR #3019, opened 2022-08-11
RDKTV-18791, LLAMA-7438, LLAMA-7431: std::terminate crash with MM in …

PaviP98:
…startMaintenance API

Joined the thread before executing tasks from startMaintenance API.
Whenever device has no internet connectvity, unsolicitated maintenance
was performed and thread was not yet joined. so whenever we invoke
startMaintenance API, it is better to do join before creating m_thread.


MaintenanceManager/CHANGELOG.md
```diff
@@ -5,10 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.1] - 2022-08-12
```
anand-ky, line 8: Pls update the version for https://github.com/rdkcentral/rdkservices/pull/3015/files first since it needs to first go to main and OTA_5. So update it to 1.0.1 and bring it back to sprint/2208.

You can then update this change to 1.0.2

anand-ky (changes requested):

PR #3066, opened 2022-08-18
Group external test headers (same to sysroot)

npoltorapavlo:


npoltorapavlo:
To support for example `#include "rdk/iarmbus/libIBus.h"` and `#include "libIBus.h"` (both are used)

anand-ky (approved):

PR #3077, opened 2022-08-20
RDK-37645 : xControl Cherry 'Watch Together' Development

dwolaver:
Reason for change: add support for cherry phase 1
Test Procedure: Verification build in https://gerrit.teamccp.com/#/q/topic:RDK-37645
Risks: Low
Signed-off-by: David Wolaver <david_wolaver@cable.comcast.com>


npoltorapavlo:
APIs should follow the JSON-RPC spec https://www.jsonrpc.org/specification#response_object
Please use error codes instead of the XRE Service Manager legacy param "success"

dwolaver:
> APIs should follow the JSON-RPC spec https://www.jsonrpc.org/specification#response_object Please use error codes instead of the XRE Service Manager legacy param "success"

I've opened DELIA-57705 to address compliance.  For now, the API should remain consistent with respect to return code.  Removing the "success" field would cause a breaking change to the API.  It will have to be deprecated first and then wait at least two RDK releases before removing it.

anand-ky (approved):

PR #3050, opened 2022-08-16
RDK-36108: Add JSPP support

Scony:
On behalf of @jack-howgate and @filipe-norte-red - adding JSPP support required by webkit 2.28 upgrade activities

PR #3080, opened 2022-08-22
RDK-36108: Set JSPP environment variables

Scony:
This PR is a followup of https://github.com/rdkcentral/rdkservices/pull/3050 which implements and uses a mechanism for exporting extra environment variables from webkitbrowser plugin.


emutavchi (dismissed):

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -625,6 +627,7 @@ static GSourceFuncs _handlerIntervention =
             Core::JSON::Boolean AllowFileURLsCrossAccess;
             Core::JSON::Boolean SpatialNavigation;
             Core::JSON::EnumType<HTTPCookieAcceptPolicyType> CookieAcceptPolicy;
+            Core::JSON::ArrayType<JsonObject> EnvVars;
```
MFransen69, line 630: Let's please use a strong type here and not JsonObject. That is easier to understand and type safe (and more efficient)

MFransen69 (changes requested):

MFransen69 (approved):

Issue #2847, opened 2022-07-05
servicemanager legacy code


npoltorapavlo:
"success" param in the api result is the legacy code.

In Thunder JSON-RPC there are error codes. If ERROR_NONE the message sends a "result" object. Otherwise the message sends an error code. Error codes are part of the interface: https://github.com/rdkcentral/rdkservices/blob/df8b598e1943e0f35d7af8cf77943e034b278a1d/Messenger/Messenger.json#L76-L77

A caller shouldn't have to inspect "result" object to see if the api succeeded or not.

"getMeSomeData" methods are also the legacy code.

In Thunder JSON-RPC there are properties / readonly properties: https://github.com/rdkcentral/rdkservices/blob/df8b598e1943e0f35d7af8cf77943e034b278a1d/LocationSync/LocationSyncJsonRpc.cpp#L36

anand-ky:
"success" parameter was added for legacy reasons to support comcast ServiceManager APIs. But going forward, newer APIs should return only error object to match the spec. Also existing APIs should return error object along with success:false in the response.

npoltorapavlo:
management preferred to stay with xre ServiceManager style so closing

PR #3089, opened 2022-08-24
Common code fixes

npoltorapavlo:


npoltorapavlo:
with [6d1a043](https://github.com/rdkcentral/rdkservices/pull/3089/commits/6d1a043cd39e5b451338653be64eb4776f8c1377) js promise may throw if doesn't handle failure, in this case, it is a bug of js app because failure may occur for other reasons too. If the c++ app doesn't check the return code but checks the "success" boolean, it's false if the response object is empty.

npoltorapavlo:
for #2847

anand-ky (approved):

npoltorapavlo:
for https://ccp.sys.comcast.net/browse/DELIA-57976

PR #3093, opened 2022-08-24
RDKTV-19132 : Update SystemServices to call container_setup.sh script…

ks734:
… after deleting persistent data

Reason for change : WPEFramework crashes clearing local storage
Test Procedure: Refer Jira
Risks: Low

Signed-off-by: Karthick S karthick_s@comcast.com

### DIFF
--- a/SystemServices/CHANGELOG.md
+++ b/SystemServices/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [1.0.1] - 2022-08-24
+## [1.0.2] - 2022-08-24
 ### Fixed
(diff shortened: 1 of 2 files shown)


SystemServices/CHANGELOG.md
```diff
@@ -5,10 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.1] - 2022-08-24
```
anand-ky, line 8: Pls retain the same version as main branch [1.0.2] so that we can tie the version to this change. I understand this will not do version updates in sequence but it will eventually catch up (once OTA_6 gets branched from stable2)

anand-ky (changes requested):

SystemServices/CHANGELOG.md
ks734: Done

anand-ky (approved):

Issue #2537, opened 2022-04-29
Inter-plugin communication


npoltorapavlo:
Some RDK services use **JSONRPC::LinkType** for inter-plugin communication - DisplaySettings, MaintenanceManager, RDKShell, ScreenCapture, SystemServices, DeviceProvisioning, EasterEgg.

JSON-RPC is a web socket with a security token and JSON. It's used for JS-C++, or Shell-C++. Thunder exports **JSONRPC::LinkType** for C++ apps, however, some RDK services use it, even for inter-thread comm. in the same plugin:

https://github.com/rdkcentral/rdkservices/blob/d3bb5c75aadedecc545bb970e88613ee090b4075/RDKShell/RDKShell.cpp#L514

Plugin querying via an interface - **IShell::QueryInterface** (same process), or **IShell::Root** (same process or inter-process) - are better solutions for inter-plugin communication than a web socket connection with JSON.

npoltorapavlo:
https://ccp.sys.comcast.net/browse/DELIA-56040 (ScreenCapture)
https://ccp.sys.comcast.net/browse/DELIA-56041 (MaintenanceManager)
https://ccp.sys.comcast.net/browse/DELIA-56042 (DisplaySettings)
https://ccp.sys.comcast.net/browse/DELIA-56043 (RDKShell)

npoltorapavlo:
[DELIA-56040 ](https://ccp.sys.comcast.net/browse/DELIA-56040) fixed in 2205 (#2635), 2206 (#2719), and main (#2708)

npoltorapavlo:
[DELIA-56776](https://ccp.sys.comcast.net/browse/DELIA-56776) fixed in 2206 (#2824), 2207 (#2848), and main (#2849)

npoltorapavlo:
at this point, JSONRPC::LinkType is used in:

- DisplaySettings (15)
   - org.rdk.HdmiCecSink.setupARCRouting
   - org.rdk.HdmiCecSink.getEnabled
   - org.rdk.HdmiCecSink.getAudioDeviceConnectedStatus
   - org.rdk.HdmiCecSink.sendAudioDevicePowerOnMessage
   - org.rdk.HdmiCecSink.requestShortAudioDescriptor
   - org.rdk.HdmiCecSink.requestAudioDevicePowerStatus
   - org.rdk.HdmiCecSink.arcInitiationEvent
   - org.rdk.HdmiCecSink.arcTerminationEvent
[...]

PR #3125, opened 2022-08-30
DELIA-57822: JSON-RPC misuse

npoltorapavlo:
Reason for change:
* MM checkNetwork() checks org.rdk.Network active
  before invoking org.rdk.Network.isConnectedToInternet.
  checkNetwork() returns false if inactive or
  isConnectedToInternet fails/is false,
  so the check is redundant.
* MM checkActivatedStatus() checks org.rdk.AuthService active.
  MM doesn't activate it and org.rdk.AuthService
  doesn't auto start so this might not work or
  relies on external factor.
  Not attempting to fix that.
  org.rdk.AuthService should be used via
  QueryInterfaceByCallsign,
  this suffices to check if it's active.
* MM acquires token but never uses it, this code is redundant.
* Remove UtilsSecurityToken.h, UtilsSecurityToken.cpp, and
  SecurityUtil from MM.
* RDKShell checks org.rdk.System active.
  RDKShell doesn't activate it and org.rdk.System
  doesn't auto start so this might not work or
  relies on external factor.
  Not attempting to fix that.
  org.rdk.System should be used via
  QueryInterfaceByCallsign,
  this suffices to check if it's active.
* SecurityAgent should be used via
  QueryInterfaceByCallsign.
* RDKShell used 2 variants of getThunderControllerClient(),
  one variant should be enough.
* Remove UtilsSecurityToken.h, UtilsSecurityToken.cpp, and
  SecurityUtil from RDKShell.
* DisplaySettings indefinitely checks org.rdk.HdmiCecSink
  active. On several platforms org.rdk.HdmiCecSink
  doesn't exist hence log flooding.
  DisplaySettings relies on race conditions in
  org.rdk.HdmiCecSink using sleep(...).
  Not attempting to fix that.
  org.rdk.HdmiCecSink should be used via
  QueryInterfaceByCallsign,
  this suffices to check if it's active.
* DisplaySettings attempts to activate org.rdk.HdmiCecSink.
  This looks like design flaw.
  DisplaySettings activates org.rdk.HdmiCecSink in
  blocking way.
  Not attempting to fix that.
  Controller IDispatcher interface can activate.
* DisplaySettings attempts to activate org.rdk.HdmiCecSink
  twice in onTimer(), the second one is redundant because
  timer will reschedule.
* Remove UtilsSecurityToken.h, UtilsSecurityToken.cpp, and
  SecurityUtil from DisplaySettings.
* Delete UtilsSecurityToken.h, UtilsSecurityToken.cpp.

Test Procedure:
* MM logs "AuthService is active".
* RDKShell logs "SystemService is already activated".
* RDKShell logs "RDKShell got security token".
* RDKShell regression test.
* DisplaySettings logs "org.rdk.HdmiCecSink is active".


npoltorapavlo:
for #2537

npoltorapavlo:
a link to successful gerrit verification https://gerrit.teamccp.com/#/c/675391/

apatel859 (approved):

PR #3171, opened 2022-09-12
Sprint 2208 delia 57501 2

srikanth-vv:
- Suppressed warnings "deprecated" for HdmiCec, HdmiCecSink: They use "recipe-sysroot/usr/include/ccec/include/ccec/Driver.hpp" header file which gives error as "error: dynamic exception specifications are deprecated in C++11" fir functions like "virtual void  open(void) throw(InvalidStateException, IOException) = 0;"

- Suppressed warnings "attribute-warning" for Warehouse: It uses system(). Since this rdkservices is open source, it cannot be replaced with v_secure_system().

### DIFF
--- a/Warehouse/CMakeLists.txt
+++ b/Warehouse/CMakeLists.txt
@@ -31,6 +31,7 @@ set_target_properties(${MODULE_NAME} PROPERTIES
         CXX_STANDARD_REQUIRED YES)
 
 target_compile_definitions(${MODULE_NAME} PRIVATE MODULE_NAME=Plugin_${PLUGIN_NAME})
+target_compile_options(${MODULE_NAME} PRIVATE -Wno-error=attribute-warning)
 
 find_package(DS)
 if (DS_FOUND)
--- a/ControlService/test/controlSvcTestClient.cpp
+++ b/ControlService/test/controlSvcTestClient.cpp
@@ -361,7 +361,7 @@ int main(int argc, char** argv)
     std::cout << "Using callsign: " << SYSSRV_CALLSIGN << std::endl;
 
     if (NULL == remoteObject) {
-        remoteObject = new JSONRPC::Client(_T(SYSSRV_CALLSIGN), _T(""), false, sToken);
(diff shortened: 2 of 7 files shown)


Warehouse/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ set_target_properties(${MODULE_NAME} PROPERTIES
         CXX_STANDARD_REQUIRED YES)
 
 target_compile_definitions(${MODULE_NAME} PRIVATE MODULE_NAME=Plugin_${PLUGIN_NAME})
+target_compile_options(${MODULE_NAME} PRIVATE -Wno-error=attribute-warning)
```
npoltorapavlo, line 34: please remove changes to all CMakeLists.txt in this pull request. The task is to fix warnings, and not to hide them selectively

npoltorapavlo (changes requested):

Warehouse/CMakeLists.txt
anand-ky: @npoltorapavlo - Some of these need to be handled outside in the HAL before we can enable warnings as error. Srikanth is working with Amit to get that addressed. In the meantime, I would prefer to enable it asap on the recipe and have these exceptions for now. We will remove it once the HAL changes are done.

srikanth-vv: @npoltorapavlo - I'll create different tickets to handle the suppressed errors and will remove these changes to CMakeLists.txt at that time.

npoltorapavlo (approved):

anand-ky (approved):

Warehouse/CMakeLists.txt
npoltorapavlo: this broke the build on XG1V3
```
18:43:37 | cc1plus: error: -Werror=attribute-warning: no option -Wattribute-warning
```

srikanth-vv: I am seeing the same thing on PACEXID, Ciscoxid also. I think they use different compiler versions that doesn't know about "attribute-warning" option.

PR #3175, opened 2022-09-13
[OCDM] Enhancements

sramani-metro:
1. Support passing Encryption Scheme and Pattern to Decrypt
     2. New Gstreamer Adapter method that parses Sample and Subsample info
        from GstStructure in ProitectionMetdata of GstBuffer
     3. New gstreamer adapter API supports passing Subsample Info to the DRM
        implementation without processing it.


anand-ky:

PR #3181, opened 2022-09-14
WebKitBrowser: Fix cookie crypto jar linkage

emutavchi:
Older versions of cmake don't support target_link_options.


modeveci (approved):
LGTM

PR #3199, opened 2022-09-19
RDKDEV-259 :Security agent : Bug fix in security agent plugin's configuration and url for token creation in injected bundle (v3)

rjpdasilva:

This PR is the rebased version of https://github.com/rdkcentral/rdkservices/pull/2636.

As mentioned in https://github.com/rdkcentral/rdkservices/pull/2636#discussion_r974094731, this is still needed taking into consideration the current implementation of the `SecurityAgent` plugin.

### DIFF
--- a/WebKitBrowser/InjectedBundle/SecurityAgent.cpp
+++ b/WebKitBrowser/InjectedBundle/SecurityAgent.cpp
@@ -46,6 +46,8 @@ namespace JavaScript {
                     uint8_t buffer[2 * 1024];
 
                     std::string url = WebKit::Utils::GetURL();
+                    //Formatting the url string to json for creating token
+                    url = "{\"url\":\"" + url + "\"}";
 
                     std::string tokenAsString;
                     if (url.length() < sizeof(buffer)) {
--- a/SecurityAgent/SecurityAgent.config
+++ b/SecurityAgent/SecurityAgent.config
@@ -1,6 +1,7 @@
 set (autostart true)
+set(PLUGIN_SECURITYAGENT_ACL_FILE_NAME "acl.json" CACHE STRING "SecurityAgent ACL file name")
 
 map()
-    map(acl acl.json)
+    kv(acl ${PLUGIN_SECURITYAGENT_ACL_FILE_NAME})
 end()
 ans(configuration)


anand-ky:
@npoltorapavlo - Pls review

WebKitBrowser/InjectedBundle/SecurityAgent.cpp
```diff
@@ -46,6 +46,8 @@ namespace JavaScript {
                     uint8_t buffer[2 * 1024];
 
                     std::string url = WebKit::Utils::GetURL();
+                    //Formatting the url string to json for creating token
+                    url = "{\"url\":\"" + url + "\"}";
```
npoltorapavlo, line 50: when building for RDK we use patch `0002-RDK-28534-Comcast-Security-Agent-changes.patch` which enforces token to be URL:
```
-        _context.FromString(_token);
+        _context.URL = _token;
```
I'm expecting that your change would break this.
If you want this code to conform to "expected" SecurityContext behavior, you would need to update the patch as well.

rjpdasilva: Yes, it conflicts with that patch.

Can you point me to it and please explain why was the patch needed and the reason behind it?  

Could it be that it was somehow to make the plugin aligned with the WebKitBrowser injected bundle for the SecurityAgent and also with the SecurityAgent tools?  

In my view, the original code in the plugin should be fine, as you may want to have a security context with more than just the URL, so I think the changes should be done instead in the WebKitBrowser and the SecurityAgent tool.  
However, if you feel otherwise, then you can close this PR (we'll keep using it though).

npoltorapavlo: https://gerrit.teamccp.com/plugins/gitiles/rdk/yocto_oe/layers/meta-rdk-video/+/2208_sprint/recipes-extended/rdkservices/files/0002-RDK-28534-Comcast-Security-Agent-changes.patch

@anand-ky why do we need security context to be just URL?

rjpdasilva: @npoltorapavlo, I can't reach the link you've shared, seems to require some Comcast authentication.
Is it available i RDK CMF?

anand-ky: yes the changes were done to align SecurityAgent with the WebKitBrowser Injected Bundle. It makes sense to make it a json to support more parameters potentially in the future. If we can change both together now, we should be good.
@npoltorapavlo - It would be also good to upstream and remove [0002-RDK-28534-Comcast-Security-Agent-changes.patch](https://gerrit.teamccp.com/plugins/gitiles/rdk/yocto_oe/layers/meta-rdk-video/+/2208_sprint/recipes-extended/rdkservices/files/0002-RDK-28534-Comcast-Security-Agent-changes.patch)

npoltorapavlo: @anand-ky there are other places where URL is used,
https://github.com/rdkcentral/rdkservices/blob/720a0c4d0229cee479bf7b74367aad708549c07d/Messenger/MessengerSecurity.cpp#L91 or https://gerrit.teamccp.com/plugins/gitiles/rdk/yocto_oe/layers/meta-rdk-video/+/2208_sprint/recipes-extended/rdkservices/files/0001-RDK-28534-Security-Agent-Utility-and-Logging-ClientLibs.patch#162

To upstream, research is needed to understand where URL is used.

npoltorapavlo: @rjpdasilva can you please keep only the ACL file name change. Upstream and change URL to json is not so easy and can be done independently https://ccp.sys.comcast.net/browse/RDKCOM-3132. Please resubmit for sprint/22Q4

anand-ky:
We are working on getting this resubmitted fully by upstreaming the patches and will bring them as part of https://ccp.sys.comcast.net/browse/RDKCOM-3132

PR #3207, opened 2022-09-20
DELIA-58055: Change the GetConnectedSSID to call netsrvmgr

ksanth850:
Reason for change:
In-order to help Sky team and not to flood the wpeframework.log, it is decided that the WiFiManager plugin will be updated as below,
1. The Signal Threshold Monitoring API will not call getConnectedSSID API and instead check only the RSSI value locally.
2. The public API getConnectedSSID will NOT be using the caching logic and instead always calls NetSrvMgr. Fix above issue:
By doing the above 2, we should be able to see the Telemetry marker in the netsrvmgr.log file. Test Procedure:
Risks: High


WifiManager/impl/WifiManagerSignalThreshold.cpp
```diff
@@ -30,14 +30,41 @@ namespace {
...
+
+    std::string retrieveValues(const char *command, char *output_buffer, size_t output_buffer_size)
+    {
+        std::string key, value;
+        std::string rssi = "";
+
+        FILE *fp = popen(command, "r");
+        if (!fp)
+        {
+            LOGERR("Failed in getting output from command %s \n",command);
```
karuna2git, line 44: If popen fails, it must return from here..  
Otherwise, The below while loop end up in accessing the null pointer and the pclose(NULL) etc..

karuna2git (changes requested):

karuna2git (approved):

PR #3215, opened 2022-09-22
Update SystemRootPath value for downloadable plugin and store it in p…

PaviP98:
…ersistent

area to set the LD_LIBRARY_PATH dynamically.


tdeva14:
Changes look good

PR #3217, opened 2022-09-22
Browser deactivation logging

nrajan002c:
Signed-off-by: nganes961 <nambirajan_ganesan@comcast.com>

### DIFF
--- a/WebKitBrowser/WebKitBrowser.cpp
+++ b/WebKitBrowser/WebKitBrowser.cpp
@@ -309,6 +309,7 @@ namespace Plugin {
     {
         string message(string("{ \"url\": \"") + URL + string("\", \"loaded\": ") + (loaded ? string("true") : string("false")) + string(" }"));
         TRACE(Trace::Information, (_T("URLChanged: %s"), message.c_str()));
+        _lastURL = URL;
         _service->Notify(message);
         Exchange::JWebBrowser::Event::URLChange(*this, URL, loaded);
     }
@@ -350,6 +351,7 @@ namespace Plugin {
     void WebKitBrowser::Deactivated(RPC::IRemoteConnection* connection)
     {
         if (connection->Id() == _connectionId) {
+            TRACE(Trace::Error, (_T("WebKitBrowser::Deactivated: { \"URL\": %.*s }"), 80, _lastURL.c_str()));
 
             ASSERT(_service != nullptr);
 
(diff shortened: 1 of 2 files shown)


WebKitBrowser/WebKitBrowser.cpp
```diff
@@ -350,6 +351,7 @@ namespace Plugin {
     void WebKitBrowser::Deactivated(RPC::IRemoteConnection* connection)
     {
         if (connection->Id() == _connectionId) {
+            TRACE(Trace::Error, (_T("WebKitBrowser::Deactivated: { \"URL\": %s }"), _lastURL.substr(0,80).c_str()));
```
emutavchi, line 354: I think You can avoid a mem copy with '%.*s' format and specifying limit just before _lastURL.c_str(). i.e.:

```
TRACE(Trace::Error, (_T("WebKitBrowser::Deactivated: { \"URL\": %.*s }"), 80, _lastURL.c_str()));
```

could you please give it a try?

nrajan002c: @emutavchi , the above format is working fine.

emutavchi (approved):
Looks good to me. @MFransen69  could you please take a look?

WebKitBrowser/WebKitBrowser.cpp
```diff
@@ -309,6 +309,7 @@ namespace Plugin {
     {
         string message(string("{ \"url\": \"") + URL + string("\", \"loaded\": ") + (loaded ? string("true") : string("false")) + string(" }"));
         TRACE(Trace::Information, (_T("URLChanged: %s"), message.c_str()));
+        _lastURL = URL;
```
MFransen69, line 312: I would only do the string copy (relatively expensive, certainly for bigger strings) when tracing is enabled for this category (the TRACE_ENABLED(CATEGORY) macro can be used to find that out).

And although I do not see any problem with this PR in itself is what is the actual use case if I may ask? I guess tracing the url that caused Webkit to crash? Or would the error tracing category be always enabled so automated action can be taken?

MFransen69 (changes requested):

WebKitBrowser/WebKitBrowser.cpp
nrajan002c: @MFransen69 
The reason for this implementation is, to track the browser crashes by app URL. 
Use case is, We need to know the number of crashes per app basis. 
And we need this log to be printed always.
I can convert (_lastURL =URL) to strcpy.

MFransen69: Ah, okay in that case I think you do not want to use Tracing but SYSLOG, correct? Trace lines will only be outputted when for that module and category they are turned on, therefore I suggested to only do the string assignment (which is the same as a strcpy, std::string is not refcounted) when tracing is actually turned on here,  it is off by default. 

But seen your usecase is it perhaps an idea to add this functionality to the PerformanceMetrics plugin, that seems a more logical place then? All the info you need would be available there.
I would be happy to add that functionality there or to assist you in adding it there.

anand-ky:

PR #3224, opened 2022-09-23
RDK-37578: More tests for System

npoltorapavlo:
Reason for change:
- Tests for getStoreDemoLink, deletePersistentPath, onSystemPowerStateChanged, onNetworkStandbyModeChanged, onRebootRequest.
- Fix segfault in ServiceMock. QueryInterfaceByCallsign uses reinterpret_cast. This won't work if the object has multiple inheritance. ServiceMock should not implement ICOMLink, whoever did that was in a rush.

Comments:
- "getPlatformConfiguration" returns too many unrelated or uncategorized properties, this is an example of an anti-pattern and a code smell. Should not test that.
- "uploadLogs" is hard to test. Also, not sure if it's used. Should not test that.

Test Procedure: Unit tests
Risks: Low

### DIFF
--- a/Tests/mocks/COMLinkMock.h
+++ b/Tests/mocks/COMLinkMock.h
@@ -1,0 +1,18 @@
+#ifndef RDKSERVICES_TESTS_MOCKS_COMLINKMOCK_H_
+#define RDKSERVICES_TESTS_MOCKS_COMLINKMOCK_H_
+
+#include <gmock/gmock.h>
+
+#include "Module.h"
+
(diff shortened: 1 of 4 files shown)


npoltorapavlo:
SystemServices.cpp coverage before/after:
47.6 % | 975 / 2047 | 60.8 % | 73 / 120
51.9 % | 1063 / 2047 | 65.8 % | 79 / 120

anand-ky (approved):

PR #3226, opened 2022-09-26
Update SystemRootPath value for downloadable plugins and store it in the

PaviP98:
persistent path to set the LD_LIBRARY_PATH dynamically.


tdeva14:
Changes look fine

PR #3232, opened 2022-09-27
RDKCMF-8858 RDKShell: also check for gRdkShellSurfaceModeEnabled in l…

sverkoye:
…aunchApplication

* launch() already checks for gRdkShellSurfaceModeEnabled and uses rdkshell_display in that case. Do the same for launchApplication().


mfiess (approved):

pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3314

PR #3235, opened 2022-09-27
RDK-36673,38098:Generating event section automatically, link for Depr…

rekhajp98:
…ecated APIs

"RDK-38098: Provide link to Alternative API/s to use for Deprecated APIs
RDK-36673 : event section should be automatically generated"

Reason for change:
**RDK-38098**: APIs can be deprecated using the "deprecated" field in json schema. This would mark the API as deprecated in documentation. An alternate API to be used in place of this deprecated API. This alternate API is given as a link to the documentation for the new API. This new API is added in json file in thr method with key as "referenceUrl" that has deprecated value as "true"
**RDK-36673** : Currently Event section is not defined as a separate object instead it is hooked in the Summary part inside each method in Json file. This event section should be defined as a separate object inside each Method with an hyperlink to the existing event.

Test Procedure: Executed and verified by generating api documentation in docsify
Risks: Create None
Signed-off-by: Rekha Jayaram "[Rekha_Jayaram3@comcast.com](mailto:Rekha_Jayaram3@comcast.com)"

Jira link :
https://ccp.sys.comcast.net/browse/RDK-38098

### DIFF
--- a/AVInput/AVInput.json
+++ b/AVInput/AVInput.json
@@ -39,7 +39,7 @@
             }
         },
         "currentVideoMode": {
-            "summary": "Returns a string encoding the video mode being supplied by the device currently attached to the HDMI input. The format of the string is the same format used for the `resolutionName` parameter of the XRE `setResolution` messages. HDMI input is presentable if its resolution is less than or equal to the current Parker display resolution. \n \n### Events\n \n No Events.",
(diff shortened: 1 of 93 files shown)


anand-ky (approved):

anand-ky:
No Code changes. Just Doc changes to generate events section.

PR #3243, opened 2022-09-29
RDK-37900 : Unit Tests for HdcpProfile and HdmiInput profiles.

rajnishdey:
Signed-off-by: Rajnish Dey <Rajnish_Dey4@comcast.com>

### DIFF
--- a/HdmiInput/HdmiInput.cpp
+++ b/HdmiInput/HdmiInput.cpp
@@ -172,9 +172,8 @@ namespace WPEFramework
             int portId = 0;
             try {
                 portId = stoi(sPortId);
-            }catch (...) {
-		    LOGWARN("Invalid Arguments");
-		    response["message"] = "Invalid Arguments";
+            }catch (const std::exception& err) {
+		    LOGWARN("sPortId invalid paramater: %s ", sPortId.c_str());
 		    returnResponse(false);
             }
             bool success = true;
@@ -252,9 +251,8 @@ namespace WPEFramework
                            h = std::stoi(parameters["h"].String());
                        }
 		}
-                catch (...) {
-		    LOGWARN("Invalid Arguments");
-		    response["message"] = "Invalid Arguments";
+                catch (const std::exception& err) {
+		    LOGWARN("Invalid paramater X: %s,Y: %s, W: %s, H:%s ", parameters["x"].String().c_str(),parameters["y"].String().c_str(),parameters["w"].String().c_str(),parameters["h"].String().c_str());
 		    returnResponse(false);
                 }
 
@@ -326,9 +324,8 @@ namespace WPEFramework
             int portId = 0;
             try {
                 portId = stoi(sPortId);
-            }catch (...) {
-		    LOGWARN("Invalid Arguments");
-		    response["message"] = "Invalid Arguments";
+            }catch (const std::exception& err) {
+		    LOGWARN("sPortId invalid paramater: %s ", sPortId.c_str());
 		    returnResponse(false);
             }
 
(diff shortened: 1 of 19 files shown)


anand-ky:
@npoltorapavlo - Pls review

HdmiInput/HdmiInput.cpp
```diff
@@ -164,11 +172,11 @@ namespace WPEFramework
             int portId = 0;
             try {
                 portId = stoi(sPortId);
-            }catch (const device::Exception& err) {
-                LOG_DEVICE_EXCEPTION1(sPortId);
-                returnResponse(false);
+            }catch (...) {
+		    LOGWARN("Invalid Arguments");
+		    response["message"] = "Invalid Arguments";
```
npoltorapavlo, line 177: can you please avoid adding undocumented things to the JSON-RPC interface. This is your interface - HdmiInput.json. It does not have "message". According to JSON-RPC protocol https://www.jsonrpc.org/specification#response_object result object does not exist in case of error. I.e. no body will see your "message" anyway

Tests/headers/rdk/ds/dsUtl.h
```diff
@@ -0,0 +1,96 @@
...
+ * This API defines some common utilities for the Device Settings HAL.
+ *
+ * @par Document
+ * Document reference.
+ *
+ * @par Open Issues (in no particular order)
+ * -# None
+ *
+ * @par Assumptions
+ * -# None
```
npoltorapavlo, line 34: Please do not copy pasta the whole DS in here. For example, no rdk Services use dsUTL_*. Please cut it down to only things that are required to build.

npoltorapavlo (changes requested):

HdmiInput/HdmiInput.cpp
anand-ky: You can add "message" to the error object instead of result - https://www.jsonrpc.org/specification#error_object

npoltorapavlo: @anand-ky error object is internally set by Thunder to reflect the error code returned by the API. Example of error codes: https://github.com/rdkcentral/Thunder/blob/82d4791478934292f7260fef233b9743fa34d88e/Source/core/Portability.h#L670 . If people want to document a description of the error they should use schema, like here:  https://github.com/rdkcentral/rdkservices/blob/b83fe9cb6cf6c5535f82031f24865a96e3c4b61a/Messenger/Messenger.json#L74

rajnishdey: Addressed in the latest patch.

Tests/headers/rdk/ds/dsUtl.h
rajnishdey: Unit test cases need the actual rdk headers file, so included original header files.

anand-ky (approved):

Tests/headers/rdk/ds/dsUtl.h
npoltorapavlo: This request was not solved @rajnishdey  @anand-ky . Unit tests should address the rdk services. There’s no DS in that perspective, it’s external code. The purpose of those headers is to make build possible and make external calls mockable . dsUtl.h serves none of that. Same for dsDisplay.h and other declarations .  If you don’t use those declarations don’t add them. Whoever needs them for tests in future may need them in a different way. For example in tests device::List is std::vector, unlike DS. “actual“ DS is no objective and shouldn’t be. Can you fix the request @rajnishdey

npoltorapavlo:
valgrind_log reports issues for both HdcpProfile and HdmiInput. Please fix

anand-ky:
@rajnishdey - Pls address these comments and bring the other changes in 22Q4 branch.

rajnishdey:
> valgrind_log reports issues for both HdcpProfile and HdmiInput. Please fix

Sure, will check valgrind_log and fix it. Then, plan to bring it all together in 22Q4. Thanks.

PR #3249, opened 2022-09-30
Changes to avoid race condition in MM

PaviP98:


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -263,9 +264,26 @@ namespace WPEFramework {
 #else
             internetConnectStatus = isDeviceOnline();
 #endif
+
+            if ( false == internetConnectStatus ) {
+                    MaintenanceManager::_instance->onMaintenanceStatusChange(MAINTENANCE_ERROR);
+                    LOGINFO("Maintenance completed as it is offline mode");
+                    m_abort_flag=false;
```
tdeva14, line 271: This is not needed here

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -263,9 +264,26 @@ namespace WPEFramework {
...
+                    LOGINFO("Worker Thread Completed");
+	    }
```
tdeva14, line 273: There must be a return here. No point in proceeding further

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -290,51 +308,43 @@ namespace WPEFramework {
...
-                        system(cmd.c_str());
-                    }
-                }
-            }
-            /* Here in Solicited, we start with RFC so no
-             * need to wait for any DCM events */
-            else if( SOLICITED_MAINTENANCE == g_maintenance_type){
-                LOGINFO("=============SOLICITED_MAINTENANCE===============");
-                cmd = tasks[0];
+	    /*for( i = 0; i < tasks.size() && !m_abort_flag; i++) {
```
tdeva14, line 311: Any reason for this method of looping through the tasks is not working?

PR #3258, opened 2022-10-04
RDKCMF-8861 RDKShell: add support for LISA to launchApplication()

sverkoye:
* Add LISA support: https://github.com/stagingrdkm/LISA
* As alternative to Packager. First LISA is checked. If not supported it falls back to Packager.


anand-ky:
@binuinbaraj - Pls review

anand-ky (approved):

PR #3268, opened 2022-10-05
RDK-37864, RDK-37865: Software Upgrade Throttling Phase 3

satya200:
Reason for change: Added New Throttling feature
                   Add mm mode change during maintenance time

Test Procedure: 1> Trigger Download.
                2> Image download and flash should happen via c code.

Risks: Low
Priority: P2

Signed-off-by: satya200 <satyasundar_sahu@comcast.com>


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -590,6 +591,20 @@ namespace WPEFramework {
                 maintenanceManagerOnBootup();
             }
         }
+	void MaintenanceManager::sendIARMmode(const char *eventname,int *mode, int size)
```
mvalas827, line 594: You can directly call IARM API inside setMaintenanceMode if this is not being used else where.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -1106,9 +1124,11 @@ namespace WPEFramework {
 
                 std::lock_guard<std::mutex> guard(m_callMutex);
 
+		rdkvfwrfc = checkRdkvFWFlag();
                 /* check if maintenance is on progress or not */
                 /* if in progress restrict the same */
-                if ( MAINTENANCE_STARTED != m_notify_status ){
+		/* This feature is only applicable when firmware download rfc is true */
+                if ( (rdkvfwrfc == true) || (MAINTENANCE_STARTED != m_notify_status) ){
```
mvalas827, line 1131: Line 1135 can be moved before lock_guard as its just validation of input params.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -1133,6 +1153,12 @@ namespace WPEFramework {
                     LOGERR("Maintenance is in Progress, Mode change not allowed");
                     result =true;
                 }
+#if defined(USE_IARMBUS) || defined(USE_IARM_BUS)
```
mvalas827, line 1156: this should go inside if if condition where the mode is getting updated. we need to send event only when it changes. Any caller can make use of getMaintenanceMode() if it doesn't want to persist the mode.

MaintenanceManager/MaintenanceManager.h
```diff
@@ -203,6 +204,7 @@ namespace WPEFramework {
                 uint32_t startMaintenance(const JsonObject& parameters, JsonObject& response);
                 uint32_t stopMaintenance(const JsonObject& parameters, JsonObject& response);
                 uint32_t getMaintenanceMode(const JsonObject& parameters, JsonObject& response);
+		void sendIARMmode(const char *eventname,int *mode, int size);
```
mvalas827, line 207: not needed as mentioned before.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -73,6 +73,7 @@ using namespace std;
 #define PROC_DIR "/proc"
 #define TR181_AUTOREBOOT_ENABLE "Device.DeviceInfo.X_RDKCENTRAL-COM_RFC.Feature.AutoReboot.Enable"
 #define TR181_STOP_MAINTENANCE  "Device.DeviceInfo.X_RDKCENTRAL-COM_RFC.Feature.StopMaintenance.Enable"
+#define TR181_RDKVFWUPGRADER  "Device.DeviceInfo.X_RDKCENTRAL-COM_RFC.Feature.RDKFirmwareUpgrader.Enable"
```
mvalas827, line 76: RFC check not needed in MM

MaintenanceManager/MaintenanceManager.cpp
satya200: It may use in future. That's why i have added.

MaintenanceManager/MaintenanceManager.cpp
satya200: Ok.

MaintenanceManager/MaintenanceManager.cpp
satya200: This RFC check is required because we have firmware download via script method and C method. This rfc is use for deciding either download is via C or script.

satya200: This RFC check is required because we have firmware download via script method and C method. This rfc is use for deciding either download is via C or script.

MaintenanceManager/MaintenanceManager.cpp
satya200: Here i am restricting MaintenanceMode change means if the rfc is true then only "MaintenanceMode" change will work. Because in our discussion you have mention that maintenance mode change is not allowed during maintenance due to firmware download script. Now also we have firmware download via script and C method. And this is taken care based on the rfc value. That's why i have added rfc check.

MaintenanceManager/MaintenanceManager.h
satya200: It may require in future.

Jagadheesan-d (commented):
Please ensure proper comment in the commit message. Otherwise changes looks good.

satya200 (commented):
All review comments addressed.

anand-ky (approved):

PR #3283, opened 2022-10-07
XIONE-10188: Maintenance Reboot not happening

PaviP98:
There is a race condition in Maintenance manager specifically in cases where there are retries to check network connectivity. So did changes to start DCMScript.sh after checking network connectivity and unified the maintenance task execution loop in the main thread.


tdeva14 (commented):
Change looks good

mvalas827 (commented):
Looks good to me..

anand-ky (approved):

PR #3291, opened 2022-10-11
Added reference URL in interface schema

rekhajp98:
Added reference URL key in the interface.schema.json


anand-ky (approved):

anand-ky:
Doc changes. Doesn't need version change.

PR #3292, opened 2022-10-11
Added reference Key in interface.schema.json file

rekhajp98:
Added reference URL key in the interface.schema.json

### DIFF
--- a/Tools/md_generator/generate_md.py
+++ b/Tools/md_generator/generate_md.py
@@ -19,7 +19,7 @@ def convert_json_to_md():
     print()
     filename = os.path.join(dirname, r"../../*/*Plugin.json")
     flist = glob.glob(os.path.join(filename))
-    jsongenpath ="python ../json_generator/generator_json.py"
+    jsongenpath ="python3 ../json_generator/generator_json.py"
 
     for file in flist:
         os.system(r"{} --docs "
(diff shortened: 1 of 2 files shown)


Tools/md_generator/generate_md.py
```diff
@@ -19,7 +19,7 @@ def convert_json_to_md():
     print()
     filename = os.path.join(dirname, r"../../*/*Plugin.json")
     flist = glob.glob(os.path.join(filename))
-    jsongenpath ="python ../json_generator/generator_json.py"
+    jsongenpath ="python3 ../json_generator/generator_json.py"
```
anand-ky, line 22: Checking why we need this change. Won't this break for VMs that use python2? can we separate it out in a different PR?

anand-ky (changes requested):

Tools/md_generator/generate_md.py
RDKDoc: The minimum requirement to use the JSON generator script is Python 3.5 or higher with pip. It is incompatible with python version 2.

rekhajp98: The minimum requirement to use the JSON generator script is Python 3.5 or higher with pip. It is incompatible with python version 2.
As mentioned, I will bring this change in separate PR.

PR #3315, opened 2022-10-12
RDK-38096 - Add Unit Tests for TraceControl

srikanth-vv:
Added unit tests for TraceControl plugin


Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+    {
+        plugin->Deinitialize(&service);
+        PluginHost::IFactories::Assign(nullptr);
+        IarmBus::getInstance().impl = nullptr;
+    }
+};
+
+TEST_F(TraceControlInitializedTest, registeredMethods)
+{
+    SetUp(true);
```
npoltorapavlo, line 108: please move the repeating code to the fixtures

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+    JsonObject params;
+    params.FromString(response);
+    JsonArray arr = params["settings"].Array();
+    JsonObject traceObj;
+    string jsonString;
+    for (unsigned int i = 0; i < arr.Length(); i++)
+    {
+        traceObj = arr[i].Object();
+        traceObj.ToString(jsonString);
+        EXPECT_THAT(jsonString, ::testing::MatchesRegex(_T("\\{"
```
npoltorapavlo, line 127: please make array regex instead

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+    //Set Plugin_TraceControl:Information:enabled
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("set"), _T("{\"module\":\"Plugin_TraceControl\",\"category\":\"Information\",\"state\":\"enabled\"}"), response));
+
+    //Get status Plugin_TraceControl:Information:enabled
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("status"), _T("{\"module\":\"Plugin_TraceControl\",\"category\":\"Information\"}"), response));
+    params.FromString(response);
+    arr = params["settings"].Array();
+    traceObj = arr[0].Object();
+    traceObj.ToString(jsonString);
+    EXPECT_THAT(jsonString, ::testing::MatchesRegex(_T("\\{"
```
npoltorapavlo, line 143: please make array regex instead

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("status"), _T("{\"module\":\"Plugin_TraceControl\"}"), response));
+    params.FromString(response);
+    arr = params["settings"].Array();
+    //Check all categories are set to disabled
+    for (unsigned int i = 0; i < arr.Length(); i++)
+    {
+        traceObj = arr[i].Object();
+        traceObj.ToString(jsonString);
+        EXPECT_THAT(jsonString, ::testing::MatchesRegex(_T("\\{"
```
npoltorapavlo, line 160: please make array regex instead

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+    //Access trace file from the system
+    Core::File file(volatilePath+"tracebuffer");
+    file.Open(true);
+    uint8_t buffer[1024];
+    uint32_t len = file.Read(buffer, sizeof(buffer));
+    std::string s(buffer, buffer+len);
+
+    //Verify the test data is present in the trace file
+    bool result = (s.find("Test1") != string::npos);
```
npoltorapavlo, line 183: please make regex instead

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+    body->ToString(bodyStr);
+
+    JsonObject params;
+    params.FromString(bodyStr);
+    JsonArray arr = params["settings"].Array();
+    string jsonString;
+    for (unsigned int i = 0; i < arr.Length(); i++)
+    {
+        arr[i].Object().ToString(jsonString);
+        EXPECT_THAT(jsonString, ::testing::MatchesRegex(_T("\\{"
```
npoltorapavlo, line 223: please make array regex instead

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+    //Validate the status for all modules. The state should be disabled
+    body = httpResponse->Body<Web::JSONBodyType<Plugin::TraceControl::Data>>();
+    body->ToString(bodyStr);
+    params.FromString(bodyStr);
+    arr = params["settings"].Array();
+    for (unsigned int i = 0; i < arr.Length(); i++)
+    {
+        arr[i].Object().ToString(jsonString);
+        //traceObj.ToString(jsonString);
+        EXPECT_THAT(jsonString, ::testing::MatchesRegex(_T("\\{"
```
npoltorapavlo, line 255: please make array regex instead

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+    body = httpResponse->Body<Web::JSONBodyType<Plugin::TraceControl::Data>>();
+    body->ToString(bodyStr);
+    params.FromString(bodyStr);
+    arr = params["settings"].Array();
+    for (unsigned int i = 0; i < arr.Length(); i++)
+    {
+        arr[i].Object().ToString(jsonString);
+        if(jsonString.find("Plugin_TraceControl") == string::npos)
+        {
+            EXPECT_THAT(jsonString, ::testing::MatchesRegex(_T("\\{"
```
npoltorapavlo, line 289: please make array regex instead

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+    //Validate the status for Plugin_TraceControl module, Information category. The state should be disabled
+    //For the rest of the modules the state should be "disabled"
+    body = httpResponse->Body<Web::JSONBodyType<Plugin::TraceControl::Data>>();
+    body->ToString(bodyStr);
+    params.FromString(bodyStr);
+    arr = params["settings"].Array();
+    for (unsigned int i = 0; i < arr.Length(); i++)
+    {
+        arr[i].Object().ToString(jsonString);
+        if(jsonString.find("Plugin_TraceControl") == string::npos)
```
npoltorapavlo, line 330: please make array regex instead

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+ * limitations under the License.
+ **/
+
+#include "gtest/gtest.h"
+#include <gmock/gmock.h>
+
+#include "TraceControl.h"
+
+#include "FactoriesImplementation.h"
+#include "IarmBusMock.h"
```
npoltorapavlo, line 26: please remove unused things from the test, like iarm mock

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+    //Access trace file from the system
+    Core::File file(volatilePath+"tracebuffer");
+    file.Open(true);
+    uint8_t buffer[1024];
+    uint32_t len = file.Read(buffer, sizeof(buffer));
+    std::string s(buffer, buffer+len);
+
+    //Verify the test data is present in the trace file
+    bool result = (s.find("Test4") != string::npos);
```
npoltorapavlo, line 410: please make regex instead

Tests/tests/test_TraceControl.cpp
srikanth-vv: To reduce the repeated code, I added SetUp() with parameters, so that the tests can pass SetUp(true) & SetUp(false).
Let me know, which other repeated code can be moved?

Tests/tests/test_TraceControl.cpp
srikanth-vv: The problem with using array regex is the array can be of different size based on what modules are enabled/disabled.
Here is the example json output. If one plugin is not available then it may not show in this list and same time in future if a new plugin gets added that test might fail with the exact match check.
I added a generic check that will take care of these cases. The same comment will be applied to other remarks in this review.

{"settings":[{"module":"Plugin_DeviceIdentification","category":"Information","state":"disabled"},{"module":"Plugin_DeviceIdentification","category":"Error","state":"disabled"},{"module":"Plugin_DeviceInfo","category":"Fatal","state":"disabled"},{"module":"Plugin_DeviceInfo","category":"Information","state":"disabled"},{"module":"Plugin_LocationSync","category":"Information","state":"disabled"},{"module":"Plugin_LocationSync","category":"Warning","state":"disabled"},{"module":"Plugin_LocationSync","category":"Error","state":"disabled"},{"module":"Plugin_Messenger","category":"Warning","state":"disabled"},{"module":"Plugin_Messenger","category":"Information","state":"disabled"},{"module":"Plugin_Messenger","category":"Error","state":"disabled"},{"module":"Plugin_SystemServices","category":"Error","state":"disabled"},{"module":"Plugin_SystemServices","category":"Information","state":"disabled"},{"module":"Plugin_TimeSync","category":"Information","state":"disabled"},{"module":"Plugin_TraceControl","category":"Information","state":"disabled"},{"module":"Plugin_TraceControl","category":"Error","state":"disabled"},{"module":"SysLog","category":"ParsingError","state":"enabled"},{"module":"SysLog","category":"Crash","state":"enabled"},{"module":"SysLog","category":"Startup","state":"enabled"},{"module":"SysLog","category":"Shutdown","state":"enabled"},{"module":"SysLog","category":"Notification","state":"enabled"},{"module":"SysLog","category":"Error","state":"enabled"},{"module":"SysLog","category":"Fatal","state":"enabled"}]}

Tests/tests/test_TraceControl.cpp
srikanth-vv: See above response

Tests/tests/test_TraceControl.cpp
srikanth-vv: See above response

Tests/tests/test_TraceControl.cpp
srikanth-vv: See above response

Tests/tests/test_TraceControl.cpp
srikanth-vv: See above response

Tests/tests/test_TraceControl.cpp
srikanth-vv: See above response

Tests/tests/test_TraceControl.cpp
srikanth-vv: See above response

Tests/tests/test_TraceControl.cpp
srikanth-vv: I will clean this up.
Thanks

Tests/tests/test_TraceControl.cpp
srikanth-vv: See above response

Tests/tests/test_TraceControl.cpp
npoltorapavlo: move
```
        ON_CALL(service, Background())
            .WillByDefault(::testing::Return(bBackground));
```
from fixture and add to the test
other things from SetUp(bool bBackground) move to TraceControlInitializedTest()
things from TearDown() move to ~TraceControlInitializedTest()
I hope this answers your question

Tests/tests/test_TraceControl.cpp
npoltorapavlo: it should not be a problem to write a regex
you can check how it's done in other places e.g. https://github.com/rdkcentral/rdkservices/blob/dcaf5c0267ed9e645b70825095252eb13a53dbe8/Tests/tests/test_DeviceInfoWeb.cpp#L103

Tests/tests/test_TraceControl.cpp
srikanth-vv: The trace buffer is a circular buffer allocated in Thunder. The Trace() function writes data into that buffer and in this line [](https://github.com/rdkcentral/rdkservices/blob/sprint/22Q4/TraceControl/TraceOutput.h#L65), the data is retrieved from the buffer and outputs to console (printf). Since I cannot capture this console output, I am reading this trace buffer and matching for the test string.

npoltorapavlo: when unit testing need to find smallest unit and test its interface e.g. input/output. In your case unit is plugin and output is printf/syslog. The buffer file seems to be internal implementation of the unit . If owners rewrite their implementation to not use buffer file it would be complicated to fix your test. Please find a better solution that does not rely on the internal implementation. That could be wrap of syslog, for example

Tests/tests/test_TraceControl.cpp
srikanth-vv: I was able to use the regex from https://github.com/rdkcentral/rdkservices/pull/3345 and update the code.
Some conditional regex are left as it is by parsing the array.

Tests/tests/test_TraceControl.cpp
srikanth-vv: I couldn't figure out on capturing data form syslog or printf to validate the trace output. Tried mock for TraceOutput, but it didn't help to override call to Output() function. I removed this part of the code.

Tests/tests/test_TraceControl.cpp
```diff
@@ -0,0 +1,327 @@
...
+        plugin->Deinitialize(&service);
+        PluginHost::IFactories::Assign(nullptr);
+    }
+};
+
+TEST_F(TraceControlInitializedTest, registeredMethods)
+{
+    ON_CALL(service, Background())
+        .WillByDefault(::testing::Return(true));
+    EXPECT_EQ(string(""), plugin->Initialize(&service));
```
npoltorapavlo, line 95: you don't need TraceControlInitializedTest, ON_CALL, and Initialize to test handler.Exists. please remove

npoltorapavlo (changes requested):

Tests/tests/test_TraceControl.cpp
npoltorapavlo: please remove conditional regex as well. you can split it in 2 regex if you like, 1 to verify response, another to verify specific part of it. if-else is not good, as it doesn't check the specific part is present

Tests/tests/test_TraceControl.cpp
npoltorapavlo: example with system():
https://github.com/rdkcentral/rdkservices/blob/e85f325faae4ab85cc963a2fe8d75292298adbab/Tests/tests/test_SystemServices.cpp#L487

Tests/tests/test_TraceControl.cpp
srikanth-vv: I'll do it

Tests/tests/test_TraceControl.cpp
srikanth-vv: Can you let me know, which specific part the if-else is not checking?

Tests/tests/test_TraceControl.cpp
srikanth-vv: I reviewed it but it didn't work for TraceOutput.

srikanth-vv: Added mock for syslog in the latest commit.

Tests/tests/test_TraceControl.cpp
npoltorapavlo: it is not checking that Plugin_TraceControl is present. It is sad that a test uses if-else like it's random response when it is not

srikanth-vv: Thanks for the finding. I added the check for this module existence. In this particular case I am only interested in "Plugin_TraceControl" module and the rest of the string can be random. I am not expert in regex, so instead of creating complicated regex and test those, I chose to add these conditions.

Tests/mocks/Wraps.cpp
```diff
@@ -9,3 +10,15 @@ extern "C" FILE* __wrap_popen(const char* command, const char* type)
 {
     return Wraps::getInstance().popen(command, type);
 }
+
+//This function will be called for syslog() in the code (added -Wl,-wrap,syslog)
+extern  "C" void __wrap_syslog(int pri, const char *fmt, ...)
```
npoltorapavlo, line 15: this should call real syslog as well

npoltorapavlo (changes requested):

Tests/tests/test_TraceControl.cpp
npoltorapavlo: it needs to be reworked without if-else. if you struggle with that I will have to address it after your change

Tests/mocks/Wraps.cpp
srikanth-vv: What is the need for calling real syslog in this test perspective?

srikanth-vv: Done

npoltorapavlo (changes requested):

Tests/tests/test_TraceControl.cpp
srikanth-vv: I am fine with it.

npoltorapavlo: https://ccp.sys.comcast.net/browse/RDKCOM-3396

anand-ky (approved):

anand-ky:
Regex changes to be addressed in a separate PR.

npoltorapavlo:
please bring to sprint/23Q1
https://github.com/rdkcentral/rdkservices/pull/3387 pending
https://github.com/rdkcentral/rdkservices/pull/3567 pending

srikanth-vv:
> 
pr for merge to 23Q1: [3601](https://github.com/rdkcentral/rdkservices/pull/3601)

PR #3331, opened 2022-10-17
RDK-37628 Adapted RustBridge into rdkservices

Koczek9:
Adapted RustBridge from https://github.com/rdkcentral/ThunderNanoServices/tree/master/RustBridge so it can work as rdkservice. It requires a change like [0001_rust_bridge_interface.txt](https://github.com/rdkcentral/rdkservices/files/9797315/0001_rust_bridge_interface.txt) in the ThunderInterfaces.

Reason for change: There should be possibility of running thunder plugin that are written in rust with a full support of thunder features (RustAdapter is spawning other processes which is not fully thunder-like behaviour)
Test Procedure: Use hello-world plugin to test RustBridge
Risks: Low
Signed-off-by: Marcin Koczwara (marcin.koczwara@consult.red)

### DIFF
--- a/RustBridge/RustBridgeImplementation.cpp
+++ b/RustBridge/RustBridgeImplementation.cpp
@@ -294,8 +294,10 @@ namespace WPEFramework {
 					
 					searchPath.push_back(framework->DataPath());
 					searchPath.push_back(framework->PersistentPath());
-					searchPath.push_back(framework->SystemPath());
-					searchPath.push_back(framework->PluginPath());
+					//searchPath.push_back(framework->SystemPath());
+					//searchPath.push_back(framework->PluginPath());
+					searchPath.push_back(framework->VolatilePath());
+					searchPath.push_back("/usr/lib/wpeframework/plugins/");
 					
 					_callback = callback;
 
--- a/RustBridge/RustBridgeImplementation.cpp
+++ b/RustBridge/RustBridgeImplementation.cpp
@@ -1,0 +1,383 @@
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2022 Metrological
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "RustBridge.h"
+
+namespace WPEFramework {
+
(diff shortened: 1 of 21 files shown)


mhughesacn:
Hi Marcin,
BD is complaining because of the Metrological headers.  If you are using the files from ThunderNanoServices without modification (or very small mods), then please reply (or Slack if you have it) and I'll approve.  If you are making substantial changes, please put the RDK Apache header on top, and keep just two lines of credit to Metrological underneath:

Copyright 2022 Metrological

Thank you,

RustBridge/RustBridgeImplementation.cpp
```diff
@@ -0,0 +1,385 @@
...
+			
+			class Connector {
+			private:
+				using fn_SendTo = void (*)(uint32_t, const char *, Rust::PluginContext *);
+				using fn_Create = Rust::Plugin *(*)(const char *name, fn_SendTo send_to, Rust::PluginContext *, void *);
+				using fn_Destroy = void (*)(Rust::Plugin *p);
+				using fn_Init = void (*)(Rust::Plugin *p, const char *json);
+				using fn_Invoke = void (*)(Rust::Plugin *p, const char *json_req, Rust::RequestContext req_ctx);
+
+				// TODO: Discuss this since I think these are not needed. I assume it is to send events using the
```
pwielders, line 71: Taking this Metrological plugin into the RDKServices as a replacement for the RustAdapter is oke, but when we did the implementation we would like to propose some fundamental changes to improve the performance. One of them is to define a new method that reports to the bridge the exposed methods. This would handle methods that have no meaning/counter part in rust already in this plugin (avoiding needless process hops).
Might be good, as we make this now, I assume, production worthy to pick up these discussions again.

RustBridge/RustBridgeImplementation.cpp
```diff
@@ -0,0 +1,385 @@
...
+				uint32_t Initialize(const std::vector<string>& thunderPaths, const string& locator, const string& className, const string& configLine) {
+					
+					uint32_t result = Load(thunderPaths, locator);
+					
+					if (result == Core::ERROR_NONE) {
+						ASSERT(_library.IsLoaded() == true);
+						ASSERT(_fn_create != nullptr);
+						ASSERT(_fn_init != nullptr);
+						
+						// TODO: Discuss why we want to pass back a function durint the create that is already part of the
```
pwielders, line 107: More open issues to discuss..

pwielders (commented):
The RUST examples are two fold and need to be aligned. The HelloWorld is as delivered by Comcast and the RustBridge can work with it. However from a metrological perspective we would like to offer a  different approch (move some logic to the rust shim library (lib.rs).
The adavantage of moving it here is that the Rust plugin developper does not have to be aware how the JSONRPC enveloppe looks like and only need to focus on the plugin functionality and as a second advantage, if, in the future performance could be an issue, there will be possibilities to improve the performance (direct method calls, without seriaization/deserialization) without touching any of the written RUST modules.
Next to these advantages, the methods (invokable items) available in RUST are automatically registered. So please decide which example should be leading and adapt the other one.
HelloWorld is the Comcast proposal, Calculator is the Metrological proposal.

goruklu:
cherry-picked to sprint/23Q1 branch with #3614

PR #3333, opened 2022-10-17
RDK-38159: TimeZone update

varumugam123:
This change is to enable TZ update based on an external file (probably updated by some external component)


WebKitBrowser/InjectedBundle/main.cpp
```diff
@@ -125,9 +146,54 @@ static class PluginHost {
...
+            return;
+        }
+
+        GFile *file = g_file_new_for_path(kTimeZoneFile);
+        if(g_file_query_exists(file, nullptr)) {
+            HandleTimeZoneFileUpdate(nullptr, file, nullptr, G_FILE_MONITOR_EVENT_CHANGES_DONE_HINT, nullptr);
+        } else {
+            _timeZoneFileMonitor = g_file_monitor_file(file, G_FILE_MONITOR_NONE, nullptr, nullptr);
+            _timeZoneFileMonitorId = g_signal_connect(_timeZoneFileMonitor, "changed", reinterpret_cast<GCallback>(HandleTimeZoneFileUpdate), &_timeZoneFileMonitorId);
+            SYSLOG(Trace::Information, (_T("Installed file monitor for %s"), kTimeZoneFile));
```
emutavchi, line 187: Timezone might be updated after the browser startup. So, I think it makes sense to keep the monitor running

emutavchi (changes requested):

WebKitBrowser/InjectedBundle/main.cpp
varumugam123: Updated implementation.

WebKitBrowser/InjectedBundle/TimeZoneSupport.cpp
```diff
@@ -0,0 +1,94 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+#include <syslog.h>
+
+#include "TimeZoneSupport.h"
+#include "../../helpers/UtilsString.h"
```
emutavchi, line 23: WebKitBrowserPlugin should be built without dependency on any other plugins/services/helpers.

emutavchi (changes requested):

WebKitBrowser/InjectedBundle/TimeZoneSupport.cpp
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        _timeZoneFileMonitor = g_file_monitor_file(file, G_FILE_MONITOR_NONE, nullptr, nullptr);
+        _timeZoneFileMonitorId = g_signal_connect(_timeZoneFileMonitor, "changed", reinterpret_cast<GCallback>(HandleTimeZoneFileUpdate), this);
+        SYSLOG(Trace::Information, (_T("Installed file monitor for \"%s\""), _tzFile.c_str()));
+    }
+
+    void TimeZoneSupport::Deinitialize()
+    {
+        if (_timeZoneFileMonitor && _timeZoneFileMonitorId > 0)
```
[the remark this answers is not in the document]
varumugam123: Done

emutavchi (commented):
Looks good to me. The same is needed for GLib extension. Could you please update?

emutavchi (approved):

emutavchi:
@modeveci  could you please review this?

MFransen69 (commented):
Hi,

As this feature is only included with a build flag we have no problem with it in principle, I'm not sure Webkit will pick up the changes later on but I guess that was tested to work? But just to make sure and to prevent you from doing double work perhaps, you are bypassing the TimeZone support from the LocationSync service on purpose? 
There was some time ago a change request by Comcast to remove the timezone file from the SystemServices and bring it into the LocationSync and add a TimeZone interface to it. After some discussion we agreed somebody from our team would implement this to the LocationSync. 
As this was dependent on some changes in Thunder not yet available on R2 a pull request was created for that first which was only very recently merged. So soon (it is already implemented but only needs to be moved to RDKServices) we will make sure LocationSync will have the possibility to override the TimeZone.
If it is still the plan by Comcast to remove the timezone file and have LocationSync as the single source of (timezone) truth I would not know (although from a system architecture point of view I would say it would be preferable to have one source/mechanism to work with TimeZones). 
You could I guess ask Anand (Kandasamy) he was involved in the discussions. If indeed that would be the case you could in the InjectableBundle get the set TimeZone via the Thunder Substems interface or as an alternative via the TimeZone interface to the LocationSync.

emutavchi:
@MFransen69 this is needed for devices where LocatoinSync is not used to set up TimeZone. if it is fine with you, I'd like to proceed with the merge.

MFransen69:
Hi @emutavchi 

No problem, indeed already mentioned I see no direct issue as it is conditional.
I approved the changes.

MFransen69 (approved):

PR #3343, opened 2022-10-19
[RDKDEV-363]: DisplayInfo: syncup changes from metro version

HaseenaSainul:
Nexus removed from the repo and added CMake to fetch from the Nexus repo
RPI update to use ExtendedDisplayIdentification class
Linux libDRM based generic implementation added
DisplayInfo implementation cleanup
DisplayInfo deinit implementation improvement


mhughesacn:
Hi Haseena: The code match to udev is a bit awkward.  I think it is reasonable to consider it as boilerplate i.e. if you are using udev from outside, you need to have the same structure to understand the netlink header.  The code matches are all to LGPL code, so I would ask that you change the code slightly so as not to simply copy the original.  This is just to get rid of exact comment matches - I think it is reasonable to keep the structure element names.

/**
 * See https://github.com/systemd/systemd/blob/master/src/libsystemd/sd-device/device-monitor.c
 * for details of the udev_monitor_netlink_header structure, which is needed
 * in order to properly parse messages coming in through the socket connection
 * (UdevObserver class).
 */
static constexpr auto UDEV_MONITOR_MAGIC = 0xfeedcafe;
struct udev_monitor_netlink_header {
    char prefix[8];
    unsigned int magic;
    unsigned int header_size;
    unsigned int properties_off;
    unsigned int properties_len;
    unsigned int filter_subsystem_hash;
    unsigned int filter_devtype_hash;
    unsigned int filter_tag_bloom_hi;
    unsigned int filter_tag_bloom_lo;
};

HaseenaSainul:
> Hi Haseena: The code match to udev is a bit awkward. I think it is reasonable to consider it as boilerplate i.e. if you are using udev from outside, you need to have the same structure to understand the netlink header. The code matches are all to LGPL code, so I would ask that you change the code slightly so as not to simply copy the original. This is just to get rid of exact comment matches - I think it is reasonable to keep the structure element names.
> 
> ```
> /**
>  * See https://github.com/systemd/systemd/blob/master/src/libsystemd/sd-device/device-monitor.c
>  * for details of the udev_monitor_netlink_header structure, which is needed
>  * in order to properly parse messages coming in through the socket connection
>  * (UdevObserver class).
>  */
> static constexpr auto UDEV_MONITOR_MAGIC = 0xfeedcafe;
> struct udev_monitor_netlink_header {
>     char prefix[8];
>     unsigned int magic;
>     unsigned int header_size;
>     unsigned int properties_off;
>     unsigned int properties_len;
>     unsigned int filter_subsystem_hash;
>     unsigned int filter_devtype_hash;
>     unsigned int filter_tag_bloom_hi;
>     unsigned int filter_tag_bloom_lo;
> };
> ```

Updated as you suggested

pwielders (dismissed):

DisplayInfo/Linux/PlatformImplementation.cpp
```diff
@@ -0,0 +1,745 @@
...
+                    }
+                }
+            }
+
+            return receivedSize;
+        }
+
+    private:
+        /** 
+         * The following are structures copied from udev in order to properly parse messages coming in through
```
mhughesacn, line 136: Hi Haseena,
Please use the comment that I provided as well.

HaseenaSainul: done :)

mhughesacn: Thank you!

pradeeptakdas:
Internal Jira for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3380

anand-ky (approved):

PR #3354, opened 2022-10-25
RDk-38088: Minor version update on the deprecated APIs in System plugin

pkumbh631:


anand-ky (approved):
These are changes in semantic versioning & changelog for System API cleanup changes that came in US RDK-35953 which is already released.

PR #3376, opened 2022-10-31
RDKTV-20401: TV did not enter to Deepsleep , and stayed ~14 hrs in th…

PaviP98:
…e standby state.

Maintenance window is open for long time and it was not completed successfully and hence TV didn't went to deepsleep.

Whenever stopMaintenance Api executed, it will terminate the task which is currently running and will send that event status to MM, also it will update the status of remaining tasks as well. Since MM didn't trigger those scripts it will simply ignore those events and so maintenance was not completed.


tdeva14 (commented):
Changes look fine to me

PR #3380, opened 2022-11-03
RDK:38786:Adding deprecated api for HdmiInput, CompositeInput

rekhajp98:


HdmiInput/HdmiInput.json
```diff
@@ -56,6 +56,8 @@
     },
     "methods": {
         "getHDMIInputDevices": {
+            "deprecated" : true,
```
anand-ky, line 59: Apart from adding it for individual APIs, can we also add it at the top under Description section. We should add a link to the new plugin (AVInput).

anand-ky: Also add deprecated label for all the events.

rekhapabba: Added deprecation part in Description section. Added deprecated labels for all events

anand-ky (approved):

PR #3402, opened 2022-11-09
RDK-38400: Support Checkpoint and Restore in RDKShell

adrianM27:


RDKShell/RDKShell.cpp
```diff
@@ -5722,6 +5729,67 @@ namespace WPEFramework {
             returnResponse(status);
         }
 
+        uint32_t RDKShell::checkpointWrapper(const JsonObject& parameters, JsonObject& response)
+        {
+            LOGINFOMETHOD();
+            bool status = false;
+            if (parameters.HasLabel("callsign"))
```
adrianM27, line 5736: Shouldn;t we check if its already checkpointed?

RDKShell/RDKShell.cpp
```diff
@@ -5722,6 +5729,67 @@ namespace WPEFramework {
...
+            returnResponse(status);
+        }
+
+        uint32_t RDKShell::restoreWrapper(const JsonObject& parameters, JsonObject& response)
+        {
+            LOGINFOMETHOD();
+            bool status = false;
+            if (parameters.HasLabel("callsign"))
+            {
+                std::string callsign = parameters["callsign"].String();
```
adrianM27, line 5785: Shouldn't we check if it is checkpointed based on ie internal mMemCheckpointRestoreClient list with checkpointed pids/callSigns ?

RDKShell/RDKShell.cpp
```diff
@@ -7162,6 +7230,200 @@ namespace WPEFramework {
...
+                });
+            requestsThread.detach();
+        }
+
+        bool RDKShell::MemCheckpointRestoreClient::sendRcvCmd(
+            RDKShell::MemCheckpointRestoreClient::ServerRequest &cmd,
+            RDKShell::MemCheckpointRestoreClient::ServerResponse &resp,
+            uint32_t timeouteMs)
+        {
+            const char *serverSocketName = "/tmp/memcrservice";
```
adrianM27, line 7306: Not sure if we should make this socket path configurable?

goruklu: would probably be better if you use a #define  for the socket name.

RDKShell/RDKShell.cpp
```diff
@@ -7162,6 +7243,281 @@ namespace WPEFramework {
...
+            mProcessedAppsLock.lock();
+            if(mProcessedApps.find(callSign) == mProcessedApps.end())
+            {
+                //App not processed
+                isCheckpointAllowed = true;
+            }
+            else if(mProcessedApps[callSign] == RDKShell::MemCheckpointRestoreClient::RESTORING)
+            {
+                //App being processeed, then shedule checkpoint only if restore is ongoing
+                isCheckpointAllowed = true;
```
goruklu, line 7260: I don't understand how we can proceed with checkpoint() operation when the app is in "RESTORING" state ?

RDKShell/RDKShell.cpp
```diff
@@ -7162,6 +7243,281 @@ namespace WPEFramework {
 
             return result;
         }
+
+        bool RDKShell::MemCheckpointRestoreClient::checkpoint(const std::string &callSign, uint32_t timeouteMs)
```
goruklu, line 7247: I'm wondering whether we should  move all the RDKShell::MemCheckpointRestoreClient() code to a separate file other than rdkshell.cpp. That file is becoming huge...

RDKShell/RDKShell.cpp
```diff
@@ -7162,6 +7243,281 @@ namespace WPEFramework {
...
+        }
+
+        bool RDKShell::MemCheckpointRestoreClient::restore(const std::string &callSign, uint32_t timeouteMs)
+        {
+            bool isRestoreAllowed = false;
+
+            mProcessedAppsLock.lock();
+            if(mProcessedApps.find(callSign) != mProcessedApps.end()
+                && (mProcessedApps[callSign] == RDKShell::MemCheckpointRestoreClient::CHECKPOINTING
+                        || mProcessedApps[callSign] == RDKShell::MemCheckpointRestoreClient::CHECKPOINTED))
```
goruklu, line 7288: where do we set `mProcessedApps[callSign] = RDKShell::MemCheckpointRestoreClient::CHECKPOINTED`
or `mProcessedApps[callSign] = RDKShell::MemCheckpointRestoreClient::RESTORED` ? I could not find anywhere in the code ?

adrianM27: yest, it set at the end of request thread lambda: https://github.com/rdkcentral/rdkservices/pull/3402/files#diff-57606d52dd6bdec85c0397fe933c537b2288904cebdc9ce65e8d8671046b875aR7368-R7387 , the transition is as follow:
CHECKPOINTING->CHECKPOINTED
RESTORING->remove from mProcessesApps

RDKShell/RDKShell.cpp
adrianM27: Memcr is handling requests one by one. From app cycle perspective it may make some sense to trigger Checkpoint if Restore is still ongoing, since after Restore the App should be still suspended

RDKShell/RDKShell.cpp
adrianM27: yes agree, just wonder where MemCheckpointRestore.cpp  should be added, alognside RDKShell.cpp or placed into a separate subfolder ie 3rdParties, or ExtClients

RDKShell/RDKShell.cpp
adrianM27: Now I am thinking if above won't be overcomplicated for ie RA logic. And for simplicity, it is better to remove this Checkpoint condition

RDKShell/RDKShell.cpp
```diff
@@ -4590,6 +4615,7 @@ namespace WPEFramework {
             string method = "status";
             Core::JSON::ArrayType<PluginHost::MetaData::Service> joResult;
             auto thunderController = getThunderControllerClient();
+            //FIXME: if something is checkpointed, API hangs here
```
adrianM27, line 4618: Investigation is ongoing why it is hanging. It seems that inner implementations of https://github.com/rdkcentral/Thunder/blob/master/Source/WPEFramework/ControllerJsonRpc.cpp#L357 tries to talk with checkpointed app, but still not clear at what point

adrianM27:
rebased pull request for sprint/23Q1: https://github.com/rdkcentral/rdkservices/pull/3647

PR #3406, opened 2022-11-10
RDK-38482: Gtests for Frontpanel Plugin

hgfell683:
RDK-38482: Gtests for Frontpanel Plugin

Reason for change:
Creating the unit tests for the Frontpanel plugin on rdkservices 
Test Procedure: None
Risks: Low

Signed-off-by:Hayden Gfeller <Hayden_Gfeller@comcast.com>


Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+#include "FactoriesImplementation.h"
+
+#include "FrontPanelMock.h"
+#include "FrontPanelTextDisplayMock.h"
+#include "FrontPanelConfigMock.h"
+#include "IarmBusMock.h"
+#include "ServiceMock.h"
+
+#include "dsMgr.h"
```
npoltorapavlo, line 13: dsMgr.h is empty file

npoltorapavlo (changes requested):

npoltorapavlo:
please fix valgrind reportings related to FrontPanel

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,514 @@
...
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("setBlink")));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("set24HourClock")));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("is24HourClock")));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("setClockTestPattern")));
+}
+
+TEST_F(FrontPanelDsTest, setBrightnessWIndex)
+{
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setBrightness"), _T("{\"brightness\": 1,\"index\": \"power_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
```
npoltorapavlo, line 193: a test case should actually test something. how do you verify the business logic of "setBrightness"? i.e. how do you verify it works correctly instead of returning `"{\"success\":true}"`. by looking at this test case I assume "setBrightness" may just return `"{\"success\":true}"` and that will be OK for your test suite? if you are verifying that "setBrightness" triggers the backend then please add expectations for that. this concerns other test cases as well

npoltorapavlo (changes requested):

Tests/tests/test_FrontPanel.cpp
hgfell683: In this case, with setBrightness(and most of the other functions that look like this), the function, depending on the inputs, calls something akin to this(in either FrontPanel.cpp or frontpanel.cpp):

`try
                            {
                                device::FrontPanelIndicator::getInstance(fp_ind.c_str()).setBrightness(int(brightness));
                                ok = true;
                            }`
   where it calls our mocked function. The set functions never expect any returned input, since our mock functions don't 'fail' so to speak, the only things to test are the different inputs(whether it has index, and what value index contains changes what parts of the function we go in to) that are sent to the function, so we trigger the different parts of the code, and make sure, assuming the mock is successful, the function still runs through everything appropriately. As the set functions are all pretty much type void, they can't return anything, so I'm not sure what you want me to look for in this case. 
   
   Looking at other test classes, test_HdmiInput.cpp, test_FrameRate.cpp, and a few others have functions that just pass in inputs, and expect no errors and `"{\"success\":true}"`. So I'm not sure what else there is to verify as the functions in question never do anything with the output from setBrightness, and simply call it--and if it succeeds, sets it so that the function returns true.

npoltorapavlo: OK this test case verifies JSON RPC interface. got it

on set-top one can check led brightness to see that "setBrightness" works. if `device::FrontPanelIndicator` switches led in real life, you can add another test case with `ON_CALL/EXPECT_CALL` to verify that "setBrightness" does some job on that matter as well

hgfell683: Yes, the JSON RPC interface. Do you have an example of what you want for the other test case, but I'm still not sure what I should be verifying here.

Also, in regards to the Valgrind reports, I'm assuming you mean the loss record where some of the bytes are referred to as 'still reachable'? I remember looking into them before, but I couldn't figure out what was causing them to remain still reachable. Backtracing, they all seem to be objects created from things defined in frontpanel.h, and not something I'm directly calling/creating on my end.

npoltorapavlo: @hgfell683 

setBrightnessWIndex
setBrightnessClock
setBrightness
setBrightnessFalse
setBrightnessNeg1

that you added are verifying that "setBrightness" checks input parameters

checking input parameters is not the only thing "setBrightness" does

it also should have some code that sets the brightness

please add a test for the code which sets brightness. so that we know that "setBrightness" sets brightness. this is what this request is for

'still reachable' may not be leaked but can indicate something wrong in the code. should be investigated, what causes that. this is what the second request is for

hgfell683: @npoltorapavlo the setBrightness test functions still tests set the brightness. The value that always needs to be sent, regardless, is brightness (in most cases, it's brightness = 1, when it isn't set it errors out, and there is a case to test specifically for that), and that value is what is being used to set the brightness, and is passed into the setBrightness function that we mocked in devicesettings.h. What the index is really telling the code, by which section it goes into, is what light we're talking about. Whether we're setting the brightness for power led, the clock led, or another led that is present on the device. But regardless of which section it goes int based on the index, we always set trigger the setBrightness mock function. If it's not clear, I can rewrite the the names so that it's a little more clear on what this does. And if I'm still misunderstanding what you're asking, I apologize in advance.

And I'll take another look and see if I can figure out what is causing the memory issues. Fresh eyes from a relaxing weekend might help me see what I couldn't see last week.

npoltorapavlo: something like this would work:
```
TEST_F(FrontPanelDsTest, setBrightnessTriggersBrightnessViaDs)
{
    EXPECT_CALL(frontPanelIndicator...Mock, setBrightness(...))
        .Times(1)
        .WillOnce(::testing::Return());

    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setBrightness"), ...
    EXPECT_EQ(response, string("{\"success\":true}"));
}
```

npoltorapavlo (changes requested):

FrontPanel/FrontPanel.cpp
```diff
@@ -635,7 +644,7 @@ namespace WPEFramework
 #ifdef CLOCK_BRIGHTNESS_ENABLED
             try
             {
-                indicatorInfo = getFrontPanelIndicatorInfo(device::FrontPanelConfig::getInstance().getTextDisplay(0));
+		indicatorInfo = getFrontPanelIndicatorInfo(device::FrontPanelConfig::getInstance().getTextDisplay(0));
```
npoltorapavlo, line 647: please remove whitespace changes

helpers/frontpanel.cpp
```diff
@@ -300,7 +300,7 @@ namespace WPEFramework
         {
             try
             {
-                clockBrightness =  device::FrontPanelTextDisplay::getInstance("Text").getTextBrightness();
+                clockBrightness = device::FrontPanelTextDisplay::getInstance("Text").getTextBrightness();
```
npoltorapavlo, line 303: please remove whitespace changes

helpers/frontpanel.cpp
```diff
@@ -632,7 +631,7 @@ namespace WPEFramework
 
         void CFrontPanel::startBlinkTimer(int numberOfBlinkRepeats)
         {
-            LOGWARN("startBlinkTimer numberOfBlinkRepeats: %d m_blinkList.length : %zu", numberOfBlinkRepeats, m_blinkList.size());
+            LOGWARN("startBlinkTimer numberOfBlinkRepeats: %d m_blinkList.length : %d", numberOfBlinkRepeats,(int) m_blinkList.size());
```
npoltorapavlo, line 634: please revert

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+                    return IARM_RESULT_SUCCESS;
+                }));
+	FrontPanelIndicatorImplMock frontPanelIndicatorMock;
+
+    device::FrontPanelIndicator indicatorList;
+
+        indicatorList.impl = &frontPanelIndicatorMock;
+    ON_CALL(frontPanelConfigImplMock, getIndicators())
+        .WillByDefault(::testing::Return(device::List<device::FrontPanelIndicator>({indicatorList})));
+    ON_CALL(frontPanelIndicatorImplMock, getName())
```
npoltorapavlo, line 100: please reformat as asked in [readme](https://github.com/rdkcentral/rdkservices/tree/sprint/22Q4/Tests#faq)

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+{
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setBrightness"), _T("{\"brightness\": 1,\"index\": \"clock_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+}
+
+TEST_F(FrontPanelDsTest, setBrightness)
+{
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setBrightness"), _T("{\"brightness\": 1}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+}
```
npoltorapavlo, line 207: I don't think such tests make sense

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+TEST_F(FrontPanelDsTest, setClockBrightness)
+{
+    ON_CALL(frontPanelTextDisplayImplStringMock, setTextBrightness(::testing::_))
+        .WillByDefault(::testing::Return());
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setClockBrightness"), _T("{\"brightness\": 1}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+}
```
npoltorapavlo, line 269: I don't think this test makes sense

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+    ON_CALL(frontPanelConfigImplMock, getIndicators())
+        .WillByDefault(::testing::Return(device::List<device::FrontPanelIndicator>({ indicatorList })));
+
+    ON_CALL(frontPanelIndicatorMock, getBrightnessLevels(::testing::_,::testing::_,::testing::_))
+        .WillByDefault(::testing::Invoke(
+            [&](int &levels,int &min,int &max) {
+                levels=1;
+                min=0;
+                max=2;
+            }));
```
npoltorapavlo, line 320: please remove duplicate ON_CALL

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+
+    ON_CALL(frontPanelIndicatorMock, getName())
+        .WillByDefault(::testing::Return("Power"));
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("getFrontPanelLights"), _T(""), response));
+    EXPECT_EQ(response, string("{\"supportedLights\":[\"power_led\",\"clock_led\"],\"supportedLightsInfo\":{\"power_led\":{\"range\":\"boolean\",\"min\":0,\"max\":2,\"colors\":[\"white\"],\"colorMode\":0},\"clock_led\":{\"range\":\"boolean\",\"min\":0,\"max\":2,\"colors\":[\"white\"],\"colorMode\":0}},\"success\":true}"));
+    indicatorList.impl = nullptr;
+    displayList.impl = nullptr;
+    displayList2.FrontPanelIndicator::impl = nullptr;
```
npoltorapavlo, line 345: no need to set impl = nullptr

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+}
+
+TEST_F(FrontPanelDsTest, getPreferences)
+{
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("getPreferences"), _T(""), response));
+    EXPECT_EQ(response, string("{\"preferences\":{},\"success\":true}"));
+}
+
```
npoltorapavlo, line 353: I don't think such test makes sense

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+{
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("powerLedOff"), _T("{\"index\": \"data_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+}
+TEST_F(FrontPanelDsTest, powerLedOffRecord)
+{
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("powerLedOff"), _T("{\"index\": \"record_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+}
+
```
npoltorapavlo, line 396: I don't think such tests make sense

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+    IARM_Bus_PWRMgr_EventData_t eventData;
+    eventData.data.state.newState =IARM_BUS_PWRMGR_POWERSTATE_ON;
+    eventData.data.state.curState =IARM_BUS_PWRMGR_POWERSTATE_STANDBY;
+
+    handler.Subscribe(0, _T("powerModeChange"), _T("client.events.powerModeChange"), message);
+
+    dsFrontPanelModeChange(IARM_BUS_PWRMGR_NAME, IARM_BUS_PWRMGR_EVENT_MODECHANGED, &eventData , 0);
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("powerLedOn"), _T("{\"index\": \"power_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
```
npoltorapavlo, line 420: I don't think such test makes sense

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+    ON_CALL(frontPanelConfigImplMock, getTextDisplay(test))
+        .WillByDefault(::testing::ReturnRef(displayList));
+    ON_CALL(frontPanelTextDisplayImplStringMock, getCurrentTimeFormat())
+        .WillByDefault(::testing::Return(dsFPD_TIME_24_HOUR));
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("set24HourClock"), _T("{\"is24Hour\": true}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+    displayList.impl = nullptr;
+    displayList.FrontPanelIndicator::impl = nullptr;
+
```
npoltorapavlo, line 466: I don't think such tests make sense

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,523 @@
...
+    ON_CALL(frontPanelTextDisplayImplStringMock, setColor())
+        .WillByDefault(::testing::Return());
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setLED"), _T("{\"ledIndicator\": \"power_led\", \"brightness\": 50, \"color\": \"red\", \"red\": 0, \"green\": 0, \"blue\":0}"), response));
+
+        EXPECT_EQ(response, string("{\"success\":true}"));
+}
+TEST_F(FrontPanelDsTest, setPreferences)
+{
+
+        EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setPreferences"), _T("{\"preferences\":{}}"), response));
```
npoltorapavlo, line 519: I don't think such tests make sense

npoltorapavlo (changes requested):

Tests/tests/test_FrontPanel.cpp
npoltorapavlo: please add expectation for setTimeFormat. getCurrentTimeFormat may not be called - it's used for logging which is not important

hgfell683: @npoltorapavlo Pushed an updated for everything that was pointed out, please review when you have a chance.

CMakeLists.txt
```diff
@@ -14,7 +14,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
npoltorapavlo, line 17: please remove whitespace changes

FrontPanel/FrontPanel.cpp
```diff
@@ -88,13 +88,15 @@ namespace
 
     string svc2iarm(const string &name)
     {
-        const char *s = name.c_str();
+	const char *s = name.c_str();
```
npoltorapavlo, line 91: please remove whitespace changes in this file

Tests/CMakeLists.txt
```diff
@@ -87,9 +88,11 @@ link_directories(../LocationSync
         ../DeviceInfo
         ../HdmiInput
         ../HdcpProfile
-        ../TraceControl
+	../TraceControl
         ../Network
 	../WifiManager
+        ../FrontPanel
+
```
npoltorapavlo, line 95: please remove whitespace changes

Tests/mocks/devicesettings.h
```diff
@@ -980,111 +995,219 @@ class Host {
...
-        return impl->getIndicators();
-    }
-    FrontPanelTextDisplay& getTextDisplay(const std::string &name)
-    {
-        static FrontPanelTextDisplay instance;
-        return instance;
-    }
-};
-
+namespace device{
```
npoltorapavlo, line 999: please remove whitespace changes

helpers/frontpanel.cpp
```diff
@@ -491,8 +491,7 @@ namespace WPEFramework
             bool success = false;
             string ledIndicator = svc2iarm(parameters["ledIndicator"].String());
             int brightness = -1;
-
-            if (parameters.HasLabel("brightness"))
+	    if (parameters.HasLabel("brightness"))
```
npoltorapavlo, line 494: please remove whitespace changes

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+
+#include "FactoriesImplementation.h"
+
+#include "FrontPanelMock.h"
+#include "FrontPanelTextDisplayMock.h"
+#include "FrontPanelConfigMock.h"
+#include "IarmBusMock.h"
+#include "ServiceMock.h"
+
+#include "dsMgr.h"
```
npoltorapavlo, line 13: dsMgr.h is empty file

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("setBlink")));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("set24HourClock")));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("is24HourClock")));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("setClockTestPattern")));
+}
+
+TEST_F(FrontPanelDsTest, setBrightnessWIndex)
+{
+    ON_CALL(frontPanelIndicatorImplStringMock, setBrightness(::testing::_, ::testing::_))
+        .WillByDefault(::testing::Return());
```
npoltorapavlo, line 185: please check values in setBrightness

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setBrightness"), _T("{\"brightness\": 1,\"index\": \"power_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+}
+TEST_F(FrontPanelDsTest, setBrightnessClock)
+{
+    ON_CALL(frontPanelTextDisplayImplStringMock, setTextBrightness(::testing::_))
+        .WillByDefault(::testing::Return());
```
npoltorapavlo, line 192: please check value in setTextBrightness

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setBrightness"), _T("{\"brightness\": 1,\"index\": \"clock_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+}
+
+TEST_F(FrontPanelDsTest, setBrightness)
+{
+    ON_CALL(frontPanelIndicatorImplStringMock, setBrightness(::testing::_, ::testing::_))
+        .WillByDefault(::testing::Return());
```
npoltorapavlo, line 200: please check values in setBrightness

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+    ON_CALL(frontPanelTextDisplayImplStringMock, getTextBrightness())
+        .WillByDefault(::testing::Return(50));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("getClockBrightness"), _T(""), response));
+    EXPECT_EQ(response, string("{\"brightness\":50,\"success\":true}"));
+}
+
+TEST_F(FrontPanelDsTest, setClockBrightness)
+{
+    ON_CALL(frontPanelTextDisplayImplStringMock, setTextBrightness(::testing::_))
+        .WillByDefault(::testing::Return());
```
npoltorapavlo, line 261: please check value in setTextBrightness

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("is24HourClock"), _T(""), response));
+    EXPECT_EQ(response, string("{\"is24Hour\":true,\"success\":true}"));
+
+}
+
+TEST_F(FrontPanelDsTest, powerLedOffPower)
+{
+    ON_CALL(frontPanelTextDisplayImplStringMock, setTextBrightness(::testing::_))
+        .WillByDefault(::testing::Return());
```
npoltorapavlo, line 351: please check value in setTextBrightness

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("powerLedOff"), _T("{\"index\": \"power_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+}
+TEST_F(FrontPanelDsTest, powerLedOffData)
+{
+    ON_CALL(frontPanelIndicatorImplStringMock, setState(::testing::_))
+        .WillByDefault(::testing::Return());
```
npoltorapavlo, line 359: please check value in setState

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("powerLedOff"), _T("{\"index\": \"data_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+}
+TEST_F(FrontPanelDsTest, powerLedOffRecord)
+{
+    ON_CALL(frontPanelIndicatorImplStringMock, setState(::testing::_))
+        .WillByDefault(::testing::Return());
```
npoltorapavlo, line 366: please check value in setState

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+    IARM_Bus_PWRMgr_EventData_t eventData;
+    eventData.data.state.newState =IARM_BUS_PWRMGR_POWERSTATE_ON;
+    eventData.data.state.curState =IARM_BUS_PWRMGR_POWERSTATE_STANDBY;
+
+    handler.Subscribe(0, _T("powerModeChange"), _T("client.events.powerModeChange"), message);
+
+    dsFrontPanelModeChange(IARM_BUS_PWRMGR_NAME, IARM_BUS_PWRMGR_EVENT_MODECHANGED, &eventData , 0);
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("powerLedOn"), _T("{\"index\": \"power_led\"}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
```
npoltorapavlo, line 394: I don't think such test makes sense

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+    displayList.FrontPanelIndicator::impl = &frontPanelIndicatorImplMock;
+
+    std::string test = "Text";
+
+    ON_CALL(frontPanelConfigImplMock, getTextDisplay(test))
+        .WillByDefault(::testing::ReturnRef(displayList));
+    ON_CALL(frontPanelTextDisplayImplStringMock, getCurrentTimeFormat())
+        .WillByDefault(::testing::Return(dsFPD_TIME_24_HOUR));
+    ON_CALL(frontPanelTextDisplayImplStringMock, setTimeFormat(::testing::_))
+        .WillByDefault(::testing::Return());
```
npoltorapavlo, line 435: please check value in setTimeFormat

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+
+    ON_CALL(frontPanelIndicatorImplStringMock, getBrightness())
+        .WillByDefault(::testing::Return(50));
+    ON_CALL(frontPanelTextDisplayImplStringMock, getTextBrightness())
+        .WillByDefault(::testing::Return(50));
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setBlink"), _T("{\"blinkInfo\": {\"ledIndicator\": \"power_led\", \"iterations\": 10, \"pattern\": [{\"brightness\": 50, \"duration\": 100, \"color\": \"red\", \"red\": 0, \"green\":0, \"blue\":0}]}}"), response));
+   
+       	EXPECT_EQ(response, string("{\"success\":true}"));
+}
```
npoltorapavlo, line 453: I don't think such test makes sense

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+        .WillByDefault(::testing::Return(100));
+    
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setClockTestPattern"), _T("{\"show\": true, \"timeInterval\": 4}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+    
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("setClockTestPattern"), _T("{\"show\": false, \"timeInterval\": 4}"), response));
+    EXPECT_EQ(response, string("{\"success\":true}"));
+
+}
```
npoltorapavlo, line 475: I don't think such test makes sense

Tests/tests/test_FrontPanel.cpp
```diff
@@ -0,0 +1,486 @@
...
+
+TEST_F(FrontPanelDsTest, setLED)
+{
+    ON_CALL(frontPanelTextDisplayImplStringMock, setColor())
+        .WillByDefault(::testing::Return());
```
npoltorapavlo, line 480: please check value in setColor

Tests/mocks/devicesettings.h
```diff
@@ -980,111 +995,219 @@ class Host {
...
+
+                        return impl->setState(enable);
+                }
+		void setColor(Color color, bool persist){
+			return impl->setColor();
+			
+		}
+		void setColor(Color color){
+                 	return impl->setColor();
+
```
npoltorapavlo, line 1099: parameters should match

Tests/mocks/devicesettings.h
```diff
@@ -980,111 +995,219 @@ class Host {
...
+                {
+			
+                       return impl->setBrightness(brightness, persist);
+                }
+		void setBrightness(const int brightness)
+                {
+                        bool persist = true;
+                        return impl->setBrightness(brightness, persist);
+		
+                }
```
npoltorapavlo, line 1083: parameters should match

Tests/mocks/devicesettings.h
```diff
@@ -980,111 +995,219 @@ class Host {
...
+		void getBrightnessLevels(int &levels,int &min,int &max) const
+		{	
+        		return impl->getBrightnessLevels( levels, min, max);
+    		}
+		List<Color> getSupportedColors()
+		{
+			List<Color> color;
+			color.push_back(Color(kIndicatorColors[0].id));
+			return color;
+		}
```
npoltorapavlo, line 1068: please move test data to the test cases

Tests/mocks/devicesettings.h
```diff
@@ -980,111 +995,219 @@ class Host {
...
+
+                	Color(int id){
+			}
+			Color(){
+			}
+        	        
+			
+			std::string getName() const
+			{
+				return "white";
```
npoltorapavlo, line 1038: please move test data to the test cases

npoltorapavlo (changes requested):

Tests/tests/test_FrontPanel.cpp
hgfell683: @npoltorapavlo 
When you say check values, do you mean something like this:

```
ON_CALL(frontPanelIndicatorImplStringMock, setBrightness(::testing::_, ::testing::_))
        .WillByDefault(::testing::Invoke(
        [&](int brightness, bool persist){
        if(0<= brightness <= 100) //or whatever the check for valid parameters would be in this case
        {
                return;
        }
        else
         {
                //some kind of error response
         }
}));
```

npoltorapavlo: apparently should be EXPECT_CALL if intend to verify that the call is made
also EXPECT_EQ for important parameters
`EXPECT_EQ(brightness, 1);` ?

Tests/CMakeLists.txt
```diff
@@ -88,9 +89,10 @@ link_directories(../LocationSync
         ../DeviceInfo
         ../HdmiInput
         ../HdcpProfile
-        ../TraceControl
+	../TraceControl
         ../Network
 	../WifiManager
+        ../FrontPanel
```
npoltorapavlo, line 95: ../FrontPanel is added twice

npoltorapavlo (changes requested):

helpers/frontpanel.cpp
```diff
@@ -190,6 +190,21 @@ namespace WPEFramework
...
+                    delete mFrontPanelHelper;
+                    mFrontPanelHelper = nullptr;
+                }
+                delete s_instance;
+                s_instance = nullptr;
+                initDone = 0;
+                
+            }
+        }
+
```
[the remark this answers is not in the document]
hgfell683: The problem here is initDone, which is a global variable defined in frontpanel.cpp. When I was redoing the void test cases to use EXPECT_EQUAL, I discovered a issue within my tests(as USE_DS was not set, and this fpIndicators was always 0), because since CFrontPanel never leaves scope, initDone, after the first run, is set to true, and never set to false, thus causing some values to become unsettable, since they're only set if initDone is set to false.

```
        CFrontPanel* CFrontPanel::instance()
        {
            if (!initDone)
            {
                if (!s_instance)
                    s_instance = new CFrontPanel;

#ifdef USE_DS
                try
                {
                    LOGINFO("Front panel init");
                    fpIndicators = device::FrontPanelConfig::getInstance().getIndicators();

                    for (uint i = 0; i < fpIndicators.size(); i++)
                    {
                        std::string IndicatorNameIarm = fpIndicators.at(i).getName();

                        auto it = std::find(m_lights.begin(), m_lights.end(), IndicatorNameIarm);
                        if (m_lights.end() == it)
                        {
                            m_lights.push_back(IndicatorNameIarm);
                        }
                    }

#if defined(HAS_API_POWERSTATE)
                    {
                        IARM_Bus_PWRMgr_GetPowerState_Param_t param;
                        IARM_Result_t res = IARM_Bus_Call(IARM_BUS_PWRMGR_NAME, IARM_BUS_PWRMGR_API_GetPowerState,
                            (void*)&param, sizeof(param));

                        if (res == IARM_RESULT_SUCCESS) {
                            if (param.curState == IARM_BUS_PWRMGR_POWERSTATE_ON)
                                powerStatus = true;
                        }
                    }
#endif
#ifdef CLOCK_BRIGHTNESS_ENABLED
                    clockBrightness =  device::FrontPanelTextDisplay::getInstance("Text").getTextBrightness();
                    device::FrontPanelTextDisplay::getInstance("Text").setTextBrightness(clockBrightness);
#endif
                    globalLedBrightness = device::FrontPanelIndicator::getInstance("Power").getBrightness();
                    LOGINFO("Power light brightness, %d, power status %d", globalLedBrightness, powerStatus);

                    for (uint i = 0; i < fpIndicators.size(); i++)
                    {
                        LOGWARN("Initializing light %s", fpIndicators.at(i).getName().c_str());
                        if (powerStatus)
                            device::FrontPanelIndicator::getInstance(fpIndicators.at(i).getName()).setBrightness(globalLedBrightness);

                        device::FrontPanelIndicator::getInstance(fpIndicators.at(i).getName()).setState(false);
                    }

                    if (powerStatus)
                        device::FrontPanelIndicator::getInstance("Power").setState(true);

                }
                catch (...)
                {
                    LOGERR("Exception Caught during [CFrontPanel::instance]\r\n");
                }
                initDone=1;
#endif
            }

            return s_instance;
        }



```
Without a way to set initDone to 0 (which is never done in the normal code), we will get segmentation faults any time anytime fpIndicators is accessed after the first test (because it gets set once, and can never be set again, and by the following tests, we've cleared out that instance that fpIndicators was using)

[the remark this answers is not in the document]
hgfell683: Any suggestions on how to set initDone to 0 without adding a some kind of deinitialize function, since it doesn't get to set back to 0 anytime after it is set to 1. (Since we need to effectively re-set the values each time a new test is run, because we set the impl to a nullptr for the reference that is returned by getIndicators)

npoltorapavlo:
please fix already requested changes before requesting a new review

Tests/mocks/devicesettings.h
hgfell683: @npoltorapavlo The changes for this have already been made, I don't know why they're not reflecting here and sating the file is out dated. (This is what it looks like when viewing in the files changed--there are no white space changes anymore.)


update

apatel859 (approved):

npoltorapavlo:
opened https://ccp.sys.comcast.net/browse/RDKCOM-3525 to track issues introduced by this merge

PR #3430, opened 2022-11-16
RDKTV-20401: TV did not enter to Deepsleep , and stayed ~14 hrs in th…

PaviP98:
…e standby state

Maintenance window is open for long time and it was not completed successfully and hence TV didn't went to deepsleep.

Whenever stopMaintenance Api executed, it will terminate the task which is currently running and will send that event status to MM, also it will update the status of remaining tasks as well. Since MM didn't trigger those scripts it will simply ignore those events and so maintenance was not completed. Hence added changes to directly send the MAINTENANCE_ERROR event from stopMaintenance()

### DIFF
--- a/MaintenanceManager/MaintenanceManager.cpp
+++ b/MaintenanceManager/MaintenanceManager.cpp
@@ -288,14 +288,18 @@ namespace WPEFramework {
 #endif
 
             if ( false == internetConnectStatus ) {
+                m_statusMutex.lock();
                 MaintenanceManager::_instance->onMaintenanceStatusChange(MAINTENANCE_ERROR);
+                m_statusMutex.unlock();
                 LOGINFO("Maintenance is exiting as device is not connected to internet.");
                 return;
             }
 
             LOGINFO("Reboot_Pending :%s",g_is_reboot_pending.c_str());
 
+            m_statusMutex.lock();
             MaintenanceManager::_instance->onMaintenanceStatusChange(MAINTENANCE_STARTED);
+            m_statusMutex.unlock();
 
             if (UNSOLICITED_MAINTENANCE == g_maintenance_type){
                 LOGINFO("---------------UNSOLICITED_MAINTENANCE--------------");
@@ -629,7 +633,9 @@ namespace WPEFramework {
             MaintenanceManager::m_abort_flag=false;
 
             /* we post just to tell that we are in idle at this moment */
+            m_statusMutex.lock();
             MaintenanceManager::_instance->onMaintenanceStatusChange(m_notify_status);
+            m_statusMutex.unlock();
 
             m_thread = std::thread(&MaintenanceManager::task_execution_thread, _instance);
         }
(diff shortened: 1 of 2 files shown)


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -1287,7 +1286,7 @@ namespace WPEFramework {
                                 if (k_ret == 0){
                                     LOGINFO(" %s Termimated\n",script_names[i].c_str());
                                     /*this means we killed the task currently running */
-                                    task_incomplete = true;
+                                    m_task_map[task_names_foreground[i].c_str()]=false;
                                 }
```
mvalas827, line 1290: Shouldn't we break the loop here as we only run one task at a time and abort flag update should stop running any following scripts in line.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -1287,7 +1286,7 @@ namespace WPEFramework {
                                 if (k_ret == 0){
                                     LOGINFO(" %s Termimated\n",script_names[i].c_str());
                                     /*this means we killed the task currently running */
-                                    task_incomplete = true;
+                                    m_task_map[task_names_foreground[i].c_str()]=false;
```
mvalas827, line 1289: m_task_map is being updated from two threads.. plz check on possible race conditions.

mvalas827 (commented):
Test the change thoroughly by aborting at various intervals in different scenarios.

I see notify_status and m_notify_status not in sync. Try to include it as part of cleanup if you see any issues with it.

MaintenanceManager/MaintenanceManager.cpp
PaviP98: m_task_map is updated from stopMaintenance only. Test results are attached in the ticket https://ccp.sys.comcast.net/browse/RDKTV-20401?focusedCommentId=15064916&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15064916. Verified both unsolicited and solicited maintenance by invoking stopMaintenance api. No crashes or regressions observed with Maintenance Manager

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -862,6 +872,8 @@ namespace WPEFramework {
             else {
                 LOGWARN("Ignoring unexpected event - owner: %s, eventId: %d!!", owner, eventId);
             }
+          }
```
tdeva14, line 875: Can we have a log statement here to indicate that abort flag has been set and hence ignoring the event ?

PaviP98: Done

tdeva14 (commented):
Changes look good

mvalas827 (approved):
Looks good. Thanks

anand-ky (approved):

PR #3437, opened 2022-11-17
SecurityAgent: Secure REST API methods in a granular way

susanmary007:
This PR brings in the changes to secure REST API methods in a granular way. It also fixes some issues related to regex comparisons.

### DIFF
--- a/SecurityAgent/AccessControlList.h
+++ b/SecurityAgent/AccessControlList.h
@@ -40,8 +40,7 @@ namespace {
         string regex = input;
         
         // order of replacing is important
-        ReplaceString(regex,"*","^[a-zA-Z0-9.]+$");
-        ReplaceString(regex,".","\\.");
+        ReplaceString(regex,"*","[a-zA-Z0-9\.]*");
         
         return regex;
     }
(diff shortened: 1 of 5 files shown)


pradeeptakdas:
Internal JIRA for tracking 
https://ccp.sys.comcast.net/browse/RDKCOM-3431

SecurityAgent/AccessControlList.h
```diff
@@ -40,8 +40,7 @@ namespace {
         string regex = input;
         
         // order of replacing is important
-        ReplaceString(regex,"*","^[a-zA-Z0-9.]+$");
-        ReplaceString(regex,".","\\.");
```
npoltorapavlo, line 44: dot should be treated as dot

npoltorapavlo (changes requested):

SecurityAgent/AccessControlList.h
susanmary007: Thanks for the review @npoltorapavlo .

`CreateRegex() ` is used for generating regex for method/plugin names only. So that's why I was expecting `.` to be treated as a wildcard character.
So if `".nfo"` is added to the methods to allow, was expecting `"Info"` to be allowed (and was not expecting a method with name `".nfo"` )

But for URL parsing case, agree that `.` should be treated as literal dot itself. `CreateUrlRegex()` is used for creating regex for URLs and it has the logic to treat `.` as literal.
`ReplaceString(regex,".","\\.");`

This was my idea when making these changes. Please correct me if I am wrong.

npoltorapavlo: in acl `*` is wildcard. i.e. you can use "*nfo". in acl we have callsigns like `"org.rdk.Network"` (with dots)

susanmary007: Agree.. Thanks for the clarification @npoltorapavlo

PR #3452, opened 2022-11-21
SecurityAgent: Secure REST API methods in a granular way

susanmary007:
This PR brings in the changes to secure REST API methods in a granular way. It also fixes some issues related to regex comparisons.

### DIFF
--- a/SecurityAgent/AccessControlList.h
+++ b/SecurityAgent/AccessControlList.h
@@ -40,7 +40,7 @@ namespace {
         string regex = input;
         
         // order of replacing is important
-        ReplaceString(regex,"*","^[a-zA-Z0-9.]+$");
+        ReplaceString(regex,"*","[a-zA-Z0-9.]*");
         ReplaceString(regex,".","\\.");
         
         return regex;
@@ -57,7 +57,7 @@ namespace {
         ReplaceString(regex,":*",":[0-9]+");
         ReplaceString(regex,"*:","[a-z]+:");
         ReplaceString(regex,".","\\.");
-        ReplaceString(regex,"*","[a-zA-Z0-9\\.\\-]+");
+        ReplaceString(regex,"*","[a-zA-Z0-9\\.\\-]*");
         regex.insert(regex.begin(),'(');
         regex.insert(regex.end(),')');
         regex.insert(regex.begin(),'^');
@@ -295,7 +295,7 @@ namespace Plugin {
                     while ((index != _methods.end()) && (found == false)) { 
                         std::regex expression(index->c_str());
                         std::smatch matchList;
-                        found = std::regex_search(method, matchList, expression);
+                        found = std::regex_match(method, matchList, expression);
                         if (found == false) {
                             index++;
                         }
@@ -338,7 +338,7 @@ namespace Plugin {
                 while ((index != _plugins.end()) && (pluginFound == false)) {
                     std::regex expression(index->first.c_str());
                     std::smatch matchList;
-                    pluginFound = std::regex_search(callsign, matchList, expression);
+                    pluginFound = std::regex_match(callsign, matchList, expression);
                     if (pluginFound == false) {
                         index++;
                     }
--- a/SecurityAgent/SecurityAgent.cpp
+++ b/SecurityAgent/SecurityAgent.cpp
@@ -95,8 +95,11 @@ namespace Plugin {
         Config config;
         config.FromString(service->ConfigLine());
         string version = service->Version();
+        string webPrefix = service->WebPrefix();
+        string callsign = service->Callsign();
 
-        _skipURL = static_cast<uint8_t>(service->WebPrefix().length());
+        _skipURL = static_cast<uint8_t>(webPrefix.length());
+        _servicePrefix = webPrefix.substr(0, webPrefix.find(callsign));
         Core::File aclFile(service->PersistentPath() + config.ACL.Value(), true);
 
         PluginHost::ISubSystem* subSystem = service->SubSystems();
@@ -137,7 +140,7 @@ namespace Plugin {
             } else {
                 if (subSystem != nullptr) {
                     Core::SystemInfo::SetEnvironment(_T("SECURITYAGENT_PATH"), _dispatcher->Connector().c_str(), true);
-                    Core::Sink<SecurityCallsign> information(service->Callsign());
+                    Core::Sink<SecurityCallsign> information(callsign);
 
                     if (subSystem->IsActive(PluginHost::ISubSystem::SECURITY) != false) {
                         SYSLOG(Logging::Startup, (_T("Security is not defined as External !!")));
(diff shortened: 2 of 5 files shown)


npoltorapavlo (approved):

pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3499
CMF JIRA: https://jira.rdkcentral.com/jira/browse/RDKDEV-390

npoltorapavlo:
please bring to sprint/23Q1

npoltorapavlo:
@susanmary007 

in `SecurityContext::Allowed(const Web::Request& request)` you check access for callsign and method (which looks right), but only if prefix is _servicePrefix. The only possible value for _servicePrefix seems to be `/Service/SecurityAgent` (it's set in `SecurityAgent::Initialize`). So it only checks access to SecurityAgent itself? (which does not look right)
Also, if prefix is not `/Service/SecurityAgent`, e.g. `/Service/MyPlugin` then access is checked for "","". This means I don't get permissions for `MyPlugin`  but instead permissions for `"default"`. (this also does not look right)
So not sure, did you expect this to work? can I revert ?

susanmary007:
> @susanmary007
> 
> in `SecurityContext::Allowed(const Web::Request& request)` you check access for callsign and method (which looks right), but only if prefix is _servicePrefix. The only possible value for _servicePrefix seems to be `/Service/SecurityAgent` (it's set in `SecurityAgent::Initialize`). So it only checks access to SecurityAgent itself? (which does not look right) Also, if prefix is not `/Service/SecurityAgent`, e.g. `/Service/MyPlugin` then access is checked for "","". This means I don't get permissions for `MyPlugin` but instead permissions for `"default"`. (this also does not look right) So not sure, did you expect this to work? can I revert ?

Hi @npoltorapavlo 

`_servicePrefix = webPrefix.substr(0, webPrefix.find(callsign));`

`_servicePrefix` would be` /Service/` and not `/Service/SecurityAgent` .
Inside ` SecurityContext::Allowed`, it would filter out the plugin name and API name from the request, (by removing the characters till `/Service/` from the REST API request) and then use that to decide whether the request should be allowed or not.

npoltorapavlo:
thanks @susanmary007 . The problem is, this isn't working for the JSON-RPC requests where URL is `http://127.0.0.1:9998/jsonrpc` and callsign/method are passed in POST. In this case there's no prefixes so it uses permissions for "","". If `"default": "blocked"` it returns false on this stage and `SecurityContext::Allowed(const Core::JSONRPC::Message& message)` with actual callsign/method is not even called . Can we return true instead of `_accessControlList->Allowed("","")`?

susanmary007:
> thanks @susanmary007 . The problem is, this isn't working for the JSON-RPC requests where URL is `http://127.0.0.1:9998/jsonrpc` and callsign/method are passed in POST. In this case there's no prefixes so it uses permissions for "","". If `"default": "blocked"` it returns false on this stage and `SecurityContext::Allowed(const Core::JSONRPC::Message& message)` with actual callsign/method is not even called . Can we return true instead of `_accessControlList->Allowed("","")`?

Hi @npoltorapavlo 

But the changes are made in  `SecurityContext::Allowed(const Web::Request& request)` which gets triggered for only REST APIs.
For JSON-RPC requests, `SecurityContext::Allowed(const Core::JSONRPC::Message& message)` will be used (in which there are no modifications made as part of this PR) and it should work fine.

For example, if the acl file has the below configuration:
```
        "local": {
            "default": "allowed",
            "Power": {
                "default": "blocked",
                "methods": [
                    "setstate",
                    "State"
                ]
            }
        }
```

`"Power.1.setstate"` JSON-RPC call should be allowed and it should work fine.

Let me know if I am missing something here.

npoltorapavlo:
both are called in case of `http://127.0.0.1:9998/jsonrpc`. I think the related code in Thunder is :
https://github.com/rdkcentral/Thunder/blob/R2/Source/WPEFramework/PluginServer.h#L2637-L2655

susanmary007:
> both are called in case of `http://127.0.0.1:9998/jsonrpc`. I think the related code in Thunder is : https://github.com/rdkcentral/Thunder/blob/R2/Source/WPEFramework/PluginServer.h#L2637-L2655

Ok.. Thanks @npoltorapavlo .

But could you help me to understand what exactly is the usecase? Are you using JSON-RPC calls via HTTP, but you have the default rule set to "blocked" ?

npoltorapavlo:
this is needed for the Downloadable Application Containers (DAC) that will have their own tokens and ACLs

rjpdasilva:
Hi @npoltorapavlo,

We are understanding the issue, but still thing it doesn't seem right to just return `true`  of `_accessControlList->Allowed("", "")` from the Security Agent in this case.

The Default Security provider in Thunder (https://github.com/rdkcentral/Thunder/blob/R2/Source/WPEFramework/PluginServer.cpp#L104-L125) already has an exception for the `/jsonrpc` path in the URLs (https://github.com/rdkcentral/Thunder/blob/R2/Source/WPEFramework/PluginServer.cpp#L113), so I believe we probably could and should do the same for this case.

What do you think about changing https://github.com/rdkcentral/Thunder/blob/R2/Source/WPEFramework/PluginServer.h#L2637-L2655:
```cpp
                // See if we are allowed to process this request..
                if ((security == nullptr) || (security->Allowed(*request) == false)) {
                    request->Unauthorized();
                } else {
                    // If there was no body, we are still incomplete.
                    if (request->State() == Request::INCOMPLETE) {

                        Core::ProxyType<Service> service;
                        bool serviceCall;
                        uint32_t status = _parent.Services().FromLocator(request->Path, service, serviceCall);

                        request->Service(status, Core::proxy_cast<PluginHost::Service>(service), serviceCall);
                    } else if ((request->State() == Request::COMPLETE) && (request->HasBody() == true)) {
                        Core::ProxyType<Core::JSONRPC::Message> message(request->Body<Core::JSONRPC::Message>());
                        if ((message.IsValid() == true) && (security->Allowed(*message) == false)) {
                            request->Unauthorized();
                        }
                    }
                }
```
into something like:
```cpp
                // See if we are allowed to process this request..
                bool isUrlForJsonRpc = (request->Verb == Web::Request::HTTP_POST) && (request->Path == _parent.Configuration().JSONRPCPrefix());
                if ((security == nullptr) || (!isUrlForJsonRpc && security->Allowed(*request) == false)) {
                    request->Unauthorized();
                } else {
                    // If there was no body, we are still incomplete.
                    if (request->State() == Request::INCOMPLETE) {

                        Core::ProxyType<Service> service;
                        bool serviceCall;
                        uint32_t status = _parent.Services().FromLocator(request->Path, service, serviceCall);

                        request->Service(status, Core::proxy_cast<PluginHost::Service>(service), serviceCall);
                    } else if ((request->State() == Request::COMPLETE) && (request->HasBody() == true)) {
                        Core::ProxyType<Core::JSONRPC::Message> message(request->Body<Core::JSONRPC::Message>());
                        if ((message.IsValid() == true) && (security->Allowed(*message) == false)) {
                            request->Unauthorized();
                        }
                    }
                }
```

rjpdasilva:
Besides `/jsonrpc`, is there really any other URL path that makes sense?
If yes, would that fit in in some ACL entry using the existing (or adjusted) code in the `SecurityAgent` plugin?
Just returning `true` when we do not recognize the request, seems not to favor security much, but if this is the only option we have here, and taking into consideration that before it was unconditionally returning `true`, I think we could go for that, yes.

npoltorapavlo:
https://github.com/rdkcentral/rdkservices/pull/3735

PR #3457, opened 2022-11-21
RDK-37000 - Update thunder document for Key event code for sendKeyPre…

HariniElangovan890:
…ssEvent API


apatel859 (changes requested):
need to update json  file for docs. MD output will be generated by the script

PR #3462, opened 2022-11-22
Redirect logs to another systemd service.

asurdej-comcast:
With rdkbrowser2 all browser logs were part of a parent service. Webkit browser plugin prints logs as wpeframework service.

This change adds possibility to redirect all webkit browser plugin logs (stdout and stderr) to another systemd service to preserve any logging monitors and file watchers still working.

### DIFF
--- a/WebKitBrowser/LoggingUtils.cpp
+++ b/WebKitBrowser/LoggingUtils.cpp
@@ -1,0 +1,132 @@
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2022 RDK Management
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stddef.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <unistd.h>
+#include <syslog.h>
+#include <signal.h>
+#include <glib.h>
+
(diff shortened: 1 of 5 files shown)


emutavchi (dismissed):
@MFransen69 could you please review this too?

WebKitBrowser/LoggingUtils.cpp
```diff
@@ -0,0 +1,132 @@
...
+  }
+  const string kServiceExt = ".service";
+  if (target_service.size() <= kServiceExt.size() ||
+      target_service.substr(target_service.size() - kServiceExt.size(), kServiceExt.size()) != kServiceExt) {
+    fprintf(stderr, "RedirectLog: Invalid serivce name: *.service format expected\n");
+    return false;
+  }
+
+  const string targetServiceName = target_service.substr(0, target_service.size() - kServiceExt.size());
+  const string kSystemdCgroupTargetTasksFilePath = "/sys/fs/cgroup/systemd/system.slice/" + target_service + "/tasks";
```
MFransen69, line 49: Hi, could you please make inclusion of this code a build option and have it off by default. Now all webkit builds would get a dependency on systemd even if the option is not used.

MFransen69 (changes requested):

WebKitBrowser/LoggingUtils.cpp
emutavchi: @MFransen69, sure, I think compile time option can be added as well. but, there is no dependency on systemd here. the redirection is disabled by default...

MFransen69: Yes, understand but the code will be always linked in now, also when the redirection is off in the config, and calls like kSystemdCgroupTargetTasksFilePath need to be resolved even if not used for it to compile or am I missing something?

asurdej-comcast: there is no systemd compile time dependency in this code. It uses standard files and sockets APIs to interact with systemd. Any systemd specific stuff are handled in runtime, pointing to systemd specific file or connecting systemd specific socket

Anyway I can add compile time opt

MFransen69: Oops, my mistake. Sorry I didn't read very well and it looked like a systemd call at first sight :) Perhaps better to still make it build time conditional, a lot of low lever stuff being done which we normally would hide in the Thunder abstraction layer.

emutavchi (approved):

MFransen69 (approved):

PR #3478, opened 2022-11-26
RDK-38254 : Added Unit Tests for uploadReport call.

sborushevsky:


Tests/tests/test_Telemetry.cpp
```diff
@@ -234,3 +251,211 @@ TEST_F(TelemetryRfcTest, Plugin)
...
+                EXPECT_EQ(object, &rbObject);
+                EXPECT_EQ(string(name), _T("UPLOAD_STATUS"));
+                return &rbValue;
+            }));
+
+    EXPECT_CALL(rBusApiImplMock, rbusMethod_InvokeAsync(::testing::_, ::testing::_, ::testing::_, ::testing::_, ::testing::_))
+        .Times(1)
+        .WillOnce(::testing::Invoke(
+            [&](rbusHandle_t handle, char const* methodName, rbusObject_t inParams, rbusMethodAsyncRespHandler_t callback,  int timeout) {
+                sleep(2);
```
npoltorapavlo, line 422: [tests should be fast](https://github.com/rdkcentral/rdkservices/tree/sprint/22Q4/Tests#faq). a regular test suite takes milliseconds (without Valgrind/coverage). also should not rely on race condition. please remove `sleep()` from tests

npoltorapavlo (changes requested):

Tests/tests/test_Telemetry.cpp
sborushevsky: Removed sleep call.

Tests/tests/test_Telemetry.cpp
```diff
@@ -234,3 +251,335 @@ TEST_F(TelemetryRfcTest, Plugin)
...
+
+                return RBUS_ERROR_SUCCESS;
+            }));
+
+    EXPECT_CALL(rBusApiImplMock, rbus_close(::testing::_))
+        .Times(1)
+        .WillOnce(::testing::Invoke(
+            [](rbusHandle_t handle) {
+                return RBUS_ERROR_SUCCESS;
+            }));
```
npoltorapavlo, line 557: please don't [over-specify](https://github.com/rdkcentral/rdkservices/tree/sprint/22Q4/Tests#faq). `callback` is called, `onReportUpload` is checked so the test works. `rbus_open/rbusValue_GetString/rbusObject_GetValue/rbusMethod_InvokeAsync/rbus_close` calls are optional so can be `ON_CALL` instead of `EXPECT_CALL`. Also, if `::testing::Invoke` only returns, please use `::testing::Return`

npoltorapavlo (changes requested):

npoltorapavlo (approved):

PR #3494, opened 2022-12-01
RDKTV-19740: Modify the frequency of CEC messages sent from TV

dautapankumardora:
Reason for change: Improve the CEC retry mechanism on I2c cec bus busy
Test Procedure: Refer the ticket for test procedure
Risks: Low
Priority: P1
Signed-off-by: bp-tdora114 <dautapankumar.dora@sky.uk>

### DIFF
--- a/HdmiCecSink/HdmiCecSink.cpp
+++ b/HdmiCecSink/HdmiCecSink.cpp
@@ -321,6 +321,7 @@ namespace WPEFramework
 	     updateStatus = HdmiCecSink::_instance->deviceList[header.from.toInt()].m_isOSDNameUpdated;
 	     LOGINFO("updateStatus %d\n",updateStatus);
 	     HdmiCecSink::_instance->deviceList[header.from.toInt()].update(msg.osdName);
+	     HdmiCecSink::_instance->deviceList[header.from.toInt()].m_isRequestRetry = 0;
              if(!updateStatus)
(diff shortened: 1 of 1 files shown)


HdmiCecSink/HdmiCecSink.cpp
```diff
@@ -2470,7 +2477,12 @@ namespace WPEFramework
 
 					case CECDeviceParams::REQUEST_OSD_NAME :	
 					{
-						_instance->deviceList[logicalAddress].update(OSDName("NA"));
+						if ( _instance->deviceList[logicalAddress].m_isRequestRetry++ >= HDMICECSINK_REQUEST_MAX_RETRY )
```
apatel859, line 2480: So this code chanegs is adding retry to OSD name. Could we add this to VENDOR ID as well?

bijas: Vendor id is not that mandatory like physical address or osd name.so we really need not

apatel859: I know but providing vendor id reliably application write some logic for gaming console or something like that. So adding try there will help.

bijas: To be failr I dont want to add it now and make the cec bus busy.As we don't know which all devices provide that information and which ones dont.When it really needed we can add.

apatel859 (approved):

apatel859 (approved):

PR #3524, opened 2022-12-12
RDK-38124 Add Unit test for ActivityMonitor plugin

srikanth-vv:
- Added gtests for Activity Monitor plugin


Tests/tests/test_ActivityMonitor.cpp
```diff
@@ -0,0 +1,219 @@
...
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("disableMonitoring")));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("getApplicationMemoryUsage")));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Exists(_T("getAllMemoryUsage")));
+}
+
+TEST_F(ActivityMonitorTest, disableMonitoring)
+{
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("disableMonitoring"), _T("{}"), response));
+    EXPECT_EQ(response, _T("{\"success\":true}"));
+}
```
npoltorapavlo, line 70: I don't think such test makes sense

Tests/tests/test_ActivityMonitor.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+TEST_F(ActivityMonitorTest, enableMonitoringEmptyConfig)
+{
+    EXPECT_EQ(Core::ERROR_GENERAL, handler.Invoke(connection, _T("enableMonitoring"), _T("{\"config\":[],\"memoryIntervalSeconds\":\"0\", \"cpuIntervalSeconds\":\"0\"}"), response));
+    EXPECT_EQ(Core::ERROR_GENERAL, handler.Invoke(connection, _T("enableMonitoring"), _T("{\"config\":[{\"appPid\": 6763, \"memoryThresholdMB\": 10,\"cpuThresholdPercent\": 50,\"cpuThresholdSeconds\": 2}],\"memoryIntervalSeconds\":\"0\", \"cpuIntervalSeconds\":\"0\"}"), response));
+}
```
npoltorapavlo, line 76: where do these magical numbers come from?

Tests/tests/test_ActivityMonitor.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+protected:
+    IarmBusImplMock iarmBusImplMock;
+    ActivityMonitorMemoryUsageTest() : ActivityMonitorTest()
+    {
+        //Add version info
+        std::ofstream fileVer("/version.txt");
+        fileVer << "imagename:PX051AEI_VBN_2203_sprint_20220331225312sdy_NG\nSDK_VERSION=17.3\nMEDIARITE=8.3.53\nYOCTO_VERSION=dunfell\n";
+        fileVer.close();
+
```
npoltorapavlo, line 88: it's not clear why version.txt is needed for tests. can you please check if it works without this file?

Tests/tests/test_ActivityMonitor.cpp
```diff
@@ -0,0 +1,219 @@
...
+        //Add Apps registry file
+        Core::File fileApps(regFile);
+        fileApps.Create();
+        fileApps.Write(regFileData, sizeof(regFileData));
```
npoltorapavlo, line 92: please move test data to the test cases

npoltorapavlo (changes requested):

Tests/tests/test_ActivityMonitor.cpp
srikanth-vv: Removed.

Tests/tests/test_ActivityMonitor.cpp
srikanth-vv: These can be any random numbers. The test case is "enableMonitoring" should return "ERROR_GENERAL", with a valid parameters, since the app registry file is not available.

Tests/tests/test_ActivityMonitor.cpp
srikanth-vv: The function "isDevOrVBNImage" uses the image name to check if it is PROD or DEV.

npoltorapavlo: > can you please check if it works without this file

srikanth-vv: I verified and without version.txt:
- getAllMemoryUsage() - Returns empty application memory data as "r`esponse={"freeMemoryMB":7198,"applicationMemory":[],"success":true}`"
- enableMonitoring() - Returns "`ERROR [ActivityMonitor.cpp:899] monitoring: Failed to determine memory usage for 4808`"

npoltorapavlo (approved):

npoltorapavlo:
please bring to sprint/23Q1
https://github.com/rdkcentral/rdkservices/pull/3567 pending

PR #3543, opened 2022-12-19
RDKDEV-461: Use IARM_Bus_RemoveEventHandler instead of IARM_Bus_UnRegisterEventHandler

mikolaj-staworzynski-red:
IARM_Bus_RemoveEventHandler is complementary function for IARM_Bus_RegisterEventHandler.
    
While using IARM_Bus_UnRegisterEventHandler the following could happen:
(1) Assume usage of HdmiCec and HdcpProfile plugins, both in in-process mode
(2) Activate both plugins - they both register for IARM_BUS_DSMGR_EVENT_HDMI_HOTPLUG
(3) Deactivate one of the plugin (for example HdcpProfile): IARM_Bus_UnRegisterEventHandler would de-register both listeners registered in point (2)
(4) As a result HdcpProfile plugin will not receive IARM_BUS_DSMGR_EVENT_HDMI_HOTPLUG events anymore.
    
Change introduced for 5 used plugins: HdmiCec, HdcpProfile, XCast, DisplayInfo and DisplaySettings
Additionally DisplaySettings now de-register for all the events - some of them are not there previously

There is one a bit confusing behaviour with return code returned by IARM_Bus_RemoveEventHandler for the case when two or more handlers are registered, patch is provided here: https://code.rdkcentral.com/r/c/rdk/components/generic/iarmbus/+/78036


pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3522

PR #3561, opened 2022-12-21
RDKDEV-390: RDKCOM-3499 SecurityAgent: Secure REST API methods in a granular way

Sukanya673:
Currently the access to REST APIs is configured in a "all or nothing" approach ie, if an application is given access to a plugin's REST API, it will have access to all the available REST APIs.

This commit brings in the changes to support setting secure access to
 REST APIs in a API granular level so that an application can be allowed or
denied access to only certain REST APIs of a plugin.

(cherry picked from commit 370feed9bcba2130c51889372551984e231ea2f2)

RDKDEV-390: SecurityAgent: Fix issues related to regex comparisons

It was observed that if a method "State" is added to rejected methods'
 list, Filter::Allowed() would reject "SetState" / "GetState" etc..
ie, it would reject all methods having "State".

To fix this issue, "regex_search" is replaced with "regex_match". "regex_match" returns true only when the entire input sequence has been matched, while "regex_search" will succeed even if only a sub-sequence matches the regex.

The same change is made in Filter::Plugin::Allowed() for the same reason.

Another issue observed was, when "*State" is added to the rejected methods' list, "State" / "SetState" etc.. was not rejected. A similar issue was observed for "url" also.
The regex matching (eg:- localhost*) was not working properly.

This commit fix the issue by correcting the regex creation in CreateUrlRegex() and CreateRegex().

(cherry picked from commit cd3c07bf9bbc1bd04249afebb8ca2aec8c2eb6f2)


anand-ky:
@Sukanya673 - Pls update version macros also to match the version update in the CHANGELOG - https://github.com/rdkcentral/rdkservices/blob/9a2b53102abd7cece9f7c1ba83ba36a7e63f4354/SecurityAgent/SecurityAgent.cpp#L26

anand-ky (changes requested):
https://github.com/rdkcentral/rdkservices/blob/9a2b53102abd7cece9f7c1ba83ba36a7e63f4354/SecurityAgent/SecurityAgent.cpp#L26

anand-ky (approved):

PR #3563, opened 2022-12-22
Revert "DELIA-59417 Thunder plugin failing when building for g-unit t…

npoltorapavlo:
…ests for rdkservices (#3536)"

This reverts commit 07ca744c4e787fea285e730cbc869f0910d218e0.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   unit-tests:
     name: Build and run unit tests
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


npoltorapavlo:
reverting because the fix was already done is https://github.com/rdkcentral/rdkservices/pull/3525
reverting because the mentioned fix is already part of the current THUNDER_REF "0731e67a8c9dea37a728201c9d19df5fbdfcbfe5"
reverting because there's no reason to avoid ubuntu-latest considering the mentioned fix

anand-ky (approved):

PR #3567, opened 2022-12-22
Some tests are not working without valgrind

npoltorapavlo:


npoltorapavlo:
disabled tests that rely on uninitialized variables or other and not working without Valgrind

npoltorapavlo:
for https://github.com/rdkcentral/rdkservices/issues/3568

anand-ky (approved):

PR #3570, opened 2022-12-22
RDKDEV-464  rdkservice: XCast: explicit linking with cjson

mikolaj-staworzynski-red:
In the following configuration with XCast:

RFC_ENABLED not defined while compilation :  dependency to librfcapi.so is not present in linked library: libWPEFrameworkXCast.so 

If we build libWPEFrameworkXCast.so we would like to have explicit dependency to cjson.

That is probably not an issue when linked with RFC_ENABLED (librfcapi.so depends on cjson)


mikolaj-staworzynski-red:
replaced by: https://github.com/rdkcentral/rdkservices/pull/3606

PR #3586, opened 2022-12-29
RDK-28534: Upstream security agent patch

npoltorapavlo:


npoltorapavlo:
can't bring to sprint/23Q1, pending https://github.com/rdkcentral/rdkservices/pull/3452

PR #3589, opened 2023-01-02
RDKDEV-480 XCast add default values in place of missing RFC data

mikolaj-staworzynski-red:
For the case with RFC_ENABLED not defined. Default return values changed to true
for  IsDynamicAppListEnabled/IsAppEnabled


mikolaj-staworzynski-red:
replace by: https://github.com/rdkcentral/rdkservices/pull/3607

PR #3591, opened 2023-01-02
XIONE-11185, RDKTV-21162: Fix locking issues in MM

PaviP98:
Fix iarmEventhandler locking issues in MM


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -292,6 +292,9 @@ namespace WPEFramework {
                 MaintenanceManager::_instance->onMaintenanceStatusChange(MAINTENANCE_ERROR);
                 m_statusMutex.unlock();
                 LOGINFO("Maintenance is exiting as device is not connected to internet.");
+		if (UNSOLICITED_MAINTENANCE == g_maintenance_type){
```
tdeva14, line 295: better to check for "!g_unsolicited_complete" as well just like the other condition

PR #3599, opened 2023-01-03
RDKTV-11877 : Added define to disable geography based Time Zone setting.

sborushevsky:


anand-ky:
@MFransen69 @pwielders - Pls review locationSync change. This is based on the discussions we had with Sky last month (https://etwiki.sys.comcast.net/display/RDKV/Timezone+Changes) where we don't want to set timezone based on geolocation and rely on the timezone coming from account information and being set through setTimeZoneDST.
We are adding a build flag to control TZ being set from LocationSync.

MFransen69 (commented):
Hi,

I understand the change but I think it will cause confusion as Thunder as well as the LocationSync will now report a timezone that is not actually set. 
Is there actually a use for LocationSync in these systems? Why not turn it off completely?
As an alternative may I propose to make the whole TimeZone handling configurable? (so in that case not only exclude the SetEnvironment call as done in this proposal but make the whole timezone handling here and in the Thunder subsystems consistent in case it is turned off and also not do it with a build flag but with a configuration option).
And as the LocationSync in RDKServices and the Metrological implementation are diverging as we could not sync our changes as the required Thunder PR was not merged to R2 until only recently I would propose that we add this configuration option in our latest version and then make sure it is synched to RDKServices.

anand-ky (approved):

PR #3608, opened 2023-01-04
RDKDEV-481: ONEMPERS-287 XCast: retry reconnection till success

mikolaj-staworzynski-red:
The commit changes XCast plugin reconnection behavior. The previous implementation retries connection to the remote server up to 4 times, with the increasing interval before retries. There is no reason to stop retrying as plugin without this connection is useless, so right now it repeats the fourth retry (60 seconds timeout) till success.


anand-ky:
@apatel859 - Pls review

apatel859 (approved):

anand-ky (approved):

PR #3614, opened 2023-01-04
RDK-37628 Adapted RustBridge into rdkservices

goruklu:
Adapted RustBridge from https://github.com/rdkcentral/ThunderNanoServices/tree/master/RustBridge so it can work as rdkservice. It requires a change like [0001_rust_bridge_interface.txt](https://github.com/rdkcentral/rdkservices/files/9797315/0001_rust_bridge_interface.txt) in the ThunderInterfaces.

### DIFF
--- a/RustBridge/RustBridge.cpp
+++ b/RustBridge/RustBridge.cpp
@@ -149,11 +149,20 @@ namespace WPEFramework {
 			// The expectation is that the JavaScript service opens up a connection to us, so we can forward the 
 			// incomming requests, to be handled by the Service.
 			if (_javascriptService == 0) {
+				/*
 				Web::ProtocolsArray protocols = channel.Protocols();
 				if (std::find(protocols.begin(), protocols.end(), string(_T("json"))) != protocols.end()) {
 					_javascriptService = channel.Id();
 					assigned = true;
 				}
+				*/
+
+				//TODO: probably this condition should be applied
+				//if (channel.State == PluginHost::Channel::ChannelState.JSON)
+				//{
+					_javascriptService = channel.Id();
+					assigned = true;
+				//}
 			}
 			return(assigned);
 		}
@@ -167,7 +176,7 @@ namespace WPEFramework {
 		// -------------------------------------------------------------------------------------------------------
 		//   IDispatcher methods
 		// -------------------------------------------------------------------------------------------------------
-		Core::ProxyType<Core::JSONRPC::Message> RustBridge::Invoke(const Core::JSONRPC::Context& context, const Core::JSONRPC::Message& inbound) /* override */
+		Core::ProxyType<Core::JSONRPC::Message> RustBridge::Invoke(const string& token, const uint32_t channelId, const Core::JSONRPC::Message& inbound) /* override */
 		{
 			string method;
 			Registration info;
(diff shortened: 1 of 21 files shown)


RustBridge/RustBridgeImplementation.cpp
```diff
@@ -294,8 +294,10 @@ namespace WPEFramework {
 					
 					searchPath.push_back(framework->DataPath());
 					searchPath.push_back(framework->PersistentPath());
-					searchPath.push_back(framework->SystemPath());
-					searchPath.push_back(framework->PluginPath());
+					//searchPath.push_back(framework->SystemPath());
+					//searchPath.push_back(framework->PluginPath());
+					searchPath.push_back(framework->VolatilePath());
+					searchPath.push_back("/usr/lib/wpeframework/plugins/");
```
pwielders, line 300: This is not the recommended way to hardcode paths in the code. It will lead to specific code deployments and hard to reuse code. I think the path indicated here is the path that is under compilations from recipes exactly the SystemPath, so why would you comment the system path and hard code it here ?

RustBridge/RustBridge.cpp
```diff
@@ -149,11 +149,20 @@ namespace WPEFramework {
 			// The expectation is that the JavaScript service opens up a connection to us, so we can forward the 
 			// incomming requests, to be handled by the Service.
 			if (_javascriptService == 0) {
+				/*
 				Web::ProtocolsArray protocols = channel.Protocols();
 				if (std::find(protocols.begin(), protocols.end(), string(_T("json"))) != protocols.end()) {
 					_javascriptService = channel.Id();
 					assigned = true;
 				}
+				*/
```
pwielders, line 158: This is dropping functionality that is, according to the websocket spec allowed, see: https://www.rfc-editor.org/rfc/rfc6455#section-1.9 . In R2, this header was treated as a string (so only 1 subprotocol could be considered, in R3, on request from some customers, it was required to follow the spec more closely and make this, as the spec suggests a list of identifiers, hence why compilation in R2 fails as that is not a list, it is a string (see https://github.com/rdkcentral/Thunder/blob/R2/Source/websocket/WebSocketLink.h#L649).
Suggest to , in R2, to stll check the subprotocol header but just "scan" for "json" so that it is mandataory at least for the service to use this header as an upgrade scenario, alwasy prepare for the future, don't code for it but think about it ;-)
So a good replacement in R2 could be (pseudeo code):
'''
if (channel.Protocol().find("json") != string::npos) {
    _javascriptService = channel.Id();
    assigned = true;
}
'''

RustBridge/RustBridgeImplementation.cpp
goruklu: removed the hard coded path

RustBridge/RustBridge.cpp
goruklu: agreed, replaced with a simple search for "json" in the subprotocol string as you recommended

goruklu:
@pwielders I also re-enabled the method support check code. We'll just use a patch to disable it in our builds where needed.

goruklu:
requires https://github.com/rdkcentral/ThunderInterfaces/pull/226

anand-ky (approved):

PR #3619, opened 2023-01-05
SERXIONE-1281 : Fix deadlock in Maintenance Manager code

PaviP98:
Reason for change: If AS calls startMaintenance before we acquire internet access, the Unsolicited maintenance cycle launched from boot will be locked due to the same mutex used in the startMaintenance task.

Test Procedure: Verify that the maintenance tasks launch as expected

Risks: Low
Priority: P0


tdeva14 (approved):
Change looks good

anand-ky (approved):

PR #3628, opened 2023-01-06
DELIA-59688: Upstream security agent patch 231

npoltorapavlo:


gladish (commented):
Can this be used instead of copying the buffer into a std::string?

SYSLOG(Logging::Notification, (_T("Creating Token for %.*s"), length, buffer));

anand-ky (approved):

PR #3630, opened 2023-01-06
RDK-39371: Added rdkvfwupgrader support during stop maintenance

satya200:
Reason for change: During stop maintenance mm needs pid for the process
                   So if the firmwaredownload rfc true need to check
                   rdkvfwupgrader pid insted of deviceInitiated.sh

Test Procedure: 1> Trigger download.
                2> When download is in progress run stop maintenance

Risks: Low

Priority: P1

Signed-off-by: satya200 <satyasundar_sahu@comcast.com>


Jagadheesan-d:
Changes look okay, and as discussed with Maheedhar, will be taking up unifying differences in script calls at various places in a different ticket.

anand-ky (approved):

PR #3637, opened 2023-01-09
RDKAPI-1 - Enhance UsbAccess RDK Service to support multiple USB drives

srikanth-vv:
- Modified UsbAccess APIs as per [Support for multiple USB drives](https://etwiki.sys.comcast.net/pages/viewpage.action?spaceKey=RDKV&title=UsbAccess+RDK+Service+support+for+multiple+USB+drives)

- Added fix for [RDKTV-21360](https://ccp.sys.comcast.net/browse/RDKTV-21360)


UsbAccess/UsbAccess.cpp
```diff
@@ -216,12 +274,55 @@ namespace Plugin {
...
+			pathParam = parameters["path"].String();
+
+		bool bLinkExists = false;
+
+        //Loop through all the existing IDs and set bLinkExists to true, if it already has the link
+        for(auto itr : m_CreatedLinkIds)
+        {
+            string linkPath = LINK_PATH+std::to_string(itr);
+			bLinkExists = isSymlinkExists(linkPath, pathParam);
+			if(bLinkExists)
```
anand-ky, line 290: If link already exists, log suitable message and return that link. It shouldn't error out.

UsbAccess/UsbAccess.cpp
```diff
@@ -129,6 +129,47 @@ namespace Plugin {
                 mod_time = st.st_mtime;
             return mod_time;
         }
+
+        bool isAbsPath(const string& mountPath, const string& givenPath) {
+            if(givenPath.rfind(mountPath, 0) == 0)
+                return true;
+            else
+                return false;
```
anand-ky, line 137: Pls check for trailing slashes in the path and trim them out before doing a find.

anand-ky (approved):
Pls make follow-up PR to update UsbAccess.json to update documentation for these APIs. Also address the 2 review comments as we discussed.

UsbAccess/UsbAccess.cpp
srikanth-vv: Fixed in pr: https://github.com/rdkcentral/rdkservices/pull/3673

UsbAccess/UsbAccess.cpp
srikanth-vv: The code already takes care of the trailing slashes, since getMounted() API doesn't return the trailing slash and the rfind() method gets succeed with or without trailing slash.

PR #3651, opened 2023-01-11
RDKDEV-649: Run correctly as an out-of-process service

steveford-obs:
Run the DVB correctly as an out-of-process plugin so the DVB stack isn't running in the WPEFramework process. This protects the wpeframework process if the DVB crashes.

Split the functionality so the DVB libraries are only included in the plugin that runs out-of-process so the size of the WPEFramework process is reduced.


pradeeptakdas:
@steveford-obs , can you please create a PR against sprint/23Q1?

pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3545

PR #3658, opened 2023-01-13
REFPLTV-1573: RDKServices: LoadFailed event is not triggerring with t…

sravanthi298:
…he URL which is failed to load by the browser

Reason for change: LoadFailed is displaying for valid url.

Test Procedure: LoadFailed event should trigger based on URL loaded by browser.

Risks: Low

Signed-off-by: srowth513 <Sravanthi_Rowthu@comcast.com>

### DIFF
--- a/WebKitBrowser/WebKitImplementation.cpp
+++ b/WebKitBrowser/WebKitImplementation.cpp
@@ -1962,8 +1962,6 @@ static GSourceFuncs _handlerIntervention =
         {
             _adminLock.Lock();
 
-            _URL = URL;
-
             std::list<Exchange::IWebBrowser::INotification*>::iterator index(_notificationClients.begin());
             {
                 while (index != _notificationClients.end()) {
--- a/WebKitBrowser/WebKitBrowser.cpp
+++ b/WebKitBrowser/WebKitBrowser.cpp
@@ -290,7 +290,7 @@ namespace Plugin {
 
     void WebKitBrowser::LoadFinished(const string& URL, int32_t code)
     {
-        string message(string("{ \"url\": \"") + URL + string("\", \"loaded\":true, \"httpstatus\":") + Core::NumberType<int32_t>(code).Text() + string(" }"));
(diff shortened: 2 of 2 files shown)


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -1962,8 +1962,6 @@ static GSourceFuncs _handlerIntervention =
         {
             _adminLock.Lock();
 
-            _URL = URL;
```
emutavchi, line 1965: This will make URL property query return stale values. And it is not clear why it is removed. So, I'd prefer not to submit this change.

emutavchi (changes requested):

WebKitBrowser/WebKitImplementation.cpp
sravanthi298: When curl command with Invalid URL param is issued, then LoadFailed log is displaying for both invalid and valid URL's  i.e, https://www.tdk-page-test.com/ (invalid) and https://www.google.com/(valid).

Code flow:

Here, in (*index)->LoadFailed(_URL); (https://github.com/rdkcentral/rdkservices/blob/8cbee6bc86dfd911485ca63365798996ecee816c/WebKitBrowser/WebKitImplementation.cpp#L2026) function call, _URL is taking previous loaded URL and showing LoadFailed for previously loaded URL instead of current invalid URL.

Observed _URL is updated in OnURLChanged() method as shown (_URL = URL;)
https://github.com/rdkcentral/rdkservices/blob/8cbee6bc86dfd911485ca63365798996ecee816c/WebKitBrowser/WebKitImplementation.cpp#L1965

And also _URL is used only to pass as argument for LoadFailed (https://github.com/rdkcentral/rdkservices/blob/8cbee6bc86dfd911485ca63365798996ecee816c/WebKitBrowser/WebKitImplementation.cpp#L2026). Nowhere it is being used.

So commented the line to resolve the logging issue.

sravanthi298:
Will recheck-in with proper fix.

PR #3684, opened 2023-01-19
REFPLTV-1573: RDKServices: LoadFailed event is not triggerring with the URL which is failed to load by the browser

sravanthi298:
REFPLTV-1573: RDKServices: LoadFailed event is not triggerring with the URL which is failed to load by the browser

Reason for change: LoadFailed log is displaying for valid url.

Test Procedure: LoadFailed event should trigger based on URL loaded by browser.

Risks: Low

Signed-off-by: srowth513 <Sravanthi_Rowthu@comcast.com>


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2016,7 +2016,7 @@ static GSourceFuncs _handlerIntervention =
 
             _adminLock.Unlock();
         }
-        void OnLoadFailed()
+        void OnLoadFailed(const string _URL)
```
emutavchi, line 2019: This could be very confusing, i mean shadowing the class property with an argument. Lets make this explicit and make a change in notification call bellow.  i.e. name the parameter as 'URL' and pass it to the notification handler below.

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2564,7 +2564,7 @@ static GSourceFuncs _handlerIntervention =
                 browser->_ignoreLoadFinishedOnce = true;
                 return;
             }
-            browser->OnLoadFailed();
+            browser->OnLoadFailed(failingURI);
```
emutavchi, line 2567: You need similar change for the case when browser plugin is built with WK APIs (wpe-2.22 or older)

emutavchi (changes requested):

emutavchi (approved):

modeveci (dismissed):
LGTM, @MFransen69 can you also look at this?

MFransen69 (approved):

modeveci (approved):

emutavchi (approved):

PR #3704, opened 2023-01-25
RDK-37901: Gtest for RDKShell Plugin

zhilarasi:


Tests/tests/test_RDKShell.cpp
```diff
@@ -0,0 +1,1099 @@
...
+                                                                                "}"), response));
+   ON_CALL(compositormock, enableInputEvents(::testing::_, ::testing::_))
+            .WillByDefault(::testing::Invoke(
+                [](const std::string& client, bool enable){
+		      EXPECT_EQ(client, string("searchanddiscovery"));
+		      EXPECT_EQ(enable, true);
+                      return true;
+                   }));
+
+   EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("enableInputEvents"), _T("{"
```
npoltorapavlo, line 148: handler.Invoke doesn't need a previous handler.Invoke. please don't mix several tests in one. if it can be a separate test make it a separate test. this applies to all tests in this file

Tests/tests/test_RDKShell.cpp
```diff
@@ -0,0 +1,1099 @@
...
+{
+    ServiceMock service;
+    EXPECT_CALL(compositormock, showWatermark())
+	    .Times(::testing::AnyNumber())
+         .WillRepeatedly(::testing::Return(true));
+    EXPECT_EQ(string(""), plugin->Initialize(&service));
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("showWatermark"), _T("{\"show\":true}"), response));
+    EXPECT_EQ(response, _T("{\"success\":true}"));
+    ON_CALL(compositormock, hideWatermark())
+            .WillByDefault(::testing::Invoke(
```
npoltorapavlo, line 394: if ::testing::Invoke only returns use ::testing::Return instead

Tests/mocks/base64.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
npoltorapavlo, line 1: for RDKShell , please replace trower-base64 with WPEFramework built-in alternative - Core::ToString. this will allow to remove trower-base64 from RDKShell/CMakeLists.txt, and remove this file and its mock

npoltorapavlo (changes requested):

Tests/mocks/base64.h
binuinbaraj: Ezhil experimented using WPEFramework ToString, but for a 1080p image from RDKShell compositor, the image data size needs a uint32. WPEFramework ToString supports takes a uint16, so we cannot use this. 
` void ToString(const uint8_t object[], const uint16_t length, const bool padding, string& result) `
Can we take removing trower dependency later.

npoltorapavlo: I think this is a matter of refactoring. As a first step it can be a wrapper with the same interface and unit tests for it. This should allow to remove trower-base64 from builds, reduce image size, simplify testing

PR #3710, opened 2023-01-25
RDK-38588 - GTests for HdmiCec

hgfell683:
Reason for change:
Adding Gtests for HdmiCec
TestProcedure: None
Risks: Low

### DIFF
--- a/Tests/mocks/devicesettings.h
+++ b/Tests/mocks/devicesettings.h
@@ -421,6 +421,17 @@ typedef enum _dsDisplayEvent_t {
     dsDISPLAY_EVENT_MAX
 } dsDisplayEvent_t;
 
+enum {
+	ACTIVE_SOURCE = 0x82,
+	ROUTING_CHANGE = 0x80,
+	ROUTING_INFORMATION = 0x81,
+	SET_STREAM_PATH = 0x86,
+};
+
+typedef struct _Throw_e{
+
+}Throw_e;
+
 namespace device {
 
 template <class T>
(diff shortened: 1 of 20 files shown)


Tests/mocks/devicesettings.h
```diff
@@ -1218,3 +1231,420 @@ class FrontPanelConfig {
...
+		MAX_LENGTH = 128,
+	};
+	CECFrameImpl* impl;
+	void getBuffer(const uint8_t **buf, size_t *len) const{
+		return getInstance().impl->getBuffer(buf, len);
+	}
+	CECFrame(const uint8_t *buf = NULL, uint16_t len = 0){
+		
+	}
+	private:
```
npoltorapavlo, line 1285: please don't copy -paste unnecessary code. e.g. here `CECFrame::getInstance()`  is not used, class `CECFrameImpl` is not used, MAX_LENGTH is not used, `len_` is not used. Reading further, `LogicalAddress::getInstance()` is not used, `LogicalAddressImpl` is not used, etc.

npoltorapavlo, line 1285: in tests there's no practical need in `private` keyword

Tests/mocks/devicesettings.h
```diff
@@ -1215,6 +1228,399 @@ class FrontPanelConfig {
...
+
+};
+class CECBytesImpl {
+	public:
+	virtual const std::string toString() const = 0;
+};
+class CECBytes{
+    public:
+    CECBytesImpl* impl;
+    static CECBytes& getInstance()
```
npoltorapavlo, line 1283: if code creates CECBytes object and calls toString on it that will trigger object's impl, not CECBytes::getInstance().impl. No service code uses CECBytes::getInstance(). No static methods on CECBytes that need a static mock. So getInstance() is not needed. Also, I can't see any usages of CECBytes in project. Is it needed?

npoltorapavlo (changes requested):

Tests/mocks/devicesettings.h
hgfell683: @npoltorapavlo So, in the base code, several classes inherit from CECBytes, which is the class that has toString(). There are several classes that are needed, but not used (because they are called through the notify function, which is private, and not called directly within the code. Notify() does get called in actual through some weird use of inheritance and overriding). Using the inheritance of CECBytes, I can have the classes that don't get called and not throw errors without needed to create additional mock files that won't be used. I can revert to not using CECBytes, if its preferred.

npoltorapavlo: ok, but how does it work if you set mock on a static CECBytes (`CECBytes::getInstance().impl = &cecBytesImplMock;`) and in CECBytes::toString(void) you call impl->toString(); and not getInstance().impl->toString();?

hgfell683: Oh, I see what you mean. You're right, it should be getInstance().impl->toString.

Tests/mocks/devicesettings.h
```diff
@@ -1215,6 +1228,399 @@ class FrontPanelConfig {
...
+class LogicalAddressImpl{
+	public:
+		virtual int toInt() const = 0;
+		virtual int getType() const = 0;
+        virtual std::string toString() const = 0;
+
+};
+class LogicalAddress : public CECBytes{
+	public: 		
+		static const LogicalAddress kTv;
```
npoltorapavlo, line 1302: how is kTv used?

Tests/mocks/devicesettings.h
```diff
@@ -1215,6 +1228,399 @@ class FrontPanelConfig {
...
+		return impl->getBuffer(buf, len);
+	}
+	CECFrame(const uint8_t *buf = NULL, uint16_t len = 0){}
+
+};
+class CECBytesImpl {
+	public:
+	virtual const std::string toString() const = 0;
+};
+class CECBytes{
```
npoltorapavlo, line 1280: can you help understand what are all those CECBytes classes for, do they carry any logic other than keeping a value passed in constructor and converting it to string? if that's the case you can just put some trivial implementation in here

npoltorapavlo (changes requested):

Tests/mocks/devicesettings.h
hgfell683: ``So, as an example, in the base code, cecBytes toString function looks like this:

`virtual const std::string toString(void) const {
		std::stringstream stream;
		for (size_t i = 0; i < str.size(); i++) {
			stream << std::hex << (int)str[i];
		}
		return stream.str();
    };`
    
    So it pretty much is just convering it to a string.

Tests/mocks/devicesettings.h
hgfell683: It doesn't seem to be needed upon an initial look. Will remove, and make sure it's not actually needed.

Tests/mocks/devicesettings.h
hgfell683: @npoltorapavlo Done and done.

npoltorapavlo: now you can remove CECBytesImpl, LogicalAddressImpl, PhysicalAddressImpl, LogicalAddress& getInstance(), and implement CECBytes(const uint8_t *buf, size_t len)

npoltorapavlo (changes requested):

Tests/mocks/devicesettings.h
hgfell683: Updated, though CECBytes(const uint8_t *buf, size_t len) seems to be needed, in its relation to some of the the construtors of its children.

npoltorapavlo:
Please make sure the test names are clear and reflect what the test does. If the test verifies that a plugin sends an event when IARM triggers it, please name it correspondingly, for example "MyPlugin" "SendsMyEventWhenIARMTriggers" . Set up expectations , for example `IShell::Submit` is known to be called when an event is sent, this function is mockable. I would suggest to not write a test if you don't know what to verify and cannot derive a descriptive test name. Also, in any kind of code (not just tests) I don't think it is normal to resolve race conditions with `sleep()`. There should exist a more efficient and reliable way.

npoltorapavlo:
Like was requested, the fixture/test names should correspond to what they do. If HdmiCecInitializedEventDsTest is a fixture to verify events, then why is it used in tests that don't verify events? It is confusing that you use ContainsRegex as there's no volatile things, like time, cpu load. `ContainsRegex(_T(".*\"success\":true.*")` checks any object in any part of response can have \"success\":true?

Tests/tests/test_HdmiCec.cpp
```diff
@@ -0,0 +1,258 @@
...
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("getDeviceList"), _T(""), response));
+    //m_osdName and m_vendorID do not get populated in the population function, like logical address, and thus are empty with actual implementtion of .toString()
+    EXPECT_EQ(response, string(_T("{\"numberofdevices\":14,\"deviceList\":[{\"logicalAddress\":1,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":2,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":3,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":4,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":5,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":6,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":7,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":8,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":9,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":10,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":11,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":12,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":13,\"osdName\":\"\",\"vendorID\":\"\"},{\"logicalAddress\":14,\"osdName\":\"\",\"vendorID\":\"\"}],\"success\":true}")));
+
+
+}
```
npoltorapavlo, line 258: It is not clear what you test in getDeviceList. It's a black box with functionality unknon to the reader. It is not clear what's the setup - why sleep is needed. It is not clear what are the magic values returned, where they come from, are they valid/not valid. If there's no good test, don't test please.

npoltorapavlo (changes requested):

npoltorapavlo:
https://github.com/rdkcentral/rdkservices/pull/3816 implements similar mocks and classes

Tests/tests/test_HdmiCec.cpp
hgfell683: @npoltorapavlo 

The reason the sleep is needed is that we're dealing with threads. HDMICec is not designed to be booted up, and immediately try and grab variables that are assigned in the the threads that get spun of on it's creation. The sleep is in there, is such that we try and wait until something gets populated (ideally the entire array of 16), before calling the the function the gets deviceList. Without a sleep, we will get a random number of devices, anywhere from 0-16, depending on how long errant processes took.

As for what we are testing, this is the important part of the getDeviceListFunction
```
int i = 0;
			for(i=0; i< LogicalAddress::UNREGISTERED; i++ ) {
				if (BIT_CHECK(deviceList[i].m_deviceInfoStatus, BIT_DEVICE_PRESENT)) {
					JsonObject device;
					device["logicalAddress"] = HdmiCec::_instance->deviceList[i].m_logicalAddress.toInt();
					device["osdName"] = HdmiCec::_instance->deviceList[i].m_osdName.toString().c_str();
					device["vendorID"] = HdmiCec::_instance->deviceList[i].m_vendorID.toString().c_str();
					deviceListArg.Add(device);
				}
			}
```

Earlier, since the implementation was simple, you told me to use the actual implementation of the simple functions (toString, to toInt, etc), rather then mocking and explicitly saying what values to return. Thus these are the values that those actual implementations return in this instance.

I can do two things, change the actual implementation for toString and toInt back to the mocked version, such that I can specify the returns , and check specifically for those values (which does work, as that was the implementation I had before the change), or I remove it, which I don't really want to do. I know what we're testing(as it's just getting a list of the 'connected devices' and the various addresses that are related to it), I just can't control what the actual implementation returns in this instance.

anand-ky (approved):

PR #3555, opened 2022-12-21
TraceControl: Add module name to trace

paulo-a-junior:
The current implementation of the TraceControl plugin won't output the origin of the trace.

With this commit, the resulting log message will have the form: 

`[<time>]:[<module-name>]:[<file-name:line-number>] <log-category>: <data>`

This would allow for easier handling of log messages in a scenario where multiple plugins have their tracing flag enabled at the same time.

This change is targeting only the non-abbreviated form of the logs, as the understanding is that adding the module name would make it "less abbreviated".

To capture syslog messages for a given module, use `journalctl |grep <module-name>`.

PR #3718, opened 2023-01-30
TraceControl: Add module name to trace

paulo-a-junior:
See: https://github.com/rdkcentral/rdkservices/pull/3555

The current implementation of the TraceControl plugin won't output the origin of the trace.

With this commit, the resulting log message will have the form: 

`[<time>]:[<module-name>]:[<file-name:line-number>] <log-category>: <data>`

This would allow for easier handling of log messages in a scenario where multiple plugins have their tracing flag enabled at the same time.

This change is targeting only the non-abbreviated form of the logs, as the understanding is that adding the module name would make it "less abbreviated".

To capture syslog messages for a given module, use `journalctl |grep <module-name>`.


paulo-a-junior:
Please disconsider this PR.

PR #3722, opened 2023-01-30
TraceControl: Add module name to trace

paulo-a-junior:
Forward-port of https://github.com/rdkcentral/rdkservices/pull/3555. See PR for details

Current implementation of the TraceControl plugin won't output the origin of the trace.

With this commit, the resulting log message will have the for: [<time>]:[<module-name>]:[<file-name:line-number>] <log-category>: <data>

This would allow for easier handling of log messages in a scenario where multiple plugins have their tracing flag enabled at the same time.

To capture syslog messages for a given plugin, use `journalctl |grep <module-name>`.


anand-ky:
Need to update changelog and versioning - https://github.com/rdkcentral/RDKServices#versioning

anand-ky:
Merged as part of https://github.com/rdkcentral/rdkservices/pull/3831

PR #3728, opened 2023-02-01
RDKTV-21748: Removing OSD name Updated to FA on Feature Abort

dhivya-ilangovan:
Reason for change: OSD name is set as "NA" by default and set to "FA" when Feature Abort command is received. Changing this values to empty string ""
Test Procedure: Steps mentioned in ticket
Risks: Low


dhivya-ilangovan:
Created new PR: https://github.com/rdkcentral/rdkservices/pull/3737 to avoid multiple commits

PR #3729, opened 2023-02-02
RDK-37910 Added HdmiCecSink test framework update

ranjithkumas:
RDK-37910 Added HdmiCecSink test framework update


npoltorapavlo:
When adding changes on review, please avoid unnecessary whitespace changes. It obscures review and can cause conflicts when merging. Regarding the code in Tests/mocks/ I think you have a misunderstanding. It is not a folder with copy-pasted files. It's mocks (fakes). The normal procedure is,  you create an empty file, for example ccec.h, then you build and see the compilation errors, and implement the missing definitons until it compiles. There's no other reason to add code in Tests/mocks/. Regarding the tests, your task is like, write  tests for HdmiCecSink. That is, you need to determine what HdmiCecSink does, for example what "requestAudioDevicePowerStatus" does. If you can't explain what it does, and the test doesn't address it, the test is pretty much useless and we don't need it.

npoltorapavlo:
As already requested, When adding changes on review, avoid unnecessary whitespace changes. As already requested, do not blindly copy paste. As already requested, If you can't explain what API does, don't test it. This is not unit testing. Nobody needs this nonsense. Pleae close the pr.

PR #3730, opened 2023-02-03
RDKTV-21475: [One time issue]: HDMI-AVR is not auto connected when we update the build from 5.11p2s3 to 5.11p3s3

aishwariya15:
RDKTV-21475: [One time issue]: HDMI-AVR is not auto connected when we update the build from 5.11p2s3 to 5.11p3s3

Reason for change: Adding debug prints to debug the issue if the issue is observed in thread exiting.
Test Procedure: Build and Verify.
Risks: Low
Signed-off-by: Aishwariya B <aishwariya.bhaskar@sky.uk>


aishwariya15:
Wrong sequence of commits

PR #3737, opened 2023-02-06
RDKTV-21748: Removing OSD name Updated to FA on Feature Abort

dhivya-ilangovan:
Reason for change: OSD name is set as "NA" by default and set to "FA" when Feature Abort command is received. Changing this values to empty string ""
Test Procedure: Steps mentioned in ticket
Risks: Low


HdmiCecSink/HdmiCecSink.cpp
```diff
@@ -440,7 +440,7 @@ namespace WPEFramework
 					
 					case GIVE_OSD_NAME :
 			 		{
-			 			HdmiCecSink::_instance->deviceList[header.from.toInt()].update(OSDName("FA"));
+			 			HdmiCecSink::_instance->deviceList[header.from.toInt()].update(OSDName(""));
```
apatel859, line 443: any reason why (VendorID((uint8_t *)"FA" is not udpated at 443 line?

PR #3752, opened 2023-02-09
RDKTV-21739-In USB drive, File name having special characters like &,…

srikanth-vv:
- Modified regex to list all files except starting with . (They are treated as hidden files)


srikanth-vv:
Special characters are causing issues with media player [(RDKTV-21929)](https://ccp.sys.comcast.net/browse/RDKTV-21929). This solution may not work.

PR #3753, opened 2023-02-09
RDKDEV-677: Change to resolve deadlock and timeout when tuning using user defined tuning parameters

steveford-obs:
Changed DTV StartServiceSearch functions to directly include tuning parameters, rather than using the separately created tuning parameter objects because accessing one interface object from another seemed to be causing deadlock.


pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3610

anand-ky (approved):

PR #3781, opened 2023-02-15
SERXIONE-1534: First install JailUI not rebooting device when connect…

PaviP98:
…ed to Wifi

There is some race conditio in MM code when startMaintenance() was invoked by AS within seconds multiple times. Hence Fixed race condition in MM


tdeva14 (approved):
Change looks good

anand-ky (approved):

PR #3782, opened 2023-02-15
New apis of memory pressure limits

modeveci:
In 2.28 (and before that), we passed the memory limits to the MemoryPressureHandler through the env
vars WPE_POLL_MAX_MEMORY and WPE_POLL_MAX_MEMORY_GPU (for the video
memory). In 2.38 those env vars are not used, and there's API to set those values (and more) through the WebKitMemoryPressureSettings type.

**WebKitMemoryPressureSettings** can be used for both **WebProcessLimit** and **NetworkProcessLimit** (in MB). The default value is the size of RAM with maximum of 3GB.

(New APIs currently not configured by Browser plugin) 

GPU memory limit is by default 1GB and can be manipulated by **webkit_memory_pressure_settings_set_video_memory_limit** in MB. 

Set the percentage of the limits where the process will be killed. Default is 0 which means
processes are never killed. When set to 1.1 for example, if the limit set is 100MB, processes
will be killed when they use more than 110MB. This is applied both to normal and GPU memory.
**webkit_memory_pressure_settings_set_kill_threshold(settings, 1);**

Set the percentage of the limits where the strict policy kicks in. By default this is 0.5. The
strict policy means that the system tries harder to free memory. This is applied both to normal and GPU memory.
**webkit_memory_pressure_settings_set_strict_threshold**(settings, 0.75);

Set the percentage of the limits where the conservative policy kicks in. By default this is 0.33. The
conservative policy means that the system tries to free memory not too hard. When the memory usage is under
the conservative level, wpe doesn't try to free memory at all. This is applied both to normal and GPU memory.
**webkit_memory_pressure_settings_set_conservative_threshold**(settings, 0.5);

Set the number of seconds to pass between memory measurements. This is 30 seconds by default. This should
not be smaller than 1 second. Bear in mind that very frequent polls will degrade the performance.
**webkit_memory_pressure_settings_set_poll_interval**(settings, 3);

Bear in mind that the env var **WPE_POLL_MAX_MEMORY_GPU_FILE** is still
used to hold the path of the file that contains the amount of video
memory being used by the system.


MFransen69 (approved):

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2158,10 +2173,12 @@ static GSourceFuncs _handlerIntervention =
                 Core::SystemInfo::SetEnvironment(_T("MSE_MAX_BUFFER_SIZE"), _config.MSEBuffers.Value(), !environmentOverride);
             }
 
+#ifndef WEBKIT_MEMORY_PRESSURE_API
             // Memory Pressure
             if (_config.MemoryPressure.Value().empty() == false) {
```
emutavchi, line 2178: Hi @modeveci .  I think it'd be cleaner if memory pressure value is just reinterpreted here in 2.38 case. I mean instead of setting environment variable, you could parse it and set the settings using new API in the worker later. 

Changing browser config will require changes in build recipes in different meta layers, and can leave browser in wrong configuration if those aren't updated....

modeveci: Thank you for your feedback @emutavchi. @MFransen69 could you please review this request?

PR #3787, opened 2023-02-16
RDK-37916: Unit test for OCIContainer

melhar098:


melhar098:
@npoltorapavlo I have applied the changes if you could please review

Tests/tests/test_OCIContainer.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+
+ TEST_F(OCIContainerInitializedTest, getContainerStateTest){
+
+    std::list<std::pair<int32_t, std::string>> containerslist = {{91, "com.bskyb.epgui"},{94, "Netflix"}};
+    EXPECT_CALL(dobbymock, listContainers())
+        .WillOnce(::testing::Return(containerslist));
+
+    EXPECT_CALL(dobbymock, getContainerState(91))
+        .WillOnce(::testing::Return(2));
```
npoltorapavlo, line 88: Please use constants/enums intead of the magic numbers.

npoltorapavlo (changes requested):

Tests/tests/test_OCIContainer.cpp
melhar098: I've made the changes if you could review them when you have a chance please.

Tests/mocks/Dobby.h
```diff
@@ -0,0 +1,294 @@
...
+
+    IIpcService* impl;
+
+    IpcService(const std::string& dbusAddress, const std::string& serviceName, int defaultTimeoutMs = -1){
+
+    }
+
+    IpcService(){
+        IpcService("test", "test");
+    }
```
npoltorapavlo, line 64: Why are you creating an instance of class inside class constructor? Same in DobbyProxy().

Tests/tests/test_OCIContainer.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+ TEST_F(OCIContainerInitializedTest, listContainersTest){
+
+    std::list<std::pair<int32_t, std::string>> containerslist = {{descriptor1, containerId1},{descriptor2, containerId2}};
+    EXPECT_CALL(dobbymock, listContainers())
+        .WillOnce(::testing::Return(containerslist));
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(connection, _T("listContainers"), _T("{}"), response));
+    EXPECT_EQ(response, string("{\"containers\":[{\"Descriptor\":91,\"Id\":\"com.bskyb.epgui\"},{\"Descriptor\":94,\"Id\":\"Netflix\"}],\"success\":true}"));
+}
```
npoltorapavlo, line 88: This doesn't seem like you understand the concern. When you set id 91 and check id 91 it is fine, reader understands it is the same id. When you set 2 and check "Running" it's not fine, reader doesn't understand 2 = "Running".  When you set "descriptor1" and check 91 it's not fine, reader doesn't understand "descriptor1" = 91. When you set "containerId2" and check "Netflix" it's not fine, reader doesn't  understand "containerId2" = "Netflix". I hope you can see the point.

npoltorapavlo (changes requested):

melhar098:
@npoltorapavlo I added an enum for ContainerState and removed the class instances

Tests/tests/test_OCIContainer.cpp
```diff
@@ -8,13 +8,17 @@
...
-namespace{
-    const int32_t descriptor1 = 91;
-    const int32_t descriptor2 = 94;
-    const string containerId1 = "com.bskyb.epgui";
-    const string containerId2 = "Netflix";
-    const int stateRunning = 2;
-    const int listenerId = 5;
+namespace {
+
+    enum ContainerState {
```
npoltorapavlo, line 13: Looks like you have the same enum as IDobbyProxyEvents::ContainerState. Why not just use it?

Tests/tests/test_OCIContainer.cpp
```diff
@@ -75,40 +80,40 @@ TEST_F(OCIContainerTest, RegisteredMethods){
...
-
-    std::list<std::pair<int32_t, std::string>> containerslist = {{descriptor1, containerId1},{descriptor2, containerId2}};
+TEST_F(OCIContainerInitializedTest, getContainerInfoTest)
+{
+    std::list<std::pair<int32_t, std::string>> containerslist = {{91, "com.bskyb.epgui"},{94, "Netflix"}};
     EXPECT_CALL(dobbymock, listContainers())
         .WillOnce(::testing::Return(containerslist));
 
-    EXPECT_CALL(dobbymock, getContainerInfo(descriptor1))
+    EXPECT_CALL(dobbymock, getContainerInfo(91))
```
npoltorapavlo, line 116: Where do other magic numbers come from, for example "pids":[13132,13418] or "limit" : 419430400 ?

npoltorapavlo (changes requested):

Tests/tests/test_OCIContainer.cpp
melhar098: These numbers I have being returned were taken from a real world example of what we expect to see being returned from this method. OCIContainer::getContainerInfo calls on DobbyProxy::getContainerInfo to return this information in a string

Tests/tests/test_OCIContainer.cpp
melhar098: I have applied this change if you could review whenever you get a chance please

npoltorapavlo (approved):

PR #3797, opened 2023-02-21
RDKTV-21929 - Multiple Issues seen when TV has 2 USBs connected

srikanth-vv:
- Filter out directories that has characters outside ASCII range and ?
- This is a temporary fix until locales are implemented.


anand-ky:
Putting this on hold until we add support for UTF8 chars.

srikanth-vv:
This is a work around and not needed since utf-8 is supported

PR #3798, opened 2023-02-21
RDKTV-21739 - Expand regex and add debug prints

srikanth-vv:
- Changed regex to allow all the files
- Added debug prints to log file & directory names in wpeframework.log


srikanth-vv:
Worng branch, I need xumo_1

PR #3806, opened 2023-02-27
DELIA-60222: Fix undefined symbols on qemu x86 platform

Scony:
On qemu x86 platform the gcc with `-O2` is used. Such flag leaves few symbols undefined (stripped when `-Os` is used) therefore making the injected bundle shared object not loadable.


emutavchi (dismissed):

MFransen69 (dismissed):

Scony:
Added same changes for `libWPEExtension.so` as well

MFransen69 (approved):

emutavchi (approved):

PR #3694, opened 2023-01-23
DELIA-59848: Gensmak app launch ends up in APPS-04035 error

varumugam123:
Reason for change: A failed resource loda from a child page from the app results in maintenance of incorrect http status code value in WebKitImplementation (as code is read on every policy decision callback, irrespective of the source frame i.e main frame or not)... Test Procedure: None
Priority: P1
Risks: None

Signed-off-by: Vivek.A <vivek_arumugam@comcast.com>

PR #3813, opened 2023-03-01
DELIA-59848: Gensmak app launch ends up in APPS-04035 error

varumugam123:
Reason for change: A failed resource loda from a child page from the app results in maintenance of incorrect http status code value in WebKitImplementation (as code is read on every policy decision callback, irrespective of the source frame i.e main frame or not)... Test Procedure: None
Priority: P1
Risks: None


varumugam123:
This is a cherry-pick of https://github.com/rdkcentral/rdkservices/pull/3694

emutavchi (approved):

PR #3816, opened 2023-03-02
RDK-37910 Added HdmiCecSink test framework

ranjithkumas:
RDK-37910 Added HdmiCecSink test framework

### DIFF
--- a/Tests/mocks/Operands.h
+++ b/Tests/mocks/Operands.h
@@ -1,0 +1,869 @@
+#pragma once
+
+
+#include <stdint.h>
+#include <cstring>
+#include <cstdio>
+#include <vector>
+#include <sstream>
+#include <string>
+#include <exception>
+typedef uint32_t Op_t;
+
+enum {
+    ACTIVE_SOURCE = 0x82,
+    FEATURE_ABORT = 0x00,
+    ABORT = 0xFF,
+    GIVE_AUDIO_STATUS = 0X71,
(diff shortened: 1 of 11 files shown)


Tests/mocks/Operands.h
```diff
@@ -0,0 +1,850 @@
+#pragma once
```
npoltorapavlo, line 1: "Operands.h" doesn't sound clear. If these classes relate to ccec, it can be "ccec.h", or a more appropriate name.  There is a pr with conflicting change https://github.com/rdkcentral/rdkservices/pull/3710/files#diff-f57c4c7c1defa058360a77be6fbc5dadbb9e2d19ccb354da66e7156ab9252ecc . I think we need only one variant of these classes.

npoltorapavlo (changes requested):

Tests/mocks/Operands.h
ranjithkumas: Modified "Operands.h" as "HdmiCec.h" , For #3710  also will use this "HdmiCec.h" in moving forward got confirmation with [hgfell683] .

npoltorapavlo:
As was requested already (in previous pr) please fix your editor to stop replacing spaces with tabs and doing other unnecessary whitespace changes in literally every line of original code.

anand-ky:
Closing older PRs. Pls create new PRs on latest sprint branch.

PR #3846, opened 2023-03-10
[RDK-40623] Unit Test for System Plugin API: getDeviceInfo

sethulakshmisk:
Added Skelton Test APIs for all the success and failure use cases of getDeviceInfo


anand-ky (commented):
Skeleton looks good.

Tests/tests/test_SystemServices.cpp
```diff
@@ -1306,3 +1307,445 @@ TEST_F(SystemServicesEventIarmTest, onRebootRequest)
...
+}
+
+/**
+ * @brief : getDeviceInfo  When QueryParam passed without label "param"
+ *          Check if QueryParams  contains no label as "params"
+ *          then getDeviceInfo shall successfully retrieve the device info and  returns it in the response
+ * @param[in]   :   "params" :{}
+ * @return      : {"make":"SKY","success":true}
+ */
+TEST_F(SystemServicesTest, getDeviceInfoSuccess_onQueryParameterHasNoLabelParam)
```
anand-ky, line 1643: Shouldn't this return all the parameters if no param is passed. I see we are checking only for make?

PR #3847, opened 2023-03-10
RDKTV-22146 DALS support in hdmicecSink

kprasad96:
Added methods for handling Dynamic Auto LipSync feature. 
1. Request Current Latency - Process to handle the incoming <Request Current Latency> message and respond with <Report Current Latency>
2. reportCurrentLatencyWrapper - To get the Latency Values such as Video Latency, Low Latency Mode, Audio Output Compensated and Audio Output Delay from DisplaySettings whenever there is change in latency .
  The above method will invoke reportCurrentLatency() 
3. reportCurrentLatency - To encode the latency values and Broadcast it to Connected Devices

### DIFF
--- a/HdmiCecSink/HdmiCecSink.cpp
+++ b/HdmiCecSink/HdmiCecSink.cpp
@@ -1036,7 +1036,7 @@ namespace WPEFramework
 		return;
 
 	    LOGINFO("Send Report Current Latency message \n");
-	    _instance->smConnection->sendTo(LogicalAddress::BROADCAST,MessageEncoder().encode(ReportCurrentLatency(physical_addr,video_latency,latency_flags,audio_output_delay)));
+	    _instance->smConnection->sendTo(LogicalAddress::BROADCAST,MessageEncoder().encode(ReportCurrentLatency(physical_addr,m_video_latency,m_latency_flags,m_audio_output_delay)));
 
         }
 
--- a/HdmiCecSink/HdmiCecSink.cpp
+++ b/HdmiCecSink/HdmiCecSink.cpp
@@ -64,6 +64,7 @@
 #define HDMICECSINK_METHOD_SEND_GIVE_AUDIO_STATUS          "sendGetAudioStatusMessage"
 #define HDMICECSINK_METHOD_GET_AUDIO_DEVICE_CONNECTED_STATUS   "getAudioDeviceConnectedStatus"
 #define HDMICECSINK_METHOD_REQUEST_AUDIO_DEVICE_POWER_STATUS   "requestAudioDevicePowerStatus"
+#define HDMICECSINK_METHOD_SET_LATENCY_INFO	"setLatencyInfo"
 
 #define TEST_ADD 0
 #define HDMICECSINK_REQUEST_MAX_RETRY 				3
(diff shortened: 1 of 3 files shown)


HdmiCecSink/HdmiCecSink.cpp
```diff
@@ -64,6 +64,8 @@
 #define HDMICECSINK_METHOD_SEND_GIVE_AUDIO_STATUS          "sendGetAudioStatusMessage"
 #define HDMICECSINK_METHOD_GET_AUDIO_DEVICE_CONNECTED_STATUS   "getAudioDeviceConnectedStatus"
 #define HDMICECSINK_METHOD_REQUEST_AUDIO_DEVICE_POWER_STATUS   "requestAudioDevicePowerStatus"
+#define HDMICECSINK_METHOD_REQUEST_CURRENT_LATENCY	"requestCurrentLatency"
```
anand-ky, line 67: Pls update semantic versioning minor version and changelog since this has new API changes. Also update minor version macro.

anand-ky (changes requested):

HdmiCecSink/HdmiCecSink.cpp
```diff
@@ -83,6 +85,10 @@
 #define SYSTEM_AUDIO_MODE_OFF 0x00
 #define AUDIO_DEVICE_POWERSTATE_OFF 1
 
+#define DEFAULT_VIDEO_LATENCY 100
```
apatel859, line 88: Are these value defiend somewhere else? What if thosd define gets updated?

HdmiCecSink/HdmiCecSink.h
```diff
@@ -624,6 +630,11 @@ namespace WPEFramework {
             std::queue<SendKeyInfo> m_SendKeyQueue;
             std::condition_variable m_sendKeyCV;
 	    std::condition_variable m_ThreadExitCV;
+	    
+	    /* DALS - Latency Values */
+	    uint8_t video_latency;
```
apatel859, line 635: rename m_

HdmiCecSink/HdmiCecSink.json
```diff
@@ -567,7 +567,66 @@
...
+	    "params": {
+                "type":"object",
+                "properties": {
+                    "physicalAddress":{
+                        "summary": "Physical address of the target device",
+                        "type":"string",
+                        "example": "0.0.0.0"
+                    },
+		    "videoLatency": {
+			"summary": "Indicates the Video Latency of the target device",
```
apatel859, line 601: in what unit??

HdmiCecSink/HdmiCecSink.json
```diff
@@ -829,4 +864,4 @@
         }
         
     }
-}
```
apatel859, line 832: Do you want to delete this?

kprasad96: It has been added in the next line

HdmiCecSink/HdmiCecSink.json
```diff
@@ -567,7 +567,66 @@
...
+                        "summary": "Physical address of the target device",
+                        "type":"string",
+                        "example": "0.0.0.0"
+                    },
+		    "videoLatency": {
+			"summary": "Indicates the Video Latency of the target device",
+			"type":"integer",
+			"example": 2
+		    },
+		    "lowLatencyMode": {
```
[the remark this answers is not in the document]
kprasad96: The incoming low latency Mode value would be in "int" format. And it has to be added as bit 2 in latency_flags variable.
Hence converting it to bool would require one more typecasting while adding it latency_flags.

HdmiCecSink/HdmiCecSink.cpp
kprasad96: For now, default values are not implemented on TV Server side. Latency Info will be updated only when there is a change in latency and hence here is the only place where default values are being handled

apatel859 (approved):

kprasad96:
Bringing the changes as a Single commit

PR #3867, opened 2023-03-16
RDK-39845 : Added uploadLogsAsync and abortLogUpload calls.

sborushevsky:


SystemServices/SystemServices.cpp
```diff
@@ -2540,6 +2572,79 @@ namespace WPEFramework {
...
+    }
+
+    uint32_t SystemServices::abortLogUpload(const JsonObject& parameters, JsonObject& response)
+    {
+
+        lock_guard<mutex> lck(m_uploadLogsMutex);
+
+        if (-1 != m_uploadLogsPid) {
+
+            // Kill child processes
```
anand-ky, line 2604: Can we look at aborting log upload without killing the processes. This might leave log abort operation in unknown state. Checking if we can add option in uploadLogs script to gracefully abort log upload.

anand-ky (changes requested):

anand-ky (approved):

anand-ky:
Pls submit follow-up PR to make json changes for new uploadAsync, abortLogUpload APIs & onlogUpload event. Also deprecate UploadLogs in the schema so that it reflects in documentation.

PR #3870, opened 2023-03-18
RDK-38587 - GTests for Xcast plugin

hgfell683:
RDK-38587 - GTests for Xcast plugin
Reason for change:
Adding Gtests for Xcast plugin
TestProcedure: None
Risks: Low


Tests/mocks/devicesettings.h
```diff
@@ -1208,3 +1208,319 @@ class FrontPanelConfig {
 
 }
 
+
+#define RT_OK 0
```
npoltorapavlo, line 1212: is it part of devicesettings?

npoltorapavlo (changes requested):

mhughesacn:
Hi @hgfell683
BlackDuck is complaining about a code match at devicesettings.h lines 1278-1303 which matches quite closely to pxScene/pxCore src/rtRef.h.  If this is the source, it must be credited by inserting a comment above the class definition at l.1268:
Based on pxCore, Copyright 2015-2018 John Robinson

If you have another source, please comment.
Thank you - Martin Hughes, RDK CMF Compliance Team

npoltorapavlo:
@hgfell683 can you please not copy paste. What exactly from rtRef do you need for the unit tests

hgfell683:
@npoltorapavlo Basically, the rtObject files are all interconnected through a series inheritance, mainly through the overloading of operators, and how, holding objects that are then used for various comparisons. The main function that uses rtRef is rtObjectRef through the use of template functions/classes in rtRef (specifically inherited by rtRef<rtIObject>). rtObjectRef uses most of the template function that comes from rtRef<rtIObject> for most of its comparisons within the xCast plugin. I've tried changing and modifying how this works to slim down on the added code, but most of the time editing the functionality in rtRef causes some kind of error within the code.

npoltorapavlo:
@hgfell683 at a glance, rtRef is a smart pointer class. Can we use std smart pointer instead?

hgfell683:
@npoltorapavlo I could try and see if that works, but I don't think rtRef is a smart pointer, as it's a template class. Template classes, and smart pointers do have similar syntax, as they both use <> to define themselves when used.

Tests/mocks/devicesettings.h
hgfell683: RT_OK, RT_ERROR and so on are apart of the rtError class in the base code, which are included in the various files where the values are used(xCast, and the rtObject classes). I can move it elsewhere, if you'd prefer it to be somewhere else.

mhughesacn:
Sorry, @hgfell683 - Please will you make sure that new code files have proper Apache headers?  This won't block but is still required.  Thank you!

anand-ky (approved):

PR #3872, opened 2023-03-20
RDKTV-22068 - Updated Network Plugin to handle the boot order of netsrmgr

karuna2git:
Updated Network Plugin to handle the boot order of netsrmgr.
- Removed error message that was returned if Initialize() was called when the netsrvmgr is NOT running.
- Removed the thread that runs continuously to look for the netsrvmgr process and subscribe to events.
- Changed the retry logic into 1sec timeout and only once if the plugin is not inited (or the netsrvmgr dependency was not met)


npoltorapavlo:
if Network plugin is part of wpeframework service, and it requires netsrvmgr service, shouldn't this be done via systemd, for example After=netsrvmgr.service in wpeframework.service?

PR #3874, opened 2023-03-20
RDK-40858 : Identify the device 's network state for Offline TTS

cmurug728:
### **To identify the network state** 

1) For the first time isConnected method Will return the device state by calling Network Plugin's **isConnectedToInternet** API and registering the event **onConnectionStatusChanged**
2)Later based on the registered event m_isConnected is updated and returns the state

### DIFF
--- a/TextToSpeech/impl/NetworkStatusObserver.cpp
+++ b/TextToSpeech/impl/NetworkStatusObserver.cpp
@@ -1,0 +1,113 @@
+#include "NetworkStatusObserver.h"
+
+#if defined(SECURITY_TOKEN_ENABLED) && ((SECURITY_TOKEN_ENABLED == 0) || (SECURITY_TOKEN_ENABLED == false))
+#define GetSecurityToken(a, b) 0
+#define GetToken(a, b, c) 0
+#else
+#include <WPEFramework/securityagent/securityagent.h>
+#include <WPEFramework/securityagent/SecurityTokenUtil.h>
+#endif
+
+#define MAX_SECURITY_TOKEN_SIZE 1024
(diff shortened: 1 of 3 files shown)


TextToSpeech/impl/NetworkStatusObserver.cpp
```diff
@@ -0,0 +1,113 @@
...
+        m_networkService = new WPEFramework::JSONRPC::LinkType<Core::JSON::IElement>(_T(NETWORK_CALLSIGN_VER),"", false, token);
+
+        JsonObject joGetParams, joGetResult;
+        uint32_t status = m_networkService->Invoke<JsonObject, JsonObject>(1000, "isConnectedToInternet", joGetParams, joGetResult);
+        if (status > 0) {
+            TTSLOG_ERROR("%s call failed %d",NETWORK_CALLSIGN_VER, status);
+            return false;
+        } else if (joGetResult.HasLabel("connectedToInternet")) {
+             m_isConnected=joGetResult["connectedToInternet"].Boolean();
+        }
```
varumugam123, line 89: If "connectedToInternet" label is not present, should we be resetting the m_isConnected to false?

TextToSpeech/impl/NetworkStatusObserver.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+        if(!m_networkService->Subscribe<JsonObject>(1000, "onConnectionStatusChanged"
```
varumugam123, line 91: Should we retry to install event handler if there is a timeout?

varumugam123:
@ramasat Could you please review & approve this change?

PR #3882, opened 2023-03-20
Fix worker thread crash after plugin deactivation

nrajan002c:


MFransen69 (approved):

nrajan002c:
There is time gap observed When the NTP time update in the system. This causes false positive 'not operational' return status to Monitoring services and thus plugin is getting deactivated.
To avoid this, we are modifying the clock to 'std::chrono::steady_clock' based .

PR #3894, opened 2023-03-27
RDKDEV-724 serialize DisplaySettings and HdmiCec API invocations

tomasz-karczewski-red:
This PR introduces UtilsSynchro, that are used to serialize API calls and iarm handlers invocation for DisplaySettings and HdmiCec plugins.
Additionally, 2 problems are addressed:
- 'IARM_BUS_PWRMGR_NAME' corrected to 'IARM_BUS_DSMGR_NAME' when removing handler for IARM_BUS_DSMGR_EVENT_AUDIO_FORMAT_UPDATE in DisplaySettings
- when waiting for condition events in HdmiCec, periodically check if the threads should exit already, to prevent possible blocking on join() in CECDisable later on

### DIFF
--- a/DisplaySettings/DisplaySettings.cpp
+++ b/DisplaySettings/DisplaySettings.cpp
@@ -52,6 +52,8 @@
 #include "UtilsisValidInt.h"
 #include "dsRpc.h"
 
+#include "UtilsSynchroIarm.hpp"
+
 using namespace std;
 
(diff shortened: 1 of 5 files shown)


tomasz-karczewski-red:
moved to sprint/23Q2 - https://github.com/rdkcentral/rdkservices/pull/3948

PR #3899, opened 2023-03-28
RDKTV-21933: removing tower-base64 dependancy

zhilarasi:


helpers/UtilsString.h
binuinbaraj, line 1: Can you rename the api as encodeBase64 so its generic.

binuinbaraj (approved):

PR #3918, opened 2023-03-31
Change the clock type to monotonic based

nrajan002c:


emutavchi (changes requested):
Also, could you please elaborate on the issue and the reasoning for this change? Thanks!

nrajan002c:
There is time gap observed When the NTP time update in the system. This causes false positive 'not operational' return status to Monitoring services and thus plugin is getting deactivated.
To avoid this, we are modifying the clock to 'std::chrono::steady_clock' based .

emutavchi (approved):

MFransen69 (approved):

PR #3933, opened 2023-04-04
DELIA-60903: Add Find My Remote method to Remote Control thunder plugin 

egalla204:
trigger "Find my remote" operation from the RemoteControl plugin which will allow users to start a beeping noise in the remote


cbucht200 (changes requested):
The documentation change needs to occur in the RemoteControl.json file, and then regen the docs with the proper tools. Please see documentation section of README

PR #3937, opened 2023-04-05
[ONEM-23925] - add browserversion property to config of WebKitBrowser

jaroslaw-bojko-red:
Reupload


jaroslaw-bojko-red:
@emutavchi @MFransen69 what is the current state of this task? Will it be merged?

emutavchi:
@jaroslaw-bojko-red sorry for the delayed response. Could you please describe the use case for this? Is this supposed to work together with https://github.com/rdkcentral/rdkservices/blob/sprint/23Q2/WebKitBrowser/WebKitBrowser.cpp#L44 or replace it?  it might be better to use the already-defined version instead of introducing a new one that is configured outside of the component. Having multiple versions of a single component could be confusing. 

requesting @anand-ky  and @pwielders to review as well.

MFransen69:
Added @modeveci as reviewer as well.

modeveci:
If you want to fetch the wpe webkit version, there are APIs which bring major,minor,macro versions of wpewebkit:
for example for 2.38.6 version you can get
webkit_get_major_version() -> e.g: 2
webkit_get_minor_version(). -> e.g:38
webkit_get_micro_version(). -> e.g: 6

jacek-skiba-red:
We want to add `browserversion` (eg. 2.22, 2.28, 2.38) to WebkitBrowser.json config. In our usecase other component (UI) is checking `browserversion` property, and based on that value middleware logic is build. 

Change is not related to API version of webkit browser plugin.

MFransen69:
Hi @jaroslaw-bojko-red 

Okay, thanks for letting us know. 
I think storing version information in a file meant to configure a component will be confusing. One might expect it to influence the behavior of the component, not just to be a stored value. From that perspective it breaks the Thunder plugin architecture as the json configuration file should contain key-values that change the configuration of the plugin. One might even think by changing the value one can choose with browser version to load :)
I also agree with Eugene to not store information  that is already available elsewhere, you must make sure they are aligned and correct, so better to have that in one place. 
I don't know if the browser version itself is already exposed via the JSON and/or COM RPC interface but if not, we are happy to add that. Then via these interfaces you could retrieve the correct browser version from an external component. And if you really want you can of course store the information in a different (json) file you define yourself.

jacek-skiba-red:
Thanks @emutavchi and @MFransen69 for details argumentation. We agree with your opinion.

PR #3965, opened 2023-04-12
[RDKDEV-721] XCast limit payload to avoid overflow

dawid-trendota-red:
Fix for

https://jira.rdkcentral.com/jira/browse/RDKDEV-721

### DIFF
--- a/XCast/XCastCommon.h
+++ b/XCast/XCastCommon.h
@@ -24,8 +24,9 @@
 #define __XCAST_COMMON_H__
 /*
  * The maximum DIAL payload accepted per the DIAL 1.6.1 specification.
+ * payload from xdialserver is URI encoded
  */
-#define DIAL_MAX_PAYLOAD (4096)
+#define DIAL_MAX_PAYLOAD (4096*3)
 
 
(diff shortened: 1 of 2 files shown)


pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-3724

XCast/XCastCommon.h
```diff
@@ -24,8 +24,9 @@
 #define __XCAST_COMMON_H__
 /*
  * The maximum DIAL payload accepted per the DIAL 1.6.1 specification.
+ * payload from xdialserver is URI encoded
  */
-#define DIAL_MAX_PAYLOAD (4096)
+#define DIAL_MAX_PAYLOAD (4096*3)
```
apatel859, line 29: max payload allwayed is 4kb why multiplied by 3 which makes it 12kb?

dawid-trendota-red: This payload here is base64 encoded, so it can require up to 3 times more memory then unencoded payload.

pradeeptakdas:
Hi @apatel859 
Can you please review and merge this PR?

PR #3977, opened 2023-04-14
RDK-39845 : Added uploadLogsAsync and abortLogUpload calls.

sborushevsky:


anand-ky:
@sborushevsky - Pls also add the changelog.md change since it's getting cherry-picked from main.

PR #3988, opened 2023-04-20
DELIA-61099:- Monitor: Run Dispatch() under lock

asurdej-comcast:
Dispatch() method of MonitorObjects can be called from different threads and it uses _monitor that is a list of MonitorObject elements. The list itself is not modified (it lives from Open() to Close() based on json config) but MonitorObject elements inside that list can be modified from other threads like StateChange() (e.g.MonitorObject::_source). This is not thread safe and causes a crash when Evaluate() method is interrupted with StateChange (DEACTIVATION state).
Put Dispatch() method under the lock to fix this.


anand-ky (approved):

MFransen69 (commented):
Approved

anand-ky:
@MFransen69 @pwielders - FYI this was a cherry-pick of changes from sprint that was reviewed by Metro. Hence I had merged to main.

MFransen69:
Correct, Pierre didn't immediately recognize that, therefore he added me as reviewer but indeed I recognized I had reviewed it before :)

PR #3999, opened 2023-04-25
RDK-40655 Local Speech Synthesis (Offline TTS) for Flex 2.0 - Nuance

vdinak240:
Reason for change: Feature development
Test Procedure: TBD
Priority: P1
Risks: None


TextToSpeech/impl/TTSSpeaker.cpp
```diff
@@ -222,6 +263,11 @@ bool TTSConfiguration::isValid() {
     return true;
 }
 
+bool TTSConfiguration::isSwitchRequired()
```
varumugam123, line 266: This function name is misleading. From what I understood from the implementation it looks like it is used in the context of "whether the local endpoint is configured", renaming it to "hasValidLocalEndpoint()" or similar would be more readable

TextToSpeech/impl/TTSSpeaker.cpp
```diff
@@ -222,6 +263,11 @@ bool TTSConfiguration::isValid() {
...
+{
+    return m_switchRequired;
```
varumugam123, line 268: If it is used to check the local endpoint configured, this boolean is redundant

TextToSpeech/impl/TTSSpeaker.cpp
```diff
@@ -350,6 +402,58 @@ int TTSSpeaker::speak(TTSSpeakerClient *client, uint32_t id, std::string text, b
...
+{
+    g_object_set(G_OBJECT(m_source), "location", url.c_str(), NULL);
+    // PCM Sink seems to be accepting volume change before PLAYING state
+    g_object_set(G_OBJECT(m_audioVolume), "volume", (double) (data.client->configuration()->volume() / MAX_VOLUME), NULL);
+    gst_element_set_state(m_pipeline, GST_STATE_PLAYING);
+    #if defined(PLATFORM_AMLOGIC) || defined(PLATFORM_REALTEK)
+    setMixGain(MIXGAIN_PRIM,data.primVolDuck);
+    #endif
+    TTSLOG_VERBOSE("Speaking.... ( %d, \"%s\")", data.id, data.text.c_str());
+    //Wait for EOS with a timeout incase EOS never comes
```
varumugam123, line 415: Having a new line before & after gst_element_set_state(), before the commit //Wait for EOS would be more readable

TextToSpeech/impl/TTSSpeaker.cpp
```diff
@@ -859,38 +981,38 @@ std::string TTSSpeaker::constructURL(TTSConfiguration &config, SpeechData &d) {
     std::string tts_request;
     TTSURLConstructer url;
     tts_request = url.constructURL(m_defaultConfig,d.text,false);
```
varumugam123, line 983: here the variable `url` is little misleading, 'urlConstructor' would be more suitable also we shall stick to the uniform naming conventions

TextToSpeech/impl/TTSSpeaker.cpp
```diff
@@ -643,20 +747,37 @@ void TTSSpeaker::createPipeline() {
...
+           } 
+           else
+           {
+               TTSLOG_INFO("mp3 audio playback is enabled");
+               m_pipelinetype = MP3;
+           }   
+        }   
+        else
+        {
+            if((getUrlType(tts_url)) == PCM){
```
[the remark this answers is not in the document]
varumugam123: Setting pipelinetype & pcmaudioenabled shouldn't depend on local endpoint configuration

TextToSpeech/impl/TTSURLConstructer.cpp
```diff
@@ -23,106 +23,171 @@ TTSURLConstructer::~TTSURLConstructer()
...
+
+std::string TTSURLConstructer::constructURL(TTSConfiguration &config,std::string text,bool isfallback)
+{
+     // EndPoint URL
+    //std::string tts_request;
+    if(!(config.apiKey().empty()))
+    {
+       if(config.isSwitchRequired())
+       {
+          if(WPEFramework::Plugin::TTS::NetworkStatusObserver::getNetworkStatusObserver()->isConnected())
```
varumugam123, line 146: Detecting network state & determining whether the local endpoint should be used, shouldn't be here. It should be with TTSSpeaker. This way, this TTSURLConstructor's scope will just be constructing the URL based on configuration passed. We can determine which URL should be used from TTSSpeaker & pass it to constructor (this would also help to consider the m_remoteError case)

bool TTSSpeaker::shouldUseLocalEndpoint() {
    if(hasValidLocalEndpoint())
        return !NetworkObserver::getInstance()->isConnected() || m_remoteError;
    return false;
}

// From wherever we need the URL constructed
urlConstructor.constructURL(config, text, isFallback, shouldUseLocalEndpoint());

TextToSpeech/impl/TTSURLConstructer.cpp
```diff
@@ -23,106 +23,171 @@ TTSURLConstructer::~TTSURLConstructer()
...
+    }
+    else
+    {
+       if(config.isSwitchRequired())
+       {
+          
+          if(WPEFramework::Plugin::TTS::NetworkStatusObserver::getNetworkStatusObserver()->isConnected())
+          {
+              TTSLOG_INFO("Device online using remote endpoint\n");
+              return httpgetURL(true,config,text,isfallback);
```
varumugam123, line 172: camel case for function names

TextToSpeech/impl/TTSSpeaker.cpp
```diff
@@ -960,7 +1082,12 @@ void TTSSpeaker::GStreamerThreadFunc(void *ctx) {
         if(!speaker->m_flushed) {
             speaker->speakText(*data.client->configuration(), data);
         }
-
+        
+        if(!speaker->m_flushed && speaker->m_remoteError)
+        {
+            TTSLOG_INFO("speak with local endpoint\n");
+            speaker->speakLocal(data);
+        }
```
varumugam123, line 1090: Please have a new line after this section

TextToSpeech/impl/TTSSpeaker.cpp
```diff
@@ -960,7 +1082,12 @@ void TTSSpeaker::GStreamerThreadFunc(void *ctx) {
         if(!speaker->m_flushed) {
             speaker->speakText(*data.client->configuration(), data);
         }
-
+        
+        if(!speaker->m_flushed && speaker->m_remoteError)
+        {
```
varumugam123, line 1087: So, my understanding is we fall back to local endpoint on the very first network failure observed... So even the spurious failures 500 responses we have seen would trigger using the local endpoint configured

varumugam123 (changes requested):

anand-ky:

PR #4045, opened 2023-05-16
WebKitBrowser: Allow setting arbitrary environment variables

rjpdasilva:
Add a new WebKitBrowser configuration option to allow setting arbitrary environment variables in a generic way.

Sometimes, the only way to configure lower layer components used by the WebKitBrowser (e.g. some specific GStreamer settings) is through environment variables. There are several components using this approach, and each may have several environment variables. This way, instead of opening up configuration options for each of the features associated with those environment variables, it is preferable to just add a way to generically set environment variables through the configuration, which is the goal of this change.

For this, introduce a new "environment" option to the JSON configuration. The contents for the new option is a semi-colon (';') separated list of environment variables setting like 'ENV_VAR1=value1;ENV_VAR2=value2'.

The new "environment" JSON configuration can be set in configure time using the new associated cmake variables:

 * PLUGIN_WEBKITBROWSER_ENVIRONMENT: For the generic WebKitBrowser instance.

 * PLUGIN_AMAZON_ENVIRONMENT: For the Amazon specific instance.

 * PLUGIN_APPS_ENVIRONMENT: For the Apps specific instance.

 * PLUGIN_HTML_APP_ENVIRONMENT: For the HtmlApp specific instance.

 * PLUGIN_JSPP_ENVIRONMENT: For the JSPP specific instance.

 * PLUGIN_LIGHTNING_APP_ENVIRONMENT : For the LightningApp specific instance.

 * PLUGIN_RESIDENT_APP_ENVIRONMENT: For the ResidentApp specific instance.

 * PLUGIN_SEARCH_AND_DISCOVERY_APP_ENVIRONMENT: For the SearchAndDiscoveryApp specific instance.

 * PLUGIN_UX_ENVIRONMENT: For the Apps specific instance.

 * PLUGIN_YOUTUBE_ENVIRONMENT: For the YouTube specific instance.

Difference cmake variables are created for the above instances so that it is possible to configure them differently, and not have the other instances "inherit" from the generic instance.

The processing of the new "environment" configuration option is done by the new SetConfigEnvironment() method, which honors the already exiting 'environmentOverride' boolean, at the end of the configuration process done by Configure(). It basically uses a tokenizer to get each environment variable assignment and then for each, splits the assignment into variable name/value and finally uses the framework's Core::SystemInfo::SetEnvironment() to actually set the requested environment variables to the requested values. It also does the required validations to ensure a valid environment variable name and value is parsed.

Also adjust the JSON specification for the WebKitBrowser configuration to describe the new "environment" configuration option.

Ticket: RDKDEV-755


anand-ky:
Closing older PRs. Pls bring it to latest sprint/main branches if required.

rjpdasilva:
Hi @anand-ky,

> Closing older PRs. Pls bring it to latest sprint/main branches if required.

Done in #5421.

PR #4055, opened 2023-05-18
RDKTV-22120 : Added key regeneration if loading of the key or cipher fails.

sborushevsky:


ContinueWatching/ContinueWatching.cpp
```diff
@@ -369,11 +370,46 @@ namespace WPEFramework {
...
+				}
 
 				result = SecCipher_GetInstance(sec_proc, algorithm, mode, sec_key, iv, &sec_cipher);
 				if(result != SEC_RESULT_SUCCESS)
-					throw "Failure to get SecAPI cipher handler!";
+				{
+
+					char buf[64];
+					snprintf(buf, sizeof(buf), "%016llX.*", mSecObjectId);
+					std::string cmd = "rm /opt/drm/servicemgr/";
```
anand-ky, line 394: Checking why do we need to delete this folder? Is this used only for ContinueWatching?

anand-ky (approved):

PR #4065, opened 2023-05-22
RDK-41544: Startup service configuration changes to support Flex 2.0 …

PaviP98:
…activation workflow


tdeva14 (dismissed):

tdeva14 (dismissed):
LGTM

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -356,6 +395,126 @@ namespace WPEFramework {
...
+                    if (thunder_client == nullptr) {
+                        LOGINFO("Failed to get plugin handle");
+                        sleep(5);
+                    } else {
+                        JsonObject params;
+                        JsonObject joGetResult;
+
+                        params["clientId"] = "gdi";
+                        params["distributedTraceType"] = "money";
+                        params["distributedTraceId"] = "trace-id=8e44ec3c-7aa1-404d-a1e5-1dde72114ca3;parent-id=123;span-id=456";
```
anand-ky, line 421: What are these values "clientId", "distributedTraceType", "distributedTraceId"? Is it ok to put it here since it's open source?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -356,6 +395,126 @@ namespace WPEFramework {
...
+                            LOGINFO("getDeviceInitializationContext failed. Retrying in %d seconds", retryDelay);
+                            sleep(retryDelay);
+                        }
+                    }
+                } else {
+                    LOGINFO("%s is not active, Retrying in 5 seconds", secMgr_callsign);
+                    sleep(5);
+                }
+            } while (!success);
+            return success;
```
anand-ky, line 471: This looks like indefinite loop. Checking why do we need it? Can you pls check and cap the number of retries? Why doesn't getDeviceInitializationContext return the correct context the first time? Is it dependent on something else in SecManager getting initialized?

anand-ky (changes requested):

MaintenanceManager/MaintenanceManager.cpp
mvalas827: Can we have this either as a dependency at plugin level or as an event based trigger for Unsolicited maintenance?.

MaintenanceManager/MaintenanceManager.cpp
mvalas827, line 421: Can we manage these part of plugin configuration. Hardcoding is not a preferred way.

tdeva14:
LGTM

anand-ky (approved):
giving conditional approval. These are temporary changes to unlock development. The indefinite retry for getDeviceInitializationContext should be replaced by event from SecManager notifying when device initialization context is available from WhoAmI service.

PR #4066, opened 2023-05-23
[RDK-40623] Unit Test for System Plugin

sethulakshmisk:
Implemented a test API for fopen() Error case in SystemServices.
For this,added a mock method to simulate a failure scenario for fopen. 
Note:The Mock class also includes the default behavior for fopen.


Tests/mocks/Wraps.h
```diff
@@ -10,6 +10,7 @@ class WrapsImpl {
     virtual FILE* popen(const char* command, const char* type) = 0;
     virtual int pclose(FILE *pipe) = 0;
     virtual void syslog(int pri, const char* fmt, va_list args) = 0;
+	virtual FILE* fopen(const char* filename, const char* mode) = 0;
```
anand-ky, line 13: Pls fix the indent in many places. Looks like it's taken 4 more spaces or a extra tab. There are other places in the PR that needs to be fixed

sethulakshmisk: Thanks for the feedback.
The alignment issues in all the files have been resolved now.

PR #4070, opened 2023-05-24
Fix mixed content

jacek-manko-red:
"enable-websecurity" property does not exist on wpe 2.38. It has been renamed to "disable-web-security", with reversed meaning.

As a result, the entire g_object_set call fails and the other properties are not set.

"disable-web-security" disables the same-origin policy, which is separate from mixed content and not required here.


jacek-manko-red:
Please let me know if the pull request should be created for sprint/23Q2 branch instead

emutavchi (changes requested):
Changes to main branch should go through the verification and formal approval process. It will be done as part of wpe-2.38 upgrade which is ongoing effort.

Also, the proposed change will break builds with wpe-2.28. 23Q3 sprint code the needed changes for wpe-2.38. Could you please use that branch?

anand-ky:

PR #4087, opened 2023-06-01
RDK-41544: Startup service configuration changes to support Flex 2.0 activation

PaviP98:


anand-ky (approved):
giving conditional approval. These are temporary changes. The indefinite retry for getDeviceInitializationContext should be replaced by event from SecManager notifying when device initialization context is available from WhoAmI service ([RDK-41666](https://ccp.sys.comcast.net/browse/RDK-41666))

PR #4098, opened 2023-06-05
RDK-40634, RDK-41635: WebRTC on WPE-2.38

varumugam123:
Reason for change: With latest wpe, WebRTC is a runtime feature that needs to be explicitely enabled through API
Test Procedure: Refer ticket
Priority: P1
Risks: None


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2833,6 +2833,7 @@ static GSourceFuncs _handlerIntervention =
 
             auto* preferences = webkit_settings_new();
 
+            webkit_settings_set_enable_webrtc(preferences, TRUE);
```
varumugam123, line 2836: Need a check on WPE version to be able to build with older versions

emutavchi (approved):

modeveci (approved):

PR #4128, opened 2023-06-15
RDKTV-22121 CEC Report Features Thunder

kprasad96:
Added support for <Give Features> and <Report Features> CEC message
Also added API's to set and get CEC Version from RFC


kprasad96:
Bringing in review comments as another change

PR #4160, opened 2023-06-27
Improve memory pressure settings to account for GPU memory

Scony:
This PR supplements the system memory pressure settings which are already present.


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2754,12 +2757,15 @@ static GSourceFuncs _handlerIntervention =
                 }
```
emutavchi, line 2758: I believe we don't want to change pressure settings if Memory config is not set. Could you please revert back to set pressure setting only if config is set?

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2773,17 +2779,44 @@ static GSourceFuncs _handlerIntervention =
...
-                    webkit_memory_pressure_settings_free(memoryPressureSettings);
-                } else
-#endif
-                {
-                    wkContext = webkit_web_context_new_with_website_data_manager(websiteDataManager);
+                    webkit_memory_pressure_settings_set_memory_limit(webProcessMemoryPressureSettings, _config.Memory.WebProcessLimit.Value());
                 }
+
+                const char* webProcessGPULimitEnvvar;
+                if ((_config.Memory.IsSet() == true) && (_config.Memory.WebProcessGPULimit.IsSet() == true)) {
```
emutavchi, line 2790: Same here

emutavchi (changes requested):

WebKitBrowser/WebKitImplementation.cpp
emutavchi: i.e. change settings when either Memory config  or GPU pressure env variable are set

emutavchi (approved):

modeveci:
I have seen a few points need to be discussed in this PR:

1. If I am not wrong, GPU memory settings are introduced by this PR. 
WPE_POLL_MAX_MEMORY_GPU is supported until 2.28. From 2.38 on, it has webkit_memory_pressure_settings_set_video_memory_limit API instead of env variable. If we want to have this change for 2.28 as well, then why it is implemented to extract from environment variable in HAS_MEMORY_PRESSURE_SETTINGS_API block instead of following same logic here https://github.com/rdkcentral/rdkservices/blob/sprint/23Q2/WebKitBrowser/WebKitImplementation.cpp#L2199C6-L2199C38

2. Isnt it good to introduce a config option to set GPU file? WPE_POLL_MAX_MEMORY_GPU_FILE: WPE_POLL_MAX_MEMORY_GPU_FILE is used to hold the path of the file that contains the amount of video memory being used by the system.

3. Poll intervals are hardcoded as 1.0, I can imagine there will be cases to differentiate WebProcess,NetworkProcess, GPU/Video intervals based on use-case to optimise performance degradation. Why not having them separately?

As we are improving configurable APIs for memory settings:

"memory": {
   "webprocess": {
      "limit":
      "pollinterval"
      // Thresholds are also available APIs for memory settings, if we need, we can introduce.
      "thresholds": {
         "kill":
         "strict":
         "conservative"
      }
   },
   "networkprocess": {
       "limit":
      "pollinterval"
      // Thresholds are also available APIs for memory settings, if we need, we can introduce.
      "thresholds": {
         "kill":
         "strict":
         "conservative"
      }
   }
   "video" OR "gpu": {
      "limit":
      "pollinterval"
      "file"
      // Thresholds are also available APIs for memory settings, if we need, we can introduce.
      "thresholds": {
         "kill":
         "strict":
         "conservative"
      }
}

What do you think for these configuration options? We can discuss further during our weekly meeting. 
@emutavchi

Scony:
@modeveci 

1. This PR was meant for `2.38` but I got your point - I'll discuss with @emutavchi 
2. & 3. Sounds like good idea to me

emutavchi:
@modeveci we have configurations where WPE_POLL_MAX_MEMORY_GPU and file env variable are set only for browser instances running inside container (it doesn't make sense to set outside of container, and may actually introduce undesirable results). So, we still need to parse WPE_POLL_MAX_MEMORY_GPU and set it accordingly for 2.38.

As for point 3, I think it would be better to do those improvements in separate PR. and just hardcode poll interval to 1.0 at the moment  (i.e. the default we had with wpe-2.28)

modeveci:
It is ok for point 3. We can create a ticket for this.

We have some questions regarding your comment:
- Is the configuration you mentioned for "container initialization"?
- Is the env. variable set at the start-up of the container?

in the meantime, we will talk with browser team to understand why they left the gpu file as env. variable but gpu limit as API.

@MFransen69

emutavchi:
@modeveci 

> Is the configuration you mentioned for "container initialization"?

yes. it is part of container spec 

> Is the env. variable set at the start-up of the container?

I believe crun sets it. from browser perspective it is set before WPEProcess runs.

modeveci:
@emutavchi , @MFransen69 
I have talked with the team and they are ok for GPU settings to read also from environment variables. And it will have higher priority over "GPU memory settings API". 
So basically, WPE_POLL_MAX_MEMORY_GPU and WPE_POLL_MAX_MEMORY_GPU_FILE will be read by webkit to apply settings, we dont need to parse here to pass any value. 

we will create a ticket for point 3. I think we can close this ticket.

Scony:
@modeveci so to sum up - the changes in this PR are not required anymore?

MFransen69:
Hi @Scony 
Ozgur is on vacation, back next Monday. But indeed that is what he explained me, Webkit will bring back the functionality to read these environment variables so it does not have to be done in the plugin, that is why he proposes to close this pr. But he can fill in the details next week.

Scony:
@modeveci so to sum up - the changes in this PR are not required anymore?

PR #4163, opened 2023-06-27
RDK-42281:Pass SAT Token in the authorization header of the request, …

vdinak240:
…when interacting with the VREX TTS 2.0

Reason for change: Changes for VREX 2.0
Test Procedure: Mentioned in ticket
Risks: Low


ramasat (approved):

TextToSpeech/impl/SatToken.h
```diff
@@ -0,0 +1,37 @@
...
+    SatToken(const string callsign);
+    SatToken(const  SatToken&) = delete;
+    SatToken& operator=(const  SatToken&) = delete;
+
+    string getSecurityToken();
+    void serviceAccessTokenChangedEventHandler (const JsonObject& parameters);
+    void getServiceAccessToken();
+
+    WPEFramework::JSONRPC::LinkType<WPEFramework::Core::JSON::IElement>* m_authService{nullptr};
+    string m_SatToken;
```
varumugam123, line 29: would m_SATToken suite better?

TextToSpeech/impl/TTSManager.cpp
```diff
@@ -189,6 +189,12 @@ TTS_Error TTSManager::setAPIKey(string apikey)
     return TTS_OK;
 }
 
+TTS_Error TTSManager::setSpeechRate(const string speechrate)
+{
+    m_needsConfigStoreUpdate |= m_defaultConfiguration.setSpeechRate(speechrate);
```
varumugam123, line 194: This is another case, where ITextToSpeech::SetSpeechRate() invocation order depends on ITextToSpeech::SetConfiguration() to properly update the underlying persistent storage (JSON-RPC case works fine due to the management done in that implementation)

PR #4164, opened 2023-06-27
SERXIONE-2602,DELIA-62375: add factoryReset method to Remote Control Plugin

egalla204:
Reason for change:  Add factoryReset method to Remote Control Plugin which
sends a factory reset request to all paired and connected remotes
 And also increase the IARM timeout for all IRDB calls to 10 seconds
since we need to query a cloud IRDB service that could take longer than
the default 5 second IARM timeout
Test Procedure:     see tickets
Risks:              Low


cbucht200 (commented):
Looks fine to me.

PR #4171, opened 2023-06-30
RDK-42641 Added Report Features CEC message

kprasad96:
Added support for Report Features CEC message
 Report Features message will be broadcasted if CEC Version is 2.0


kprasad96:
Will bring in as Single commit

PR #4172, opened 2023-06-30
SERXIONE-2602,DELIA-62375: add factoryReset method to Remote Control Plugin

egalla204:
Reason for change:  Add facotryReset method to Remote Control Plugin,
and also increase the IARM timeout for all IRDB calls to 10 seconds
since we need to query a cloud IRDB service that could take longer than
the default 5 second IARM timeout
Test Procedure:     see tickets
Risks:              Low


RemoteControl/RemoteControl.cpp
```diff
@@ -35,6 +35,10 @@
 
 #define RF4CE_IEEE_MAC_ADDRESS_STR_MAX    24
 
+#define IARM_FACTORY_RESET_TIMEOUT  (15 * 1000)  // 15 seconds, in milliseconds
+#define IARM_IRDB_CALLS_TIMEOUT     (10 * 1000)  // 10 seconds, in milliseconds
+
+
```
anand-ky, line 41: Pls update minor patch version to 3 to match the CHANGELOG. This would then show up in Controller.status API for RemoteControl. 

#define API_VERSION_NUMBER_MINOR 0

anand-ky (changes requested):

anand-ky (approved):

PR #4227, opened 2023-07-17
Dev ut 23 q3

viveksinghnarwaria:
RDK-42471 : Add MaintenanceManager and Packer Unit Testcases

    Test Procedure: Refer the ticket for test procedure
    Risks: Low
    Priority: P1
    Signed-off-by: viveksingh_narwaria@comcast.com


.github/workflows/unit-tests.yml
```diff
@@ -309,6 +315,8 @@ jobs:
           /etc/dcm.properties          
           /etc/authService.conf
           /version.txt
+          /lib/rdk/getMaintenanceStartTime.sh
```
apatel859, line 318: Did you created this new script for this work?  AS per RDK-E guideline we are removing the usage of script from cpp code.

ramasat (approved):

.github/workflows/unit-tests.yml
viveksinghnarwaria: This is a mock script to read MM start time. These changes are giving permission to read and write file in docker. Maintenance manager UT will create this mock file and maintenance manager will read it. Whenever we change maintenance manager for RDK-E. The testcase needs to be updated for new design as per requirements.

apatel859 (approved):

PR #4247, opened 2023-07-25
RDKTV-24212: Maintenance Manager subscribes for network event and

PaviP98:
triggers critical tasks like DCM and xconf once device get connected to internet


tdeva14 (approved):
LGTM

apatel859 (approved):

PR #4249, opened 2023-07-25
PerformanceMetrics plugin update

nrajan002c:
The following changes are done in this PR,

1. Existing coldLaunch logic doesn't work. Hence modified it.
2. Want to get the hostname only, the complete url is having lot of special character which doesn't suit for Telemetry.
3. Default Load average digits are too high. So I have reduced the number of digits.


PerformanceMetrics/PerformanceMetrics.h
```diff
@@ -75,6 +75,14 @@ namespace Plugin {
 
                 _parent.PluginDeactivated(*service);
             }
+            void Activation (const string&, PluginHost::IShell* service) override
```
MFransen69, line 78: As discussed on slack I now understand why these lines are needed, it is because of this patch file in the RDKServices build: https://gerrit.teamccp.com/plugins/gitiles/rdk/yocto_oe/layers/meta-rdk-video/+/23Q3_sprint/recipes-extended/wpe-framework/wpeframework/r4/activationevent.patch

However we cannot just merge this lines at that will break build(systems) that do not include the patch.
I will have a discussion on the need for this patch (if merging would be really necessary in the meantime may I propose to just as a workaround for now remove the override keyword and also not make them virtual, then it will build with and without the patch (and as downside only add a very small piece of unneeded code that perhaps the optimizer throws away anyway)

PerformanceMetrics/PerformanceMetrics.h
```diff
@@ -116,7 +124,7 @@ namespace Plugin {
         public:
             static constexpr char startURL[] = _T("about:blank");
 
-            virtual void LoadFinished(const string& URL, const int32_t httpstatus, const bool success, const uint32_t totalsuccess, const uint32_t totalfailed) = 0;
+            virtual void LoadFinished(const string& URL, const int32_t httpstatus, const bool success, const uint32_t totalsuccess, const uint32_t totalfailed, bool coldLaunch=false) = 0;
```
MFransen69, line 127: This is the interface that is used by the monitoring engine part to communicate with the different reporting engines (systlog, tracing and perhaps in the future there might be more). By changing the interface and not adapting the tracing implementation that is now broken.
But the whole idea is to have the engine part now knowing about the details of what needs to be reported. The whole concept of a cold start is rather special and not all reporting engines might use that concept, therefore up to know it was only available in the syslog reporting engine. If there is a reason to change the coldstart reporting of course feel free to do so (it was implemented exactly as in the example and verified by QA) but please do not do it in the monitoring engine, that should not know about  these details, so please do so in the SyslogOutput.cpp so that other reporting engines are not forced to know about this detail and we keep the interface clean. I think you could just do it there as well but if not all info you need is available there reach out and we can see if we need to change the interface.

MFransen69 (changes requested):

PerformanceMetrics/CMakeLists.txt
```diff
@@ -28,12 +28,12 @@ set(MODULE_NAME ${NAMESPACE}${PROJECT_NAME})
 message("Setup ${MODULE_NAME} v${PROJECT_VERSION}")
 
 set(PLUGIN_PERFORMANCEMETRICS_AUTOSTART "true" CACHE STRING "Automatically start Performance Metrics plugin")
-set(PLUGIN_PERFORMANCEMETRICS_LOGGER_IMPLEMENTATION "TRACING" CACHE STRING "Defines what implementation to use for logging the Performance Metrics")
+set(PLUGIN_PERFORMANCEMETRICS_LOGGER_IMPLEMENTATION "SYSLOG" CACHE STRING "Defines what implementation to use for logging the Performance Metrics")
```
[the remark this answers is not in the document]
nrajan002c: Yes. I understood the build flag. I changed 'SYSLOG' to 'TRACING'.

PerformanceMetrics/CMakeLists.txt
```diff
@@ -28,12 +28,12 @@ set(MODULE_NAME ${NAMESPACE}${PROJECT_NAME})
...
 
 # Plugins built from this repository that can be autmatically enabled or enabled manually when built externally
 set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER "${PLUGIN_WEBKITBROWSER}" CACHE BOOL "Enable monitor for the Performance Metrics plugin")
 set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_APPS "${PLUGIN_WEBKITBROWSER_APPS}" CACHE BOOL "Enable monitor for the Performance Metrics Apps plugin")
-set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_RESIDENT_APP "${PLUGIN_WEBKITBROWSER_RESIDENT_APP}" CACHE BOOL "Enable monitor for the Performance Metrics Resident App plugin")
+#set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_RESIDENT_APP "${PLUGIN_WEBKITBROWSER_RESIDENT_APP}" CACHE BOOL "Enable monitor for the Performance Metrics Resident App plugin")
```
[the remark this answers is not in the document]
nrajan002c: PLUGIN_WEBKITBROWSER_RESIDENT_APP is global flag which is used in webkitbrowser plugin as well.
We will not be able to set this flag to OFF. This will disable the ResidentApp plugin. 
Need to look for other options for this.

PerformanceMetrics/PerfMetricsWebKitBrowser.config
```diff
@@ -1,7 +1,7 @@
 set(autostart ${PLUGIN_PERFORMANCEMETRICS_AUTOSTART})
 
 map()
-    kv(callsign "WebKitBrowser")
+    kv(classname "WebKitBrowser")
```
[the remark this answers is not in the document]
nrajan002c: Sorry, I couldn't understand this suggestion. I will reach out you for clarification.

PerformanceMetrics/PerformanceMetrics.h
nrajan002c: Removed the new override apis as discussed.

PerformanceMetrics/PerformanceMetrics.h
nrajan002c: Moved the cold launch logic internal to Syslogoutput.cpp.

PerformanceMetrics/CMakeLists.txt
```diff
@@ -90,4 +91,6 @@ if(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_YOUTUBE)
     write_config( PLUGINS PerfMetricsYouTube )
 endif()
 
-
+if (PLUGIN_PERFORMANCEMETRICS_RDKBROWSER)
```
MFransen69, line 94: By adding a copy of write_config( PLUGINS PerfMetricsWebKitBrowser ) you now  write the config file twice. As mentioned PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER must be true otherwise none of the monitoring would make sense, and then if you select (PLUGIN_PERFORMANCEMETRICS_RDKBROWSER you would write PerfMetricsWebKitBrowser here as wel as earlier in the CMake file. It is better to do that only once and use some logic in the if clause (although I think you can remove the new one and leave the below one as is, if PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER is not true it also has no use monitoring on classname as the plugin will not be in the image, so if it is true you always want to generate the config, only the content will be dependent on the classname flag)

PerformanceMetrics/PerfMetricsWebKitBrowser.config
```diff
@@ -1,7 +1,11 @@
 set(autostart ${PLUGIN_PERFORMANCEMETRICS_AUTOSTART})
 
 map()
-    kv(callsign "WebKitBrowser")
+    if( NOT "${PLUGIN_PERFORMANCEMETRICS_RDKBROWSER}" STREQUAL "")
```
MFransen69, line 4: As mentioned above I would make it boolean flag as it can only have two values and the just hardcode the key/value pair

MFransen69 (changes requested):

PerformanceMetrics/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_APPS "${PLUGIN_WEBKITBROWSER_APPS}"
 set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_RESIDENT_APP "${PLUGIN_WEBKITBROWSER_RESIDENT_APP}" CACHE BOOL "Enable monitor for the Performance Metrics Resident App plugin")
 set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_UX "${PLUGIN_WEBKITBROWSER_UX}" CACHE BOOL "Enable monitor for the Performance Metrics UX plugin")
 set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_YOUTUBE "${PLUGIN_WEBKITBROWSER_YOUTUBE}" CACHE BOOL "Enable monitor for the Performance Metrics Youtube plugin")
+set(PLUGIN_PERFORMANCEMETRICS_RDKBROWSER "classname" CACHE STRING "Enable monitor for the Performance Metrics plugin in RDK")
```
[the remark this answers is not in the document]
nrajan002c: Yes. I renamed it to PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_CLASSNAME as suggested.

PerformanceMetrics/CMakeLists.txt
nrajan002c: I removed the duplicate codes. And used OR for PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_CLASSNAME.

PerformanceMetrics/PerfMetricsWebKitBrowser.config
nrajan002c: I set the key based on the boolean flag

PerformanceMetrics/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_APPS "${PLUGIN_WEBKITBROWSER_APPS}"
 set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_RESIDENT_APP "${PLUGIN_WEBKITBROWSER_RESIDENT_APP}" CACHE BOOL "Enable monitor for the Performance Metrics Resident App plugin")
 set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_UX "${PLUGIN_WEBKITBROWSER_UX}" CACHE BOOL "Enable monitor for the Performance Metrics UX plugin")
 set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_YOUTUBE "${PLUGIN_WEBKITBROWSER_YOUTUBE}" CACHE BOOL "Enable monitor for the Performance Metrics Youtube plugin")
+set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_CLASSNAME "${PLUGIN_WEBKITBROWSER_CLASSNAME}" CACHE BOOL "Enable monitor for the Performance Metrics plugin by classname")
```
MFransen69, line 39: We do not need to set the default value with another buildflag here (above that was handy to turn on for example monitoring youtube automatically when youtube webkit plugin was enabled).
So I propose to change this to:

set(PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_CLASSNAME OFF CACHE BOOL "Enable monitor for the Performance Metrics plugin by classname")

Now the default is off (not to break the old usage) and PLUGIN_PERFORMANCEMETRICS_WEBKITBROWSER_CLASSNAME  can be set to ON in a build that wants to monitor webkit on classname.

MFransen69 (changes requested):

PerformanceMetrics/CMakeLists.txt
nrajan002c: Done.

MFransen69 (approved):

emutavchi (approved):

PR #4264, opened 2023-08-01
RDK-42993 - [RDKE]

sethulakshmisk:
RDKServices: Avoid using script/system commands using system functions RDCentral - Helpers - Replace System


anand-ky:
Not required for main for now.

PR #4265, opened 2023-08-01
RDK-42993 - [RDKE]

sethulakshmisk:
RDKServices: Avoid using script/system commands using system functions RDCentral - Helpers - Replace System


anand-ky:
Closing this PR since it's on sprint. We will replace it currently only on RDK-E branch.

PR #4274, opened 2023-08-02
Topic/simplification rdkvs

KaurSup:
This is a subtask of RDK 42211

Reason for change: Trying to make RDK-E lean, as part of this gerrit, updating the popen calls in Rdkservices, Network plugin
Test Procedure: Basic sanity and tested using curl commands as specified here : https://github.com/rdkcentral/rdkservices/tree/sprint/23Q3/Network#readme
Risks: Medium.
Priority: P0

### DIFF
--- a/Network/NetUtils.cpp
+++ b/Network/NetUtils.cpp
@@ -28,7 +28,6 @@
 //Defines
 
 #define NETUTIL_DEVICE_PROPERTIES_FILE          "/etc/device.properties"
-#define COMMAND_RESULT_FILE "cmdresult"
 
 
 namespace WPEFramework {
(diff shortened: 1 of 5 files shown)


Network/NetworkTraceroute.cpp
```diff
@@ -101,21 +100,41 @@ namespace WPEFramework {
...
+                   pipe = v_secure_popen("r", "traceroute -w %d -m %d -q %d %s %d 2>&1",
+                           wait,
+                           maxHops,
+                           packets,
+                           endpoint.c_str(),
+                           DEFAULT_IPV4_PACKET_LENGTH);
                 }
 
-                if (NetUtils::execCmd(command, output, &result) < 0)
+                if (pipe == NULL)
```
anand-ky, line 111: Do we need to remove NetUtils::execCmd? We can just change the popen call within this wrapper API to v_secure_popen so that other modules within Network can continue to use the helper function execCmd.

Network/NetUtils.cpp
```diff
@@ -79,91 +78,6 @@ namespace WPEFramework {
...
-
-            // If we have requested a result, store the result of the command in /tmp/cmdresult
-            if (result)
-            {
-                getTmpFilename(COMMAND_RESULT_FILE, resultFile); //append a count to get a unique filename
-                commandString += "; echo $? > ";
-                commandString += resultFile;
-            }
-
-            pipe = popen(commandString.c_str(), "r");
```
anand-ky, line 120: Do we need to remove NetUtils::execCmd? We can just change the popen call within this wrapper API to v_secure_popen so that other modules within Network can continue to use the helper function execCmd.

anand-ky (changes requested):

Network/NetworkTraceroute.cpp
KaurSup: @anand-ky, thanks for your comment.
v_secure_popen() does not accept the command argument as string. So 
pipe = popen(commandString.c_str(), "r"); 
can not be replaced as 
pipe = v_secure_popen("r", commandString.c_str());
Also the command must be a static string (with formatting), … cannot be a variable string.
I'm attaching and email from David,Jones explaining this :

image

anand-ky: sounds good, thanks.

anand-ky (approved):

KaurSup:
@anand-ky I did not get any compilation or linking error, so I did not include it. I have tested with the curl commands and the test report is attached in the Jira

PR #4334, opened 2023-08-15
[webkitbrowser][memory_pressure]: Update memory pressure configurations

modeveci:


WebKitBrowser/CMakeLists.txt
```diff
@@ -61,8 +61,12 @@ set(PLUGIN_WEBKITBROWSER_TRANSPARENT "false" CACHE STRING "Set transparency")
 set(PLUGIN_WEBKITBROWSER_STARTURL "about:blank" CACHE STRING "Initial URL for WebKitBrowser plugin")
 set(PLUGIN_WEBKITBROWSER_USERAGENT "" CACHE STRING "User agent string")
 set(PLUGIN_WEBKITBROWSER_MEMORYPROFILE "512m" CACHE STRING "Memory Profile")
-set(PLUGIN_WEBKITBROWSER_MEMORYPRESSURE_WEBPROCESSLIMIT "300" CACHE STRING "Memory Pressure Webprocess Limit")
-set(PLUGIN_WEBKITBROWSER_MEMORYPRESSURE_NETWORKPROCESSLIMIT "100" CACHE STRING "Memory Pressure Networkprocess Limit")
+set(PLUGIN_WEBKITBROWSER_MEMORYPRESSURE_WEBPROCESS_SETTINGS_LIMIT "300" CACHE STRING "WebkitBrower Memory Pressure Webprocess Limit")
+set(PLUGIN_WEBKITBROWSER_MEMORYPRESSURE_WEBPROCESS_SETTINGS_GPU_LIMIT "1000" CACHE STRING "WebkitBrower Memory Pressure Webprocess GPU Limit")
```
emutavchi, line 65: I'd prefer to keep these GPU-related settings unset to avoid overriding values set in the container configuration. It also makes sense to keep unset since GPU file is not specified.

WebKitBrowser/HtmlApp.config
```diff
@@ -100,12 +100,36 @@ ans(configuration)
...
-    kv(webprocesslimit ${PLUGIN_HTML_APP_MEMORYPRESSURE_WEBPROCESSLIMIT})
+if(PLUGIN_HTML_APP_MEMORYPRESSURE_WEBPROCESS_SETTINGS_LIMIT OR PLUGIN_HTML_APP_MEMORYPRESSURE_WEBPROCESS_SETTINGS_GPU_LIMIT OR PLUGIN_HTML_APP_MEMORYPRESSURE_WEBPROCESS_SETTINGS_GPU_FILE OR PLUGIN_HTML_APP_MEMORYPRESSURE_WEBPROCESS_SETTINGS_POLLINTERVAL)
+    key(webprocesssettings)
+    map()
+if(PLUGIN_HTML_APP_MEMORYPRESSURE_WEBPROCESS_SETTINGS_LIMIT)
+        kv(limit ${PLUGIN_HTML_APP_MEMORYPRESSURE_WEBPROCESS_SETTINGS_LIMIT})
 endif()
-if(PLUGIN_HTML_APP_MEMORYPRESSURE_NETWORKPROCESSLIMIT)
-    kv(networkprocesslimit ${PLUGIN_HTML_APP_MEMORYPRESSURE_NETWORKPROCESSLIMIT})
+if(PLUGIN_HTML_APP_MEMORYPRESSURE_WEBPROCESS_SETTINGS_GPU_LIMIT)
```
emutavchi, line 109: I believe you need both to be specified.

emutavchi (changes requested):

WebKitBrowser/CMakeLists.txt
modeveci: Environment variables have higher precedence over these settings. So even if these values are set through .json file, env. variables will always take into account with high priority. 
But it is ok to remove default setting of gpu_limit from CMakeLists.

WebKitBrowser/HtmlApp.config
modeveci: Sure, gpufile and gpulimit will be set together

emutavchi (approved):

MFransen69 (approved):

PR #4336, opened 2023-08-16
RDKTV-24365: Add Return Response from AVInput thunder plugin

Neethu379:
Reason for change: Add Return Response from AVInput thunder plugin
Test Procedure: Check curl commads
Risks: None


apatel859 (changes requested):
You have to update the CHANDGELOG file

apatel859 (approved):

PR #4364, opened 2023-08-28
RDKCOM-4226  - URLs are not launching with HTML App, Lighting app on …

dhivyapriyaMurugesan:
…thunder R4.2 rdk-next

Reason for change: after adding of resumed:true in json files able to load the url

Test Procedure: Verified in va platforms

Risks: Low

### DIFF
--- a/WebKitBrowser/HtmlApp.config
+++ b/WebKitBrowser/HtmlApp.config
@@ -1,5 +1,5 @@
 set(autostart ${PLUGIN_HTML_APP_AUTOSTART})
-
+set(resumed true)
 set(preconditions Graphics)
 
 if(PLUGIN_HTML_APP_STARTUPORDER)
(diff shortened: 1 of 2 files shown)


karuna2git (commented):
Good to merge..
Requesting @emutavchi  to review

emutavchi (approved):
@modeveci  could you please review as well?

modeveci:
@MFransen69 , it seems ok to me as it changes default resume as true for HTMLApp and Lightning instances. Could you review as well?

MFransen69:
I approved it as I don't think we use these configs ourselves (correct @modeveci ?) and understand what you are trying to solve. But just to make sure I guess there is some other mechanism in place that makes sure there are not two webkit instances resumed at the same time (e.g. they are not activated at the same time)? In a typical usage a higher level component would be responsible for suspending and resuming instances (or activating/deactivating if required for memory reasons) via json (or com) rpc before URL's are set, to control what instance is active and which not.

MFransen69 (approved):

PR #4370, opened 2023-08-30
SERXIONE-2617: WPEProcess crash with TTS::TTSManager::interrupted

cmurug728:
Reason for change: During Destruction Process of TTSSpeaker m_callback is set to NULL and it accessed by GstremerThreadFuc which leads to crash
Test Procedure: Sanity Testing in apps
Priority: P1
Risks: None

### DIFF
--- a/TextToSpeech/impl/TTSManager.cpp
+++ b/TextToSpeech/impl/TTSManager.cpp
@@ -41,7 +41,8 @@ TTSManager::TTSManager(TTSEventCallback *callback) :
 
 TTSManager::~TTSManager() {
     TTSLOG_TRACE("TTSManager::~TTSManager");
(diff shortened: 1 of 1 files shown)


varumugam123:
Wondering instead of all these changes, if we just set a dummy callback instance to the m_callback the rest of the code can be left untouched. I guess @ https://github.com/rdkcentral/rdkservices/blob/sprint/23Q3/TextToSpeech/impl/TTSManager.h#L61 if you create an object of TTSEventCallback inline and assign its address on destruction crash can be avoided.

Can you give it a try?

varumugam123 (approved):
Change looks good to me. Please get this change merged by @ramasat

anand-ky (approved):

PR #4371, opened 2023-08-30
DELIA-63155 - Thunder Services - Unsafe string Functions - Phase 4

melhar098:
Plugins: HmdiCecSink, HdmiCecSource, HdmiCec_2, HdmiInput

### DIFF
--- a/HdmiCecSink/HdmiCecSink.cpp
+++ b/HdmiCecSink/HdmiCecSink.cpp
@@ -206,7 +206,7 @@ namespace WPEFramework
 
                 in.getBuffer(&buf, &len);
                 for (unsigned int i = 0; i < len; i++) {
-                   sprintf(strBuffer + (i*3) , "%02X ",(uint8_t) *(buf + i));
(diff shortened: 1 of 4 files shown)


HdmiCecSink/HdmiCecSink.cpp
```diff
@@ -1319,11 +1319,11 @@ namespace WPEFramework
 
 				if ( HdmiCecSink::_instance->deviceList[n].m_physicalAddr.getByteValue(0) != 0 )
 				{
-					sprintf(&routeString[length], "%s%d", "HDMI",(HdmiCecSink::_instance->deviceList[n].m_physicalAddr.getByteValue(0) - 1));
+					snprintf(&routeString[length], sizeof(routeString) - length, "%s%d", "HDMI",(HdmiCecSink::_instance->deviceList[n].m_physicalAddr.getByteValue(0) - 1));
```
apatel859, line 1322: length is inited with 0 and never updated in this function. Do we really need length?

melhar098: @apatel859 you are right it is not needed. Removed length.

apatel859 (approved):

PR #4372, opened 2023-08-30
DELIA-63157: Thunder Services - Unsafe string Functions - Phase 5

melhar098:
Convert string functions to safer versions and making sure there is a null terminator at the end after string function calls in RDKServices test files

### DIFF
--- a/Tests/tests/test_SystemServices.cpp
+++ b/Tests/tests/test_SystemServices.cpp
@@ -2542,7 +2542,7 @@ TEST_F(SystemServicesTest, getDeviceInfoSuccess_onValidInput)
                   const char key_estb_mac[] = "12:34:56:78:90:AB";
                   char buffer[1024];
                   memset(buffer, 0, sizeof(buffer));
-                  strcpy(buffer, key_estb_mac);
+                  strncpy(buffer, key_estb_mac, sizeof(buffer) - 1);
                   FILE* pipe = fmemopen(buffer, strlen(buffer), "r");
                   return pipe;
                }));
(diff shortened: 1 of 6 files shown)


Tests/tests/test_SystemServices.cpp
```diff
@@ -3700,7 +3700,7 @@ TEST_F(SystemServicesEventTest, onMacAddressesRetrieved)
             if (valueToReturn != NULL) {
                   char buffer[1024];
                   memset(buffer, 0, sizeof(buffer));
-                  strcpy(buffer, valueToReturn);
+                  strncpy(buffer, valueToReturn, sizeof(buffer) - 1);
```
anand-ky, line 3703: Checking why we are doing sizeof(buffer) - 1. It should be sizeof(buffer). it's already memset to 0. Pls check other strncpy where we have added -1

anand-ky (changes requested):

Tests/tests/test_SystemServices.cpp
melhar098: So for 
`strncpy(buffer, valueToReturn, sizeof(buffer) - 1);`
the reason I do - 1 here is because the last byte in the buffer is already set to null because of the memset call, so I figured we never want to overwrite it so that the buffer is always null terminated no matter what.

The issue with `strncpy(dest, src, n);` 
If n is reached before the end of src, no null terminator is added. So for example
```
string str = "abcdefgh";
char buff[5];
memset(buff, '\0', sizeof(buff));
strncpy(buff, src, sizeof(buff)); 
```
Here the buffer would contain "abcde", not abcd\0.

If there wasn't a memset, then I would have to add the below line of code after to ensure there is always a null terminator
`buff[sizeof(buff) - 1] = '\0';`

anand-ky (approved):

PR #4374, opened 2023-08-30
DELIA-63162: Thunder Services - Unsafe string Functions - Phase 7

melhar098:
Convert string functions to safer versions and making sure there is a null terminator at the end

Plugin: XCast

### DIFF
--- a/XCast/XCast.cpp
+++ b/XCast/XCast.cpp
@@ -458,14 +458,17 @@ bool XCast::getEntryFromAppLaunchParamList (const char* appName, DynamicAppConfi
         for (DynamicAppConfig* regAppLaunchParam : m_appConfigCache) {
             if (0 == strcmp (regAppLaunchParam->appName, appName)) {
                 isEntryFound = true;
-                strcpy (retAppConfig.appName, regAppLaunchParam->appName);
+                strncpy (retAppConfig.appName, regAppLaunchParam->appName, sizeof(retAppConfig.appName));
+                retAppConfig.appName[sizeof(retAppConfig.appName) - 1] = '\0';
 
                 if (regAppLaunchParam->query) {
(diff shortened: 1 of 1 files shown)


XCast/XCast.cpp
```diff
@@ -886,47 +889,47 @@ void XCast::getUrlFromAppLaunchParams (const char *app_name, const char *payload
...
+        strncat( url, "source_type=12", url_len - strlen(url) - 1);
         if(payload != NULL)
         {
             const char * pUrlEncodedParams;
             pUrlEncodedParams = payload;
             if( pUrlEncodedParams ){
-                strcat( url, "&dial=");
-                strcat( url, pUrlEncodedParams );
+                strncat( url, "&dial=", url_len - strlen(url) - 1);
+                strncat( url, pUrlEncodedParams, url_len - strlen(url) - 1);
```
anand-ky, line 926: Checking why we are doing url_len - strlen(url) - 1. It should be url_len - strlen(url). it's already memset to 0. Pls check other strncpy & strncat where we have added -1.

anand-ky (changes requested):

XCast/XCast.cpp
melhar098: For `strncat(dest, src, count);`
This function call does add a null terminator but it is not included in "count". So the function writes count number of characters to the buffer PLUS a null terminator. So there needs to be one spot left in the buffer so the function can append the null terminator after it reaches count.


For reference:
https://en.cppreference.com/w/c/string/byte/strncat#:~:text=The%20behavior%20is%20undefined%20if%20the%20destination%20array%20does%20not%20have%20enough%20space%20for%20the%20contents%20of%20both%20dest%20and%20the%20first%20count%20characters%20of%20src%2C%20plus%20the%20terminating%20null%20character.

anand-ky (approved):

PR #4409, opened 2023-09-12
RDKTV-25173 - adding rawDeviceType parameter in getDiscoveredDevices

DamianoBaroneSky:
Reason for change: AS team are asking if rawDeviceType parameter as appears in  onDiscoveredDevice event
can also be added to getDiscoveredDevices method

Test Procedure: getDiscoveredDevices should return rawDeviceType

Risks: Low
Priority: P1


karuna2git (commented):
can you change the CHANGELOG.MD and the increment the API_VERSION_NUMBER_PATCH by 1.. ?

karuna2git (approved):

PR #4436, opened 2023-09-21
DELIA-63356: [RDK Services]: org.rdk.Wifi.1.getCurrentState API is no…

VenkateshUmmadiSetty:
…t returning the wifi states properly and added persist parameter in wifi setEnabled API

Reason for change: Fixed wifi state cache if wifi interface was enabled and added persist param to persist wifi state after reboot

Risks: Low


karuna2git:
Can you update the documentation also?
There is a WifiManger.md to be updated and also the wifimanager.json to be updated to be at par with code changes.

karuna2git (approved):

PR #4520, opened 2023-10-12
RDK-42088 : Avoid scripts from maintenance manager module

V3110:
Reason for change: Calculate maintenance start time and save to conf file which reads from MM module.
Test Procedure: NA
Risks: None
Priority: P2


tdeva14 (approved):
LGTM

anand-ky (approved):

PR #4524, opened 2023-10-13
SERXIONE-3497: Missing TTS event Logging in TTS Thunder Plugin

cmurug728:
Reason for change: To add Missed TTS event logs
Test Procedure: Sanity Testing for webaps
Risks: None

### DIFF
--- a/TextToSpeech/TextToSpeechImplementation.cpp
+++ b/TextToSpeech/TextToSpeechImplementation.cpp
@@ -406,31 +406,37 @@ namespace Plugin {
 
     void TextToSpeechImplementation::onTTSStateChanged(bool state)
     {
+        TTSLOG_INFO("Notify onttsstatechanged, state: %s", (state ? "true" : "false"));
         dispatchEvent(STATE_CHANGED, JsonValue((bool)state));
     }
 
     void TextToSpeechImplementation::onVoiceChanged(std::string voice)
     {
+        TTSLOG_INFO("Notify onvoicechanged, voice: %s", voice.c_str());
         dispatchEvent(VOICE_CHANGED, JsonValue((std::string)voice));
     }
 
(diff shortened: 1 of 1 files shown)


TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -406,31 +406,37 @@ namespace Plugin {
 
     void TextToSpeechImplementation::onTTSStateChanged(bool state)
     {
+        TTSLOG_INFO("Notify onTTSStateChanged, state: %d", state);
```
varumugam123, line 409: Please ensure the format we print the state & id values (i.e for state should we just print 0/1 or should it be true / false, likewise for id, should it be %d  or %u) are same as previous version.

varumugam123:
@cmurug728 Please let me know if the format specifiers are same as before. If they are, we can proceed to merge this change

TextToSpeech/TextToSpeechImplementation.cpp
cmurug728: @varumugam123 
**For state :** I will change it into true/false

**For id:** https://github.com/rdkcentral/rdkservices/blob/sprint/23Q4/TextToSpeech/TextToSpeechImplementation.cpp#L320 I referred from here,  used %d as format specifier

cmurug728:
> @cmurug728 Please let me know if the format specifiers are same as before. If they are, we can proceed to merge this change



> In earlier version, the logs were done with event names in small letter for e.g: https://github.com/rdkcentral/rdkservices/blob/release/xumo_1/TextToSpeech/TextToSpeech.h#L62
> 
> Should the logging be retained exactly as before?

Done @varumugam123

varumugam123:
Change looks good to me, please get this approved & merged by @ramasat

varumugam123 (approved):

ramasat (approved):

PR #4536, opened 2023-10-17
Event not triggering for Volume level 'zero'

arunl2910:


pradeeptakdas:
Internal ticket created for tracking https://ccp.sys.comcast.net/browse/RDKCOM-4353

anand-ky:
@apatel859,
Pls review.

apatel859 (changes requested):
There is no use of cache_volumelevel variable. current implementation is wrong:

either u can make cache_volumelevel as static then it will make sense

arunl2910:
> There is no use of cache_volumelevel variable. current implementation is wrong:
> 
> either u can make cache_volumelevel as static then it will make sense

@apatel859 as per discussion have addressed the above points, please review the changes

apatel859 (approved):

PR #4576, opened 2023-10-25
DELIA-63302 : RDKServices: In HdmiCecSink plugin osdName is not retur…

dhivyapriyaMurugesan:
…ning proper result returning empty string

Reason for change: Modified the osdName reading value from osdName.toString().c_str()

Test Procedure: Verified and its returning TV Box

Risks: Low


apatel859 (approved):

apatel859 (changes requested):
You have to update change logs as well

PR #4589, opened 2023-10-27
WebKitBrowser: Add simple testing framework

asurdej-comcast:
This change intorduces a simple testing framework that allows a web page to open child pages (in separate WPEWebProcess) and listen to child page results.
It consist of:
1) custom JS APIS (Extension part):
- window.testRunner.runTestCase() that opens test case child page -- tc.addEventListener() listen for events
-- tc.destroy() destro child page
- window.testCase.reportResult() for test case to report its result 2) Native code runs in the main process that helps to create/destroy subviews. Implemented via WebKit user messages IPC

Each sub page is run in separate webkit context that makes it different from generic window.open() but it allows to crash a test case without crashing whole instance.

Hidden under compile and runtime flags. Disabled by default


emutavchi (approved):

modeveci (approved):
@MFransen69 can you also review? Thanks

MFransen69 (approved):

PR #4593, opened 2023-10-30
RDKDEV-860: [RDK Services]: Display Framerate is not persist after re…

VenkateshUmmadiSetty:
…boot using org.rdk.FrameRate.2.setDisplayFrameRate api

Reason for change: Added persist variable for setDisplayFrameRate API to persist display framerate if we pass persist as true other wise display framerate not persist

Risks: Low

### DIFF
--- a/FrameRate/FrameRate.cpp
+++ b/FrameRate/FrameRate.cpp
@@ -51,7 +51,7 @@
 
 #define API_VERSION_NUMBER_MAJOR 1
 #define API_VERSION_NUMBER_MINOR 0
-#define API_VERSION_NUMBER_PATCH 5
+#define API_VERSION_NUMBER_PATCH 6
 
 namespace WPEFramework
 {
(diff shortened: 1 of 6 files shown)


FrameRate/FrameRate.cpp
```diff
@@ -266,13 +266,17 @@ namespace WPEFramework
             LOGINFOMETHOD();
             returnIfParamNotFound(parameters, "framerate");
 
+	    bool hasPersist = parameters.HasLabel("persist");
+	    bool persist = hasPersist ? parameters["persist"].Boolean() : false;
+	    if (!hasPersist) LOGINFO("persist: false");
```
apatel859, line 271: you should print in both true and false case

VenkateshUmmadiSetty: added print in else case for persist as true
if (!hasPersist) LOGINFO("%s persist: false",__FUNCTION__); else LOGINFO("%s persist: true",__FUNCTION__);

apatel859 (approved):

PR #4603, opened 2023-11-01
RDK-37917: TTS- Add Unit Tests for RDKServices (Phase1)

cmurug728:
Reason for change: Add Unit test for TTS
Test Procedure: TBD
Priority: P1
Risks: None

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -8,8 +8,8 @@ on:
 
 env:
   BUILD_TYPE: Debug
-  THUNDER_REF: "R2-v1.14"
-  INTERFACES_REF: "R2-v1.14"
+  THUNDER_REF: "5e7c0b1ed3c3dd0fc31c86518a364388dc24273b"
+  INTERFACES_REF: "669d9c6e5ed7a5938ff26e1e7736adf485c7a205"
 
 
 jobs:
@@ -75,6 +75,11 @@ jobs:
           sudo apt update
           &&
           sudo apt install -y libsqlite3-dev libcurl4-openssl-dev valgrind lcov clang libsystemd-dev
+
+      - name: Install GStreamer
+        run: |
+           sudo apt update
+           sudo apt install -y libunwind-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev
 
       - name: Checkout Thunder
         if: steps.cache.outputs.cache-hit != 'true'
@@ -189,6 +194,8 @@ jobs:
           telemetry_busmessage_sender.h
           maintenanceMGR.h
           pkg.h
+          &&
+          cp -r /usr/include/gstreamer-1.0/gst /usr/include/glib-2.0/* /usr/lib/x86_64-linux-gnu/glib-2.0/include/* .
 
       - name: Set clang toolchain
         if: ${{ matrix.compiler == 'clang' }}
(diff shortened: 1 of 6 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -76,6 +76,11 @@ jobs:
           &&
           sudo apt install -y libsqlite3-dev libcurl4-openssl-dev valgrind lcov clang libsystemd-dev
 
+      - name: Install GStreamer
+        run: |
+           sudo apt update
+           sudo apt install -y libunwind-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev
```
varumugam123, line 82: Do we need the gstreamer packages now?

Tests/tests/test_TextToSpeech.cpp
```diff
@@ -0,0 +1,173 @@
...
+    EXPECT_EQ(
+    Core::ERROR_NONE,
+    handler.Invoke(
+        connection,
+        _T("setttsconfiguration"),
+        _T("{\"language\": \"en-US\",\"voice\": \"carol\",\"ttsendpointsecured\":\"http://cdn-ec-ric-312.voice-guidance-tts.xcr.comcast.net/tts/v1/cdn/location?\",\"ttsendpointsecured\":\"http://cdn-ec-ric-312.voice-guidance-tts.xcr.comcast.net/tts/v1/cdn/location?\"}"),
+        response
+    )
+    );
+    EXPECT_EQ(response, _T("{\"TTS_Status\":0,\"success\":true}"));
```
varumugam123, line 166: I see that all the APIs are accessed and values are compared against default expectations i.e (isttsenabled, listvoices). Should we not test those APIs are returning valid values after a 'setttsconfiguration' invocation?

Tests/tests/test_TextToSpeech.cpp
```diff
@@ -0,0 +1,173 @@
...
+}
+
+TEST_F(TTSInitializedTest,Test_SetACL_Success) {
```
varumugam123, line 169: Should we not test, if the SetACL is really preventing the calls from unauthorized clients?

.github/workflows/unit-tests.yml
varumugam123: I realized later that this is required to satisfy the gstreamer api uses in the Speaker implementation

Tests/tests/test_TextToSpeech.cpp
```diff
@@ -0,0 +1,224 @@
...
+    EXPECT_EQ(response, _T("{\"success\":true}"));
+
+    EXPECT_EQ(Core::ERROR_NONE, handler.Invoke(
+        connection,
+        _T("setttsconfiguration"), 
+        _T("{\"language\": \"en-US\",\"voice\": \"carol\",\"ttsendpointsecured\":\"http://cdn-ec-ric-312.voice-guidance-tts.xcr.comcast.net/tts/v1/cdn/location?\",\"ttsendpointsecured\":\"http://cdn-ec-ric-312.voice-guidance-tts.xcr.comcast.net/tts/v1/cdn/location?\"}"),
+        response
+    ));
+
+    EXPECT_EQ(Core::ERROR_GENERAL, handler.Invoke(connection, _T("speak"),
```
varumugam123, line 222: In this case, shouldn't we validate the return value? when speaking with 'WebAPP' the speak() API should proceed beyond access check (but ultimately fail because of the dependency in gst impl). But if a speack() request is made with some other callsign, it should fail with access check itself. Could you cover that part?

varumugam123 (commented):
I just have two queries, the rest looks fine.

varumugam123 (approved):
@cmurug728 Changes look good to me. Please get this approved & merged by @ramasat

ramasat (approved):

PR #4611, opened 2023-11-02
Telemetry2 logging and bring the missing changes from earlier branch

nrajan002c:


PerformanceMetrics/CMakeLists.txt
```diff
@@ -62,6 +63,8 @@ set_target_properties(${MODULE_NAME} PROPERTIES
         CXX_STANDARD 11
         CXX_STANDARD_REQUIRED YES)
 
+target_include_directories(${MODULE_NAME} PRIVATE ../helpers)
```
MFransen69, line 66: Could you please move this into the "SYSLOG" elseif above? The ../helpers will not be available in all builds (for example ours :) ) and this line will break those builds

MFransen69 (changes requested):

PerformanceMetrics/CMakeLists.txt
nrajan002c: Yes. I moved to SYSLOG portion.

MFransen69 (approved):

PR #4612, opened 2023-11-02
XIONE-13388 : Increase the timeout for SystemAudioPlayer - For Realtek

karuna2git:
In Thunder-R4, The SystemAudioPlayer is the first component to start activating and for some reason the activation takes more than 5sec and its timing out.

After sometime, If we manually activate, it activates successfully.

The issue is it happens only when you freshly flash the image. I suspected the gstreamer cache being wiped-out could be causing the issue as the next reboot works.  But at this point it is not clear; so just increasing the timeout and it fixes.

### DIFF
--- a/SystemAudioPlayer/SystemAudioPlayer.cpp
+++ b/SystemAudioPlayer/SystemAudioPlayer.cpp
@@ -57,7 +57,7 @@ namespace Plugin {
 
         _service->Register(&_notification);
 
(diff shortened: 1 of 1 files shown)


anand-ky:
approving for now to unblock Thunder 4.4 migration. Need to revisit.
@vdinak240 @varumugam123 - Pls check.

anand-ky (approved):

PR #4614, opened 2023-11-03
RDKShell: Skip SetFocus to false for Hibernated plugins

adrianM27:


RDKShell/RDKShell.cpp
```diff
@@ -6606,13 +6606,39 @@ namespace WPEFramework {
...
+                            //Check plugin state, skip for hibernated state
+                            string method = "status@" + previousFocusedIterator->first;
+                            Core::JSON::ArrayType<PluginHost::MetaData::Service> joResult;
+                            auto thunderController = getThunderControllerClient();
+                            uint32_t ret = thunderController->Get<Core::JSON::ArrayType<PluginHost::MetaData::Service>>(RDKSHELL_THUNDER_TIMEOUT, method.c_str(), joResult);
+                            bool isHibernated = false;
+                            if (ret == Core::ERROR_NONE && joResult.Length() > 0)
+                            {
+                                PluginHost::MetaData::Service service = joResult[0];
+                                if (service.JSONState == PluginHost::MetaData::Service::state::HIBERNATED)
```
adrianM27, line 6619: worth to consider excluding SUSPENDED state, only If we confirm the call to Focused(false) is not needed for hibernated or suspended apps

goruklu:
this looks good to me but @binuinbaraj should review as well.

RDKShell/RDKShell.cpp
```diff
@@ -3835,6 +3851,12 @@ namespace WPEFramework {
                     }
                 }
 
+                //Reset app suspended/hibernated for launch
```
goruklu, line 3854: Might be good to use the #ifdef HIBERNATE_SUPPORT_ENABLED everywhere we use suspendOrHibernateApps array and mutex.

RDKShell/RDKShell.cpp
```diff
@@ -197,6 +197,9 @@ bool sForceResidentAppLaunch = false;
 static bool sRunning = true;
 bool needsScreenshot = false;
 
+std::mutex suspendedOrHibernatedAppsMutex;
```
goruklu, line 200: Might be good to use the #ifdef HIBERNATE_SUPPORT_ENABLED everywhere we use suspendOrHibernateApps array and mutex.

RDKShell/RDKShell.cpp
```diff
@@ -606,6 +609,10 @@ namespace WPEFramework {
                mLaunchEnabled = false;
             }
 
+            suspendedOrHibernatedAppsMutex.lock();
```
goruklu, line 612: Might be good to use the #ifdef HIBERNATE_SUPPORT_ENABLED everywhere we use suspendOrHibernateApps array and mutex.

goruklu (changes requested):

binuinbaraj (commented):
Changes looks good.

goruklu (approved):

PR #4633, opened 2023-11-08
[RDKShell] RDKTV-26298: Split part of Deactivated into Deinitialized

asurdej-comcast:
With Thunder R4 Deactivated notification comes very early, before plugin destruction (Deactivation state in R2). As a result RDKShell destroys compositor for running plugin (like WebKit browser).
This change moves previous "Deactivated" state handling code into Deinitialized that is called after plugin destruction.


asurdej-comcast:
@karuna2git @madanagopalt Can you please take a look? This seems to be related with your prev changes in RDKTV-25106 and RDKTV-24839

binuinbaraj (approved):

PR #4670, opened 2023-11-20
RDK-40525: [Coverity] Various issues in rdkservices - Phase 1

melhar098:
Reason for change: Fix Coverity issues with security vulnerabilities in RDKServices
Test Procedure: See Jira ticket
Risks: None.
Priority: 1

### DIFF
--- a/Network/Network.cpp
+++ b/Network/Network.cpp
@@ -466,6 +466,7 @@ typedef struct _IARM_BUS_NetSrvMgr_Iface_EventData_t {
 
                         IARM_BUS_NetSrvMgr_Iface_EventData_t iarmData = { 0 };
                     strncpy(iarmData.setInterface, interface.c_str(), INTERFACE_SIZE);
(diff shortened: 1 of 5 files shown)


Network/Network.cpp
srikanth-vv, line 1: The same null termination changes for "iarmData.ipversion" should also apply for "iarmData.interface" also, since they both are declared as same size and doing strncpy() in this file.

SystemServices/SystemServices.cpp
```diff
@@ -650,6 +650,7 @@ namespace WPEFramework {
             strncpy(rebootParam.requestor, "SystemServices", sizeof(rebootParam.requestor));
             strncpy(rebootParam.reboot_reason_custom, customReason.c_str(), sizeof(rebootParam.reboot_reason_custom));
```
srikanth-vv, line 651: The same null termination changes should also apply for "rebootParam.reboot_reason_custom" also.

srikanth-vv (changes requested):

srikanth-vv (dismissed):

PR #4680, opened 2023-11-27
RDK-45598: Support Negative TC for TTS

ryadav115:
Reason for change: Added validation checks
Test Procedure: Regression Testing
Risks: None
Priority: P1
Signed-off-by: Rajan Y<rajan_yadav@Comcast.Com>

### DIFF
--- a/RDKShell/RDKShell.cpp
+++ b/RDKShell/RDKShell.cpp
@@ -196,6 +196,7 @@ bool sFactoryModeBlockResidentApp = false;
 bool sForceResidentAppLaunch = false;
 static bool sRunning = true;
 bool needsScreenshot = false;
+sem_t gInitializeSemaphore;
 
 #ifdef RDKSHELL_DUAL_ODM_SUPPORT
 static Device_Mode_FactoryModes_t sFactoryMode = DEVICE_MODE_CVTE_B1_AGING;
@@ -1640,6 +1641,7 @@ namespace WPEFramework {
             }
 
             mErmEnabled = CompositorController::isErmEnabled();
+            sem_init(&gInitializeSemaphore, 0, 0);
             shellThread = std::thread([=]() {
                 bool isRunning = true;
                 gRdkShellMutex.lock();
@@ -1687,6 +1689,7 @@ namespace WPEFramework {
                 isRunning = sRunning;
                 gRdkShellMutex.unlock();
                 gRdkShellSurfaceModeEnabled = CompositorController::isSurfaceModeEnabled();
+                sem_post(&gInitializeSemaphore);
                 while(isRunning) {
                   const double maxSleepTime = (1000 / gCurrentFramerate) * 1000;
                   double startFrameTime = RdkShell::microseconds();
@@ -1852,6 +1855,7 @@ namespace WPEFramework {
         RialtoConnector *rialtoBridge = new RialtoConnector();
         rialtoConnector = std::shared_ptr<RialtoConnector>(rialtoBridge);
 #endif //  ENABLE_RIALTO_FEATURE
+            sem_wait(&gInitializeSemaphore);
             return "";
         }
 
@@ -2032,6 +2036,7 @@ namespace WPEFramework {
             gKillClientRequests.clear();
             gRdkShellMutex.unlock();
             gExternalDestroyApplications.clear();
+            sem_destroy(&gInitializeSemaphore);
         }
 
         string RDKShell::Information() const


TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -129,16 +132,16 @@ namespace Plugin {
...
             string element;
             while (voices->Next(element) == true) {
                 arr.Add(JsonValue(element));
             }
             response["voices"] = arr;
-            response["TTS_Status"] = static_cast<uint32_t> (TTS::TTS_OK);
+            response["TTS_Status"] = status;
             voices->Release();
-            returnResponse(true);
+            returnResponse(status ? false:true);
```
varumugam123, line 144: This looks incorrect based on the above comment.

TextToSpeech/impl/TTSManager.h
```diff
@@ -26,7 +26,7 @@
 #include "TTSDownloader.h"
 
 #include <vector>
-
+#include <set>
```
varumugam123, line 29: Leave a new line after the file inclusion

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -166,17 +169,56 @@ namespace Plugin {
     {
         if(_tts) {
         Exchange::ITextToSpeech::Configuration config;
-        config.ttsEndPoint = GET_STR(parameters, "ttsendpoint", "");
-        config.ttsEndPointSecured = GET_STR(parameters, "ttsendpointsecured", "");
+        std::string proxyTtsEndPoint = GET_STR(parameters, "ttsendpoint", "");
+        std::regex urlPattern(R"((https?|ftp):\/\/\S+\/\S*)",std::regex::icase);
```
varumugam123, line 173: * please leave a comma after ','
* Please adhere to this style

> if (condition) {
> } else {
> }

* After a condition / loop block, please leave a new line
* TTS endpoint can have "http://" as well as "https://"
* Secure TTS endpoint can only have "https://"

varumugam123 (changes requested):
Will continue to review the rest and share the comments

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -166,17 +169,56 @@ namespace Plugin {
...
+            config.ttsEndPointSecured = proxyTtsEndPointSecured;
+        }
+        else {
+            config.ttsEndPointSecured = "";
+        }
         config.language = GET_STR(parameters, "language", "");
-        config.voice = GET_STR(parameters, "voice", "");
-        config.speechRate = GET_STR(parameters, "speechrate", "");
-        config.volume = (uint8_t) std::stod(GET_STR(parameters, "volume", "0.0"));
+        std::set<std::string> languageSet = {"en-US","es-MX","fr-CA","en-GB","de-DE","it-IT"};
```
varumugam123, line 188: This reference list must either be populated from the configuration json file or should include all the standard languages

varumugam123 (changes requested):

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -166,17 +169,56 @@ namespace Plugin {
...
+        config.language = languageSet.find(config.language) != languageSet.end() ? config.language : "";
+        std::string proxyVoice = GET_STR(parameters, "voice", "");
+        if(proxyVoice.c_str()[0] == ' ' || Utils::isValidUnsignedInt(const_cast<char*>(proxyVoice.c_str()))) {
```
varumugam123, line 191: I think this validation is redundant (i.e checking if the incoming value is integral type when a string is expected. Instead, we need to have a list of supported voice (either populated from the config json or an entire list of possible voices) and match the incoming value against it

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -166,17 +169,56 @@ namespace Plugin {
...
+            config.voice = "";
+        }
+        else {
+            config.voice = proxyVoice;
+        }
+        std::string proxySpeechRate = GET_STR(parameters, "speechrate", "");
+        if(proxySpeechRate == "slow" || proxySpeechRate == "medium" || proxySpeechRate == "fast" || proxySpeechRate == "faster" || proxySpeechRate == "fastest") {
```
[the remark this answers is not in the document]
varumugam123: I take my word back, I realized this is a new prop introduced for sky use cases and is different from the integral 'rate' configuration (it is going to be confusing in the future as well these configurations)

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -166,17 +169,56 @@ namespace Plugin {
...
+            config.speechRate = proxySpeechRate;
+        }
+        else {
+            config.speechRate = "";
+        }
+        uint8_t proxyVolume = 0;
+        try {
+            proxyVolume = (uint8_t) std::stod(GET_STR(parameters, "volume", "0.0"));
```
varumugam123, line 206: We should also try to formalize the type of the 'volume' configuration. I don't remember why it is kept as a string (i.e passed as string from ttsclient & gets converted to integer here vice versa. But the 'rate' configuration is properly sent as int to & fro client.

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -204,9 +235,18 @@ namespace Plugin {
 
         if(parameters.HasLabel("primvolduckpercent")) {
            int8_t  primVolDuck;
-           primVolDuck = (int8_t) std::stoi(GET_STR(parameters,"primvolduckpercent","-1"));
-           _tts->SetPrimaryVolDuck(primVolDuck);
+           InputValidation::Instance().addValidator("primvolduckpercent", ExpectedValues<std::string>("^-?[0-9]+$"));
```
varumugam123, line 238: Is a negative value expected?

varumugam123, line 238: Could we keep all the addValidator in TextToSpeechImplementation::Configure() API?

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -171,7 +196,12 @@ namespace Plugin {
         config.language = GET_STR(parameters, "language", "");
         config.voice = GET_STR(parameters, "voice", "");
         config.speechRate = GET_STR(parameters, "speechrate", "");
-        config.volume = (uint8_t) std::stod(GET_STR(parameters, "volume", "0.0"));
+        std::string proxyVolume = GET_STR(parameters, "volume", "0.0");
+        InputValidation::Instance().addValidator("double_str", ExpectedValues<std::string>("^-?[0-9]+(\\.[0-9]+)?"));
```
varumugam123, line 200: keep all the addValidator() in Configure() API

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -171,7 +196,12 @@ namespace Plugin {
...
+        if(!InputValidation::Instance().validate("double_str", proxyVolume))
+            return Core::ERROR_GENERAL;
+        config.volume = (uint8_t) std::stod(proxyVolume);
+
```
varumugam123, line 204: Please remove unnecessary new lines

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -95,20 +97,21 @@ namespace Plugin {
...
             JsonArray list = parameters["accesslist"].Array();
             JsonArray::Iterator it = list.Elements();
+            uint32_t status;
             while(it.Next())
             {
                 JsonObject accesslist = it.Current().Object();
                 if (accesslist.HasLabel("method") && accesslist.HasLabel("apps"))
                 {
-                    _tts->SetACL(accesslist["method"].String(),accesslist["apps"].String());
+                    status = _tts->SetACL(accesslist["method"].String(),accesslist["apps"].String());
```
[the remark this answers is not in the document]
varumugam123: How is the above comment handled?

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -21,12 +21,36 @@
...
+#include <regex>
+#include "validator.h"
 
 #define GET_STR(map, key, def) ((map.HasLabel(key) && !map[key].String().empty() && map[key].String() != "null") ? map[key].String() : def)
+#define IS_INTEGER(string) (Utils::isValidUnsignedInt(const_cast<char*>(proxyVoice.c_str())))
 
 namespace WPEFramework {
 namespace Plugin {
 
+    struct InputValidation
```
varumugam123, line 34: Please have this either in a separate header or as part of the validator header. And BTW, please rename validator.h to a suitable name which aligns with existing naming style.

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -21,12 +21,36 @@
 #include "UtilsJsonRpc.h"
 #include "UtilsUnused.h"
 #include "impl/TTSCommon.h"
+#include "UtilsisValidInt.h"
+#include <regex>
+#include "validator.h"
 
 #define GET_STR(map, key, def) ((map.HasLabel(key) && !map[key].String().empty() && map[key].String() != "null") ? map[key].String() : def)
+#define IS_INTEGER(string) (Utils::isValidUnsignedInt(const_cast<char*>(proxyVoice.c_str())))
```
varumugam123, line 29: IS_INTEGER always checks proxyVoice.c_str(), is it expected?

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -21,12 +21,36 @@
 #include "UtilsJsonRpc.h"
 #include "UtilsUnused.h"
 #include "impl/TTSCommon.h"
+#include "UtilsisValidInt.h"
+#include <regex>
```
varumugam123, line 25: is regex header need to be mentioned explicitly now?

TextToSpeech/TextToSpeechImplementation.h
```diff
@@ -137,6 +157,8 @@ namespace Plugin {
         static TTS::TTSManager* _ttsManager;
         mutable Core::CriticalSection _adminLock;
         std::list<Exchange::ITextToSpeech::INotification*> _notificationClients;
+        std::set<std::string> expectedLanguageSet;
```
varumugam123, line 160: Do they need to be member variables?

TextToSpeech/TextToSpeechImplementation.h
```diff
@@ -86,6 +86,26 @@ namespace Plugin {
...
+                {
+                    static InputValidation s_instance;
+                    return s_instance.m_validationManager;
+                }
+                private:
+                InputValidation() = default;
+
+                private:
+                ValidationManager m_validationManager;
+            };
```
varumugam123, line 108: Why is there again a InputValidation seen here, I saw one in JsonRpc implementation!!!

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -363,10 +468,16 @@ namespace Plugin {
         CHECK_TTS_MANAGER_RETURN_ON_FAIL();
         _adminLock.Lock();
         TTS::SpeechState state;
-        auto status = _ttsManager->getSpeechState(speechid, state);
+        auto status = TTS::TTS_FAIL;
+
+        if(InputValidation::Instance().validate("speechid",speechid))
```
varumugam123, line 473: Please leave a space after comma operator

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -249,7 +340,17 @@ namespace Plugin {
...
+        if(method != "speak") {
+           return Core::ERROR_GENERAL;
+        }
+
+        if(apps.empty() || apps.c_str()[0] == ' ' || apps == "NULL") {
+           return Core::ERROR_GENERAL;
+        }
+
         _adminLock.Lock();
+
```
varumugam123, line 353: can you please remove this unnecessary new line?

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -191,6 +255,29 @@ namespace Plugin {
     uint32_t TextToSpeechImplementation::SetConfiguration(const Exchange::ITextToSpeech::Configuration &object,Exchange::ITextToSpeech::TTSErrorDetail &ttsStatus)
     {
         TTS::Configuration config;
+        ttsStatus = (Exchange::ITextToSpeech::TTSErrorDetail) Exchange::ITextToSpeech::TTS_INVALID_CONFIGURATION;
+
+        if (!object.ttsEndPoint.empty() && !InputValidation::Instance().validate("ttsendpoint", object.ttsEndPoint))
```
varumugam123, line 260: These empty value check can be included in the regexp check / validator itself and this part can be simpler

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -191,6 +255,29 @@ namespace Plugin {
     uint32_t TextToSpeechImplementation::SetConfiguration(const Exchange::ITextToSpeech::Configuration &object,Exchange::ITextToSpeech::TTSErrorDetail &ttsStatus)
     {
         TTS::Configuration config;
```
varumugam123, line 257: This declaration can be moved down (i.e only after the validation you might need the config instance)

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -181,8 +240,13 @@ namespace Plugin {
...
         _adminLock.Lock();
         std::vector<std::string> voice;
-        auto status = _ttsManager->listVoices(language,voice);
+        auto status = TTS::TTS_FAIL;
+
+        if(InputValidation::Instance().validate("language",language))
+            status = _ttsManager->listVoices(language,voice);
+
         voices = (Core::Service<RPC::StringIterator>::Create<RPC::IStringIterator>(voice));
+
```
varumugam123, line 249: Please match the new lines after Lock() & before Unlock()

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -89,9 +106,51 @@ namespace Plugin {
...
-            while(it.Next())
+
+            while(it.Next()) {
                 ttsConfig->m_others["voice_for_local_" + string(it.Label())] = it.Current().String();
+                expectedLanguageSet.insert(string(it.Label()));
+                expectedVoicesSet.insert(it.Current().String());
+            }
         }
+
+        #ifndef UNIT_TESTING
```
varumugam123, line 117: Can you define these constants somewhere else (perhaps where you declare/define InputValidator) and keep this implementation simple?

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -62,6 +65,16 @@ namespace Plugin {
         JsonObject config;
         config.FromString(service->ConfigLine());
 
+        InputValidation::Instance().addValidator("ttsendpoint", ExpectedValues<std::string>("^https?://[a-zA-Z0-9]+.*"));
+        InputValidation::Instance().addValidator("ttsendpointsecured", ExpectedValues<std::string>("^https://[a-zA-Z0-9]+.*"));
+        InputValidation::Instance().addValidator("speechid", ExpectedValues<uint32_t>("^[0-9]+$"));
+        InputValidation::Instance().addValidator("speechid", (std::function<bool(const uint32_t &)>)[](const uint32_t &val) -> bool { return val > 0; });
```
varumugam123, line 71: Can this simply be ExpectedValues<uint32_t>(1, UINT32_MAX )

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -62,6 +65,16 @@ namespace Plugin {
...
+        InputValidation::Instance().addValidator("text", ExpectedValues<std::string>("^[a-zA-Z0-9]+"));
+        InputValidation::Instance().addValidator("speechrate", ExpectedValues<std::string>({"slow", "medium", "fast", "faster", "fastest"}));
+        InputValidation::Instance().addValidator("rate", (std::function<bool(const uint8_t &)>)[](const uint8_t &val) -> bool { return val==0 || (val >=1 && val <=100); });
```
varumugam123, line 74: val >= 0 && value <= 100? BTW, in which scenario value 0 is accepted?

varumugam123 (changes requested):

TextToSpeech/TextToSpeechJsonRpc.cpp
ryadav115: For the Testcase: setting value less than minimum value (-1) is being set, at JSON RPC level just checking if the provided input is number and at COM RPC level checking the range. 

So, I thought checking negative numbers as well.

TextToSpeech/TextToSpeechImplementation.cpp
ryadav115: The cases, where the label "rate" is not present the rate value is set to 0. Which means that, in case setconfiguration is called other than rate parameter input validator should return true.

varumugam123: Ok. In that case, can you configure the ExpectedValues<uint8_t>(0, 100)?

TextToSpeech/TextToSpeechJsonRpc.cpp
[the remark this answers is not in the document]
ryadav115: if(method != "speak") {
           return Core::ERROR_GENERAL;
        }

        if(apps.empty() || apps.c_str()[0] == ' ' || apps == "NULL") {
           return Core::ERROR_GENERAL;
        }
        
        Have added this checks in the SetACL method.

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -62,6 +65,17 @@ namespace Plugin {
         JsonObject config;
         config.FromString(service->ConfigLine());
 
+        InputValidation::Instance().addValidator("ttsendpoint", ExpectedValues<std::string>("^https?://[a-zA-Z0-9]+.*"));
+        InputValidation::Instance().addValidator("ttsendpointsecured", ExpectedValues<std::string>("^https://[a-zA-Z0-9]+.*"));
+        InputValidation::Instance().addValidator("double_str", ExpectedValues<std::string>("^-?[0-9]+(\\.[0-9]+)?"));
+        InputValidation::Instance().addValidator("speechid", ExpectedValues<uint32_t>(1, UINT32_MAX));
+        InputValidation::Instance().addValidator("text", ExpectedValues<std::string>("^[a-zA-Z0-9]+"));
```
varumugam123, line 72: I am little hesitant about validating the text input, the input could have underscores, question marks or even other characters which are not ascii. We can't throw error on such cases. Is it imperative to validate text?

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -62,6 +65,17 @@ namespace Plugin {
...
+        InputValidation::Instance().addValidator("speechrate", ExpectedValues<std::string>({"slow", "medium", "fast", "faster", "fastest"}));
```
varumugam123, line 73: May I know if any of these values are expected to be sent in uppercase from app?

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -62,6 +65,17 @@ namespace Plugin {
...
+        InputValidation::Instance().addValidator("rate", (std::function<bool(const uint8_t &)>)[](const uint8_t &val) -> bool { return (val >= 0 && val <=100); });
```
varumugam123, line 74: 'rate', 'volume' & 'setPrimaryVolDuck' values can be set as 'ExpectedValues<uint8_t>(0, 100)'

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -75,12 +89,24 @@ namespace Plugin {
         ttsConfig->setPrimVolDuck(std::stoi(GET_STR(config,"primvolduckpercent", "25")));
         ttsConfig->setSATPluginCallsign(GET_STR(config, "satplugincallsign", ""));
 
+        #ifndef UNIT_TESTING
```
varumugam123, line 92: I think, we can skip this condition check here to improve readability

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -363,10 +450,16 @@ namespace Plugin {
...
-        auto status = _ttsManager->getSpeechState(speechid, state);
+        auto status = TTS::TTS_FAIL;
+
+        if(InputValidation::Instance().validate("speechid", speechid))
+            status = _ttsManager->getSpeechState(speechid, state);
+
         _adminLock.Unlock();
-        if(status == TTS::TTS_OK)
+
+        if(status == TTS::TTS_OK) {
```
varumugam123, line 460: This addition of curly braces doesn't bring any value addition & non-functional, can we remove it?

varumugam123 (changes requested):

TextToSpeech/TextToSpeechJsonRpc.cpp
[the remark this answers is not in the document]
varumugam123: No I believe, the setACL is implemented in a way that it can accept access control directives for more than on methods. For eg 
[
    {"method": "speak", "apps": ["RA", "SAD", "HtmlApp-0"]}
    {"method": "enabletts", "apps" : ["RA"]},
    {"method": "setconfiguration", "apps" : ["RA", "SAD"]}
]

In that example, if the second item (or even the first for that matter) has any issue / error, it will be shadowed by the next item's status. So, you should bail out from the while loop as soon as you detect the first error.

TextToSpeech/TextToSpeechImplementation.cpp
ryadav115: I think text validator is not necessary in our cases. I think we can remove this.

TextToSpeech/TextToSpeechImplementation.cpp
ryadav115: I am not sure if the app sends in uppercase.

TextToSpeech/TextToSpeechImplementation.cpp
ryadav115: Since, writing regrex for empty string, would not be used for all for example: speechrate. Still, we need to check for the other cases.

TextToSpeech/TextToSpeechJsonRpc.cpp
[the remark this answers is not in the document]
ryadav115: Now, I am breaking the while loop as soon as the status == Core::ERROR_GENERAL.

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -203,10 +211,17 @@ namespace Plugin {
         }
 
         if(parameters.HasLabel("primvolduckpercent")) {
-           int8_t  primVolDuck;
-           primVolDuck = (int8_t) std::stoi(GET_STR(parameters,"primvolduckpercent","-1"));
-           _tts->SetPrimaryVolDuck(primVolDuck);
+            int8_t  primVolDuck;
+            std::string primVolDuckProxy = GET_STR(parameters,"primvolduckpercent","-1");
+
+            if(!InputValidation::Instance().validate("primvolduckpercent",primVolDuckProxy))
```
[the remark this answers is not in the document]
ryadav115: Yeah, we need logging that would explicitly state the invalid configuration. We may need to discuss any other ways to address the configuration states.

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -231,6 +300,10 @@ namespace Plugin {
     uint32_t TextToSpeechImplementation::SetPrimaryVolDuck(const uint8_t prim)
     {
         _adminLock.Lock();
+
```
varumugam123, line 303: This will lead to deadlock scenarios. It is now possible to return from SetPrimaryVolDuck() function with _adminLock locked. Next time any API is called, it will lead do deadlock.

varumugam123: No need to hold the admin lock here for validation, the lock can be simply moved down

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -326,7 +406,11 @@ namespace Plugin {
     {
         CHECK_TTS_MANAGER_RETURN_ON_FAIL();
         _adminLock.Lock();
-        auto status = _ttsManager->shut(speechid);
+        auto status = TTS::TTS_FAIL;
```
varumugam123, line 409: May I know why speechid is not validated on Pause() & Resume() operations?

TextToSpeech/TextToSpeechImplementation.h
varumugam123, line 1: Can we avoid making changes in this file?

TextToSpeech/TextToSpeechJsonRpc.cpp
```diff
@@ -171,7 +176,11 @@ namespace Plugin {
         config.language = GET_STR(parameters, "language", "");
         config.voice = GET_STR(parameters, "voice", "");
         config.speechRate = GET_STR(parameters, "speechrate", "");
-        config.volume = (uint8_t) std::stod(GET_STR(parameters, "volume", "0.0"));
+        std::string proxyVolume = GET_STR(parameters, "volume", "0.0");
+
+        if(!InputValidation::Instance().validate("double_str", proxyVolume))
+            return Core::ERROR_GENERAL;
```
varumugam123, line 182: When we bail out from SetConfiguration() prematurely, should we set TTS_Status value?

TextToSpeech/TextToSpeechImplementation.cpp
```diff
@@ -190,7 +235,31 @@ namespace Plugin {
 
     uint32_t TextToSpeechImplementation::SetConfiguration(const Exchange::ITextToSpeech::Configuration &object,Exchange::ITextToSpeech::TTSErrorDetail &ttsStatus)
     {
+        ttsStatus = (Exchange::ITextToSpeech::TTSErrorDetail) Exchange::ITextToSpeech::TTS_INVALID_CONFIGURATION;
+
+        if (!object.ttsEndPoint.empty() && !InputValidation::Instance().validate("ttsendpoint", object.ttsEndPoint))
+            return Core::ERROR_GENERAL;
+
```
varumugam123, line 242: Can we simply 'OR' these conditions?

TextToSpeech/TextToSpeechImplementation.cpp
ryadav115: if we pass invalid speechid  also it will equals with  currentspeechid which is speaking, Error will be thrown at that stage if speechid != currentspeechid

PR #4685, opened 2023-11-28
 DELIA-61988 : setTimeZoneDST Thunder plugin only accepts Olsen format

santoshcomcast:
Reason for change: Fix support for both Universal and Olsen formats. And fix the invalid response.
Test Procedure: Testing purpose. Valid case Universal and Olsen format and invalid timeZone(Ex :[UTC-5])

### DIFF
--- a/SystemServices/SystemServices.cpp
+++ b/SystemServices/SystemServices.cpp
@@ -2353,6 +2353,8 @@ namespace WPEFramework {
                 JsonObject& response)
 	{
 		bool resp = true;
+                bool isUniversal = false, isOlson = true;
+
 		if (parameters.HasLabel("timeZone")) {
 			std::string dir = dirnameOf(TZ_FILE);
 			std::string timeZone = "";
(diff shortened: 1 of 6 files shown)


santoshcomcast:
@srikanth-vv 
Could you please review these chanegs?

SystemServices/SystemServices.cpp
```diff
@@ -2362,11 +2364,21 @@ namespace WPEFramework {
...
 					LOGERR("Empty timeZone received.");
 				}
-				else if( (pos == string::npos) ||  ( (pos != string::npos) &&  (pos+1 == timeZone.length())  )   )
-				{
-					LOGERR("Invalid timezone format received : %s . Timezone should be in Olson format  Ex : America/New_York .  \n", timeZone.c_str());
-				}
-				else {
+
+                                if( (timeZone.compare("Universal")) == 0) {
+                                     isUniversal = true;
```
srikanth-vv, line 2369: Make "isOlson" false here and in the next if condition can be "if(isOlson)"

srikanth-vv (changes requested):
Please add "setTimeZoneDST()" with "Universal" in google test case and make sure the event is generating with correct data.

santoshcomcast:
@srikanth-vv 
I have updated the code as per suggestion and added unit-test cases for the Universal time zone.
Please review code.

santoshcomcast:
Hi @srikanth-vv 
I updated the test case, checked the locally run, and these test cases ran successfully.
Please review the changes.

srikanth-vv (approved):

PR #4731, opened 2023-12-12
Move InputValidator to Utils

varumugam123:
This PR has two 
1) Moving InputValidation implementation to Utils namespace so that other services can make use of it
2) Initialize & add validators for JSONRPC. TextToSpeechImplementation.cpp & TextToSpeechJsonRpc.cpp belong to different processes and dont share the same InputValidation::Instance(). The previous implementation adds validators only in TextToSpeechImplementation, leaving any validations from JsonRPC ineffective

### DIFF
--- a/TextToSpeech/TextToSpeechImplementation.cpp
+++ b/TextToSpeech/TextToSpeechImplementation.cpp
@@ -62,6 +62,7 @@ namespace Plugin {
         if(!_ttsManager)
             return 0;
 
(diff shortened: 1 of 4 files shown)


mhughesacn:
Hi Vivek: There isn't a Blackduck problem here but please will you put a header on the new file (RDK Apache).  I will clear off the check if it fails again.

ramasat (approved):

PR #4740, opened 2023-12-16
RDKTV-26356: Wifi SSID's are not scanning and only connected SSID is …

rajnishdey:
…shown

Reason for change: Resolve racing condition for wlan and p2p active sacn. Change P2P scan from active to passive scanning
Test Procedure: Refer RDKTV-26356.
Risks: None
Priority: P1


apatel859 (approved):

apatel859 (changes requested):
there will be one cpp file where u have to update version

apatel859 (approved):

PR #4798, opened 2024-01-09
RDKTV-27253: Physical Address CEC msg should be send synchronously and retried

dvm05:
Reason for change: For the CEC certification we are expecting that reportphysical
address responds back with Give physical Address in  synchronize manner and also
needs to retry at least twice.
Test Procedure: Build and Verify.
Risks: Low
Priority: P1
Signed-off-by: Dakshayani MV <dakshayani.venkataramana@sky.uk>


apatel859 (changes requested):
pls update the changelog history for version update

apatel859 (approved):

PR #4801, opened 2024-01-10
DELIA-61988 : setTimeZoneDST Thunder plugin only accepts Olsen format

santoshcomcast:
Reason for change: Add Universal time zone set support in setTimeZoneDST.
Test Procedure: Testing purpose. Valid case Universal and Olsen format and invalid timeZone(Ex :UTC-5)
Risks: Low
Priority: P1


srikanth-vv (approved):
Looks good

PR #4807, opened 2024-01-10
RDK-45464: [Coverity] Various issues in rdkservices - Phase 3a

agampa263:


Timer/Timer.cpp
```diff
@@ -491,7 +495,11 @@ namespace WPEFramework
                 dataToSend.length = sizeof(buf);
                 memcpy(dataToSend.data, buf, dataToSend.length);
                 LOGINFO("Timer send CEC %s", SLEEP == m_timerItems[timerId].mode ? "Standby" : "Wake");
-                IARM_Bus_Call(IARM_BUS_CECMGR_NAME,IARM_BUS_CECMGR_API_Send,(void *)&dataToSend, sizeof(dataToSend));
+                res=IARM_Bus_Call(IARM_BUS_CECMGR_NAME,IARM_BUS_CECMGR_API_Send,(void *)&dataToSend, sizeof(dataToSend));
```
srikanth-vv, line 498: Declare res here instead of top (limits the scope)
IARM_Result_t res = IARM_Bus_Call(IARM_BUS_CECMGR_NAME,IARM_BUS_CECMGR_API_Send,(void *)&dataToSend, sizeof(dataToSend));

Timer/Timer.cpp
```diff
@@ -18,6 +18,7 @@
 **/
 
 #include "Timer.h"
+#include "irMgr.h"
```
srikanth-vv, line 21: Why do you need this include? If not needed remove it.

Timer/Timer.cpp
```diff
@@ -316,7 +317,8 @@ namespace WPEFramework
 
             item.state = INITIAL;
             item.interval = std::stod(parameters["interval"].String());
-
+            item.repeatInterval = 0.0;
+            item.remindBefore = 0.0;
```
srikanth-vv, line 321: Also initialize reminderSent as false (CID 128215)

srikanth-vv (changes requested):

PR #4815, opened 2024-01-11
RDK-46636 - NetworkManager backward compatibility

gururaajar:
Reason for change: Need to support the network and Wi-Fi manager legacy API such that Network Manager can be used without any of the code for Network and Wi-Fi manager plugin
Test Procedure: Build and verified with curl commands Risks: Low
Priority: P1
Signed-off-by: Gururaaja ESR <Gururaja_ErodeSriranganRamlingham@comcast.com>


karuna2git (changes requested):
Please review.  Thanks

karuna2git (approved):

PR #4816, opened 2024-01-11
RDKTV-27601: RDKShell: fix deadlock on gSuspendedOrHibernatedApplicat…

adrianM27:
…ionsMutex


adrianM27:
A bit of explanation:
Current lock block  of gSuspendedOrHibernatedApplicationsMutex in
https://github.com/rdkcentral/rdkservices/blob/sprint/2401/RDKShell/RDKShell.cpp#L6577-L6600
is insecure and can cause deadlock due to calls to Thunder Core and Plugins APIs . 
At the same time, this mutex was preventing state change update in RDKShell and no single Hibernate trigger while executing "set focus to false" code, which could cause unintended Wakeup. 

In the provided solution, gSuspendedOrHibernatedApplicationsMutex is unlocked before any call to Thunder / plugin API, at the same time I added a mechanism to ensure Hibernate won't be triggered during "set focus to false" execution, similar to what mutex did.

goruklu (approved):

PR #4818, opened 2024-01-12
LLAMA-13120: HDMI In/AV In onInputStatusChanged event should include …

aishwariya15:
…Plane Parameter

Reason for change: Window manager uses plane parameter to examine if the Hdmi/Composite geometry is updated against correct video window.
Test Procedure: Build and verify
Risks: Medium
Priority: P1
Signed-off-by: Aishwariya B aishwariya.bhaskar@sky.uk


apatel859 (approved):
When u will return planeType =1

PR #4826, opened 2024-01-16
RDKTV-26816: Fix Network Retry Logic in Maintenance Manager

gomathishankar37:
**Reason for change:** Fixed Network Check logic in Maintenance Manager
* This Takes 1 Try for Network Check
* 4 Retries with 30 Seconds of Sleep between every retry checks

**Test Procedure:** Check Maintenance Manager Network Check and Retry
Mechanism in both Solicited and Unsolicited Maintenance
**Risks:** Medium
**Priority:** P1


tdeva14 (approved):
LGTM

PR #4836, opened 2024-01-18
BCM72180-895 BRCM DTV build is failing

philipjms:
Reason for change: adding DTV.conf.in file which is missing


philipjms:
Please review this change

philipjms:
> Compared to the DTV.config file, the "locator" variable is missing, and variables such as "startuporder", "subtitleprocessing" and "teletextprocessing" have been hardcoded when they should be set from environment variables, as is done in DTV.config.

Thanks for the comment. I re upload with the changes mentioned after doing sanity test.

PR #4880, opened 2024-02-02
RDKDEV-922: RDKServices : getPlatformConfiguration API in system plug…

VenkateshUmmadiSetty:
…in returning empty device type

Reason for change: Added required changes in rdkservices and sysint to get System plugin getPlatformConfiguration deviceType

Risks: Low

### DIFF
--- a/SystemServices/CHANGELOG.md
+++ b/SystemServices/CHANGELOG.md
@@ -16,6 +16,10 @@ All notable changes to this RDK Service will be documented in this file.
 
 * For more details, refer to [versioning](https://github.com/rdkcentral/rdkservices#versioning) section under Main README.
 
+## [2.0.4] - 2024-02-23
+### Fixed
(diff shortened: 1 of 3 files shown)


SystemServices/CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ All notable changes to this RDK Service will be documented in this file.
 ### Security
 - resolved security vulnerabilities
 
+## [1.7.4] - 2023-01-04
```
anand-ky, line 31: Revision cannot go backwards. Pls update patch version and move it to 2.0.3. You also need to update the version macros for the system plugin - https://github.com/rdkcentral/rdkservices/blob/main/SystemServices/SystemServices.cpp#L70

anand-ky (changes requested):

anand-ky (approved):

PR #4940, opened 2024-02-22
RDK-45037 : Secure Storage Thunder Plugin

npoltorapavlo:
Reason for change: Account scope, refactoring.
Tests written for all the test levels (L0,L1,L2) for the entire plugin. Test Procedure: None
Risks: None


mhughesacn:
Hi Nikita,
There are two sets of code matches because of the use of Protocol Buffers and gRPC.  Protocol Buffers code is OK in that there is an explicit permission.  I can't find the same for gRPC but it probably uses protocol buffers too.  To be sure, please append the following to top level NOTICE.  (I am assuming that the source protocols are owned by Comcast.)

```
This repository uses code generated by the protocol buffer compiler.
The protocol buffer compiler itself is:
Copyright 2008 Google Inc.  All rights reserved.
Licensed under the BSD-3 License.

Code generated by the Protocol Buffer compiler is owned by the owner
of the input file used when generating it i.e. Comcast.
and licensed the same as this repository.

The project uses code generated by the gRPC C++ plugin.  The gRPC C++ plugin is:
Copyright 2014 gRPC authors.
(https://grpc.io/docs/what-is-grpc/faq/)
```

Thank you.

anand-ky (approved):

PR #4954, opened 2024-02-26
Poc firebolt privacy

Karthick-Somasundaresan:


FireboltPrivacy/FireboltPrivacy.cpp
```diff
@@ -0,0 +1,134 @@
...
+            // Controls
+            {}
+        );      
+    }
+
+
+    FireboltPrivacy::FireboltPrivacy()
+        : _connectionId(0)
+        , _service(nullptr)
+        , _fireboltPrivacy()
```
MFransen69, line 46: this will not initialize the pointer to nullptr, use _fireboltPrivacy(nullptr)

FireboltPrivacy/FireboltPrivacy.cpp
```diff
@@ -0,0 +1,134 @@
...
+        ASSERT(_connectionId == 0);
+        ASSERT(_fireboltPrivacy == nullptr);
+        _service = service;
+        _service->AddRef();
+
+        // Register the Process::Notification stuff. The Remote process might die before we get a
+        // change to "register" the sink for these events !!! So do it ahead of instantiation.
+        _service->Register(&_connectionNotification);
+
+        _fireboltPrivacy = service->Root<Exchange::IFireboltPrivacy>(_connectionId, 2000, _T("FireboltPrivacyImplementation"));
```
MFransen69, line 67: Better to use the comrpc timeout (or if you want your own number don't use a magical but define a constant in the header file)

FireboltPrivacy/FireboltPrivacy.cpp
```diff
@@ -0,0 +1,134 @@
...
+        if (_fireboltPrivacy != nullptr) {
+            _fireboltPrivacy->Register(&_fireboltPrivacyNotification);
+            Exchange::JFireboltPrivacy::Register(*this, _fireboltPrivacy);
+
+            Exchange::IConfiguration* configFireboltPrivacy = _fireboltPrivacy->QueryInterface<Exchange::IConfiguration>();
+            if (configFireboltPrivacy != nullptr) {
+                configFireboltPrivacy->Configure(service);
+                configFireboltPrivacy->Release();
```
MFransen69, line 75: for safety set it to nullptr as well

FireboltPrivacy/FireboltPrivacy.cpp
```diff
@@ -0,0 +1,134 @@
...
+        _service->Register(&_connectionNotification);
+
+        _fireboltPrivacy = service->Root<Exchange::IFireboltPrivacy>(_connectionId, 2000, _T("FireboltPrivacyImplementation"));
+        if (_fireboltPrivacy != nullptr) {
+            _fireboltPrivacy->Register(&_fireboltPrivacyNotification);
+            Exchange::JFireboltPrivacy::Register(*this, _fireboltPrivacy);
+
+            Exchange::IConfiguration* configFireboltPrivacy = _fireboltPrivacy->QueryInterface<Exchange::IConfiguration>();
+            if (configFireboltPrivacy != nullptr) {
+                configFireboltPrivacy->Configure(service);
```
MFransen69, line 74: shouldn't you check the Configuration error code and return a message (e.g. configuration was invalid)

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    private:
+        class Config : public Core::JSON::Container {
+        public:
+            Config(const Config&) = delete;
+            Config& operator=(const Config&) = delete;
+
+            Config()
+                : Core::JSON::Container()
+                , devicemanifest("")
```
MFransen69, line 39: think "" is not needed, it is empty by default,  now it adds the overhead of creating a std::string first

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+                Add(_T("fireboltconfigpath"), &fireboltconfigpath);
+            }
+            ~Config() override = default;
+        public:
+            Core::JSON::String devicemanifest;
+            // This path is specific to firebolt. It points to the path where
+            // firebolt config files are placed. These files are shared between Ripple and other FireboltThunderPlugins
+            // In prod it points to /etc/
+            // and in dev it points to /opt/
+            Core::JSON::String fireboltconfigpath;
```
MFransen69, line 52: Perhaps this requires some architectural design. I certainly would not give all Ripple plugins this option as when  it changes you need to change the config of all plugins. Perhaps use an environment variable to specify the ripple config root or perhaps use a ripple code library that holds a lot of default functionality for Ripple (like we have our ThunderLibaries)? Then you can have consistency for all Ripple plugins.
But perhaps another question why would you need to access current Ripple configs, would that not be covered in a normal Thunder Plugin? Why would there be a ripple ecosystem and a Thunder ecosystem? If so I guess we would advice to move everything to Thunder plugins where every plugin has its own responsibility and if plugins need to communicate then do that via a strict defines abstract (comrpc) interface.

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+	        notification->OnAllowResumePointsChanged(allowResumePoints);
+	    }
+            _adminLock.Unlock();
+	    return 0;
+        }
+        uint32_t Configure(PluginHost::IShell* service)  override {
+	    ASSERT( service != nullptr);
+            SYSLOG(Logging::Notification, (_T("service Config: %s"), service->ConfigLine().c_str()));
+            _config.FromString(service->ConfigLine());
+            _deviceManifest = _config.fireboltconfigpath.IsSet() ? _config.fireboltconfigpath.Value(): "/etc/" + _config.devicemanifest.IsSet() ? _config.devicemanifest.Value() : "firebolt-device-manifest.json" ;
```
MFransen69, line 117: should have been default in the config class constructor

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        BEGIN_INTERFACE_MAP(FireboltPrivacyImplementation)
+        INTERFACE_ENTRY(Exchange::IFireboltPrivacy)
+        INTERFACE_ENTRY(Exchange::IConfiguration)
+        END_INTERFACE_MAP
+
+    private:
+        mutable Core::CriticalSection _adminLock;
+        Config _config;
```
MFransen69, line 128: Config does not need to be a member it only used in one method. Same goes for _deviceManifest

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+        string _deviceManifest;
+        bool _allowResumePoints;
+        PluginHost::IShell* _service;
+        std::vector<Exchange::IFireboltPrivacy::INotification*> _notifications;
+    };
+
+    SERVICE_REGISTRATION(FireboltPrivacyImplementation, 1, 0)
```
MFransen69, line 135: Also add a patch level (0)

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+        INTERFACE_ENTRY(Exchange::IConfiguration)
+        END_INTERFACE_MAP
+
+    private:
+        mutable Core::CriticalSection _adminLock;
+        Config _config;
+        string _deviceManifest;
+        bool _allowResumePoints;
+        PluginHost::IShell* _service;
+        std::vector<Exchange::IFireboltPrivacy::INotification*> _notifications;
```
MFransen69, line 132: we also use a using so it can be easily changed and used as base type for the iterators

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+        INTERFACE_ENTRY(Exchange::IFireboltPrivacy)
+        INTERFACE_ENTRY(Exchange::IConfiguration)
+        END_INTERFACE_MAP
+
+    private:
+        mutable Core::CriticalSection _adminLock;
+        Config _config;
+        string _deviceManifest;
+        bool _allowResumePoints;
+        PluginHost::IShell* _service;
```
MFransen69, line 131: Why keep this? It is not used

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+            // and in dev it points to /opt/
+            Core::JSON::String fireboltconfigpath;
+        };
+
+
+    public:
+        FireboltPrivacyImplementation(const FireboltPrivacyImplementation&) = delete;
+        FireboltPrivacyImplementation& operator= (const FireboltPrivacyImplementation&) = delete;
+        FireboltPrivacyImplementation()
+            : _adminLock()
```
MFransen69, line 60: Not that _service is needed but you forgot to initialize it to nullptr (and no the compiler will not do this for you in all cases)

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+            , _config()
+            , _deviceManifest()
+            , _allowResumePoints(false)
+        {
+        }
+        ~FireboltPrivacyImplementation() override
```
MFransen69, line 66: Think the destructor can be default as all members are not needed or destructed automatically.

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+            _deviceManifest.clear();
+            if (_service) {
+                _service->Release();
+                _service = nullptr;
+            }
+        }
+
+        uint32_t Register(Exchange::IFireboltPrivacy::INotification* notification) override
+        {
+            ASSERT(notification);
```
MFransen69, line 77: we always explicitly compare with a value, so in this case == nullptr

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+            _adminLock.Lock();
+            notification->AddRef();
+            _notifications.push_back(notification);
```
MFransen69, line 80: we also always have an assert that checks if you do not register the same notification twice (as that can lead to nasty bugs)

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+            _adminLock.Unlock();
+
+            return Core::ERROR_NONE;
+        }
+
+        uint32_t Unregister(Exchange::IFireboltPrivacy::INotification* notification) override
+        {
+            ASSERT(notification);
```
MFransen69, line 88: see above

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+            _adminLock.Lock();
+            auto item = std::find(_notifications.begin(), _notifications.end(), notification);
+            ASSERT(item != _notifications.end());
```
MFransen69, line 91: also add an if to protect for the case it is not found so it will not crash (we normally do that)

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+            _notifications.erase(item);
+            (*item)->Release();
```
MFransen69, line 93: Perhaps it might work but in some container cases the iterator becomes invalid so better to do notification->Release();

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+            _adminLock.Unlock();
+
+            return Core::ERROR_NONE;
+        }
+        uint32_t AllowResumePoints(bool& allowResumePoints /* @out */) const override {
+            _adminLock.Lock();
+            allowResumePoints = _allowResumePoints;
+            _adminLock.Unlock();
+	    return 0;
```
MFransen69, line 102: return Core::ERROR_NONE; ?

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+        }
+        uint32_t SetAllowResumePoints(const bool& allowResumePoints ) override {
+            _adminLock.Lock();
+            _allowResumePoints = allowResumePoints;
+	    for(auto notification: _notifications){
+	        notification->OnAllowResumePointsChanged(allowResumePoints);
+	    }
+            _adminLock.Unlock();
+	    return 0;
```
MFransen69, line 111: see above

FireboltPrivacy/FireboltPrivacyImplementation.cpp
```diff
@@ -0,0 +1,138 @@
...
+        }
+        uint32_t Configure(PluginHost::IShell* service)  override {
+	    ASSERT( service != nullptr);
+            SYSLOG(Logging::Notification, (_T("service Config: %s"), service->ConfigLine().c_str()));
```
MFransen69, line 115: Why always log the content of the config file??? This will flood the logging with useless info. Syslog is for info valuable for operators, traces for devolopers. The operator cannot act upon this info, but I also think a trace is useless, that makes more sense after 117 to print the actual value of deviceManifest.

FireboltPrivacy/Module.h
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#ifndef MODULE_NAME
+#define MODULE_NAME Plugin_FireboltPrivacy
+#endif
+
+#include <plugins/plugins.h>
```
MFransen69, line 26: It works as Plugins drag in everything in but it would be more correct to also explicitly list at least Core and Messaging

MFransen69 (changes requested):

PR #4959, opened 2024-02-27
RDK-47583 Implement GNOME support for NetworkManage

cmuhammedrafi:
WiFiConnect, WiFiDisconnect, GetConnectedSSID APis


mhughesacn (changes requested):
NetworkManagerGnomeWIFI.c has several code matches, but they are almost boilerplate i.e. the standard way to do something.  The code matches are to GPLv2 code which is not allowed.  What needs to be done: please revisit the areas I flagged, and the files generally.  Variable names should not match exactly unless they are used in API defs.  So e.g. "ap" is OK, but "con_name" is suspect.  Exactly matching comments are a no-no.  You cannot simply copy GPL code.

Thanks, Martin (RDK CMF Compliance Team)

Aside: I am sorry for the repeated blackduck scan notices on the PR.  This is because the way we organise projects in our scan tool.

PR #5008, opened 2024-03-11
RDKTV-29404: Apache 2K /Pioneer-2 - R2 Builds show Blank screen after boot up

tabbas651:
Reason for change: PersistentStore.json is not generated  in R2 build due to write_config() empty parameter. as per Thunder R2 release , we need to pass Plugin/Project name as parameter to write_config Test Procedure: Verify in Jenkin Build
Risks: High
Signed-off-by: Thamim  Razith <tabbas651@cable.comcast.com>

### DIFF
--- a/PersistentStore/CMakeLists.txt
+++ b/PersistentStore/CMakeLists.txt
@@ -17,12 +17,12 @@
 
 cmake_minimum_required(VERSION 3.14)
 
-project(PersistentStore)
+set(PLUGIN_NAME PersistentStore)
+set(MODULE_NAME ${NAMESPACE}${PLUGIN_NAME})
 
 set(CMAKE_CXX_STANDARD 11)
 
 find_package(WPEFramework)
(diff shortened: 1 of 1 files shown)


karuna2git (approved):

npoltorapavlo:
@tabbas651 I think NAMESPACE is set by find_package(WPEFramework), if that is the case changing the order like you did NAMESPACE is used before find_package(WPEFramework) and so can be empty

karuna2git:
@npoltorapavlo , the `find_package(WPEFramework)` within the Persistent folder's CMakeList.txt is not required at all as the main CMakeList.txt itself does the `find_package(WPEFramework)`.  So `NAMESPACE` is already defined.

npoltorapavlo:
@karuna2git I expected module to work without need in global CMakeList.txt, e.g. for the testing purposes

PR #5016, opened 2024-03-13
NetworkManger connectivitymonitor static class removed.

cmuhammedrafi:
connectivitymonitor and wifisignal monitor static class logic removed

### DIFF
--- a/NetworkManager/service/NetworkConnectivity.cpp
+++ b/NetworkManager/service/NetworkConnectivity.cpp
@@ -19,8 +19,6 @@
 namespace WPEFramework {
     namespace Plugin {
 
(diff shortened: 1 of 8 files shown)


karuna2git (approved):
Prefer having Connectivity and WiFiSignalStrength classes to have SetParent() kind of function of where you can pass `this` pointer from the `NetworkManagerImplementation` constructor.
OR
Let `NetworkManagerImplementation` owns the pointer to the instances of Connectivity and WiFiSignalStrength classes which can be instantiated during the constructor.

These classses can hav has private default constructor and a separate public constructor which takes `NetworkManagerImplementation` pointer as input.


For now, this is okey as it is tested.

PR #5048, opened 2024-03-21
Improve compatibility with Thunder R4

emutavchi:
This PR

 - improves logging by switching to SYSLOG for error reporting. It is useful to see errors in production config of Thunder as well
 - switches to Core/Messaging for Thunder R4
 - removes usage of Exchange::JBrowserCookieJar. It is not properly generated. Instead explicitly send 'cookiejarchanged' notification.


modeveci:
Webkit API fetching process ID is ok for me. @MFransen69 , can you review other changes? Thanks

WebKitBrowser/WebKitBrowser.cpp
```diff
@@ -279,8 +277,13 @@ namespace Plugin {
         if (path.empty() == false) {
             string fullPath = _persistentStoragePath + path;
             Core::Directory dir(fullPath.c_str());
-            if (!dir.Destroy(true)) {
-                TRACE(Trace::Error, (_T("Failed to delete %s\n"), fullPath.c_str()));
+#if defined(THUNDER_VERSION) && THUNDER_VERSION >= 4
```
MFransen69, line 280: Do we still expect changes on R2 that this is needed? (guess those build will not have updated hashes). 
This #if will conflict with builds that do not have these defines (like ours and other customers). 
At the same time we will look to make the Destroy changes backwards compatible, it should have not been made to brake backwards compatibility I think.

WebKitBrowser/WebKitBrowser.cpp
```diff
@@ -454,7 +460,11 @@ namespace WebKitBrowser {
             _children = Core::ProcessInfo::Iterator(_main.Id());
             return ((_startTime == TimePoint::min()) || (_main.IsActive() == true) ? 1 : 0) + _children.Count();
         }
+#if defined(THUNDER_VERSION) && THUNDER_VERSION >= 4
```
MFransen69, line 463: See comment above on the #if

WebKitBrowser/WebKitBrowser.cpp
```diff
@@ -350,6 +354,8 @@ namespace Plugin {
     {
         if (connection->Id() == _connectionId) {
 
+            SYSLOG(Logging::Notification, (_T("WebKitBrowser::Deactivated: { \"URL\": %.*s }"), 80, _lastURL.c_str()));
```
MFransen69, line 357: Is there really an need to have this syslog for every webkit deactivation? Why not have it as a trace? 
If it needed can I propose to move it the PerformanceMetrics plugin?

MFransen69 (changes requested):

WebKitBrowser/WebKitBrowser.cpp
emutavchi: Yes, we expect latest browser plugin revision to compile with R2.

WebKitBrowser/WebKitBrowser.cpp
emutavchi: Same as above.

WebKitBrowser/WebKitBrowser.cpp
emutavchi: PerformanceMetrics is not available for R2. I can make it a TRACE if that is a problem.

WebKitBrowser/WebKitBrowser.cpp
emutavchi: On the topic of backward compatibility: R4.4 removed Announcements() from the RPC::IIPCServer  interface. Do you know if that was intentional as well? It looks like we'll have to patch browser plugin to remove https://github.com/rdkcentral/rdkservices/blob/sprint/2401/WebKitBrowser/Extension/main.cpp#L81.
```
    PluginHost()
        : _engine(Core::ProxyType<RPC::InvokeServerType<2, 0, 4>>::Create())
        , _comClient(Core::ProxyType<RPC::CommunicatorClient>::Create(GetConnectionNode(), Core::ProxyType<Core::IIPCServer>(_engine)))
    {
        _engine->Announcements(_comClient->Announcement());  /// <<<-----
    }
  ```
 Any suggestions on how to handle it differently (in a way that is backward compatible to R2 and R4.2)?

MFransen69 (approved):

PR #5066, opened 2024-03-27
RDKDEV-1008: Upstream change to keep the system time consistent with the serial port

jithajamesp:
Reason for change: Keep the system time consistent with the serial port

Test Procedure: Test timezone curl command return expected responses

Risks: Low


jithajamesp:
Created new pull request to 24Q2 sprint branch: https://github.com/rdkcentral/rdkservices/pull/5191

PR #5076, opened 2024-03-27
RDKCOM-3765 : APIs and events for RCU Firmware Update

dwolaver:
Reason for change: allow RCU firmware updates via Remote Control plugin api's
Test Procedure: See JIRA ticket for details
Risks: Low
Priority: P1
Signed-off-by: David Wolaver <david_wolaver@cable.comcast.com>


RemoteControl/RemoteControl.json
```diff
@@ -180,6 +178,59 @@
             "type": "integer",
             "example": 1
         },
+        "macAddress": {
+            "summary": "The MAC address of the remote in hex-colon format",
+            "type": "string",
+            "example": "E8:1C:FD:9A:07:1E"
+        },
+        "sessionId": {
```
cbucht200, line 186: Should this be part of the firmwareUpdateStatus object? If not, we use sessionIds for other constructs, may be worth signifying that this session id is for fw update in the name

RemoteControl/RemoteControl.json
```diff
@@ -730,6 +781,106 @@
...
+                        "$ref": "#/definitions/success"
+                    }
+                },
+                "required":[
+                    "status",
+                    "success"
+                ]
+            }
+        },
+        "statusFirmwareUpdate": {
```
cbucht200, line 831: Isn't the firmwareUpdateStatus part of the main status event? Do we need a separate call just for firmware update status?

RemoteControl/RemoteControl.json
```diff
@@ -770,6 +921,24 @@
                     "status"
                 ]
             }
+        },
+        "onFirmwareUpdateProgress": {
```
cbucht200, line 925: Does the app need to know the progression steps of the update? It doesn't matter to me, but if they only care about knowing if it's successful or not, this may not be needed.

RemoteControl/RemoteControl.json
dwolaver: Right now, it's not a part of the main status event because i didn't want to clutter up the main status any further than it already is.  Also, I added a separate call because it's session based and there may be a use case where the app doesn't want to process events and instead just poll periodically until the session is completed.  However, I don't feel too strongly about this and we can omit the statusFirmwareUpdate call completely if we choose.
I considered adding an active session id field to the main status for each remote.  We may want to do this...

RemoteControl/RemoteControl.json
dwolaver: I don't see any problem with adding the mac address to the firmwareUpdateStatus object...

RemoteControl/RemoteControl.json
cbucht200: Gotcha, fine with this

RemoteControl/RemoteControl.json
cbucht200: I misunderstood the document structure. no comment here.

RemoteControl/RemoteControl.json
cbucht200: Misunderstood, no comment

PR #5042, opened 2024-03-21
Update LoggingUtils to work on cgroup v2 as well

varumugam123:
Kirkstone yocto version has cgroup v2. This PR is to adapt the changes in the cgroup version. Without this, the processess (UI, Web & Network) are not properly migrated to sky-jspp.service and the logs are collected in wpeframework.log against the requirement.

PR #5098, opened 2024-04-02
Cherry-pick WebKitBrowserPlugin changes from 2401 sprint branch

emutavchi:

PR #5102, opened 2024-04-03
Update LoggingUtils to work on cgroup v2 as well

varumugam123:
Cherry-picking https://github.com/rdkcentral/rdkservices/pull/5042 from sprint/2401


varumugam123:
I see that the change is already brought to sprint/24Q2 branch through https://github.com/rdkcentral/rdkservices/pull/5098

PR #5125, opened 2024-04-05
RDKDEV-1023: Make plugin autostart configurable from recipe

jithajamesp:
Reason for change: Platform specific autostart configurations to be set in platform layers

Risks: Low


pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-4805

PR #5132, opened 2024-04-08
RDKDEV-987 - Upstream PlayerInfo Resolution 2160P60 Changes

tkathi259:
Reason for change: After setting resolution 2160P60, while fetching for PlayerInfo.resolution its returning 2160P60 instead of 2160P. As this is generic change, need to upstream the AMLS905X4-763_2160p-PlayerInfo_Resolution.patch file.

Risks: Low

Test Procedure:
TDK PlayerInfo test cases: SetAndGet_All_Supported_Resolutions


tkathi259:
According to the component owner's response, because refresh rate is important in 2160p to load different video profiles based on frame rate, and changing anything in PlayerInfo requires reverifying all apps (Netflix, YT,...) that are currently running on all devices (Comcast, Sky, and LGIs devices), we abonded this PR changes and kept them as a patch in the SOC layer.

PR #5141, opened 2024-04-09
DELIA-61725: [SECVULN] Thunder Services - SystemServices - User Controlled Upload URL

melhar098:
Reason for change: Deprecating uploadLogs API
Test Procedure: Confirm API is deprecated and that it uses uploadLogsAsync instead when called
Risks: Low
Priority: P1


SystemServices/SystemServices.cpp
```diff
@@ -467,7 +467,7 @@ namespace WPEFramework {
             registerMethod(_T("getWakeupReason"), &SystemServices::getWakeupReason, this);
             registerMethod(_T("getLastWakeupKeyCode"), &SystemServices::getLastWakeupKeyCode, this);
 #endif
-            registerMethod("uploadLogs", &SystemServices::uploadLogs, this);
+            registerMethod("uploadLogs", &SystemServices::uploadLogsAsync, this);
```
srikanth-vv, line 470: Since the uploadLogs() function is not used in the code, that can be removed. Also check any other functions used inside uploadLogs() that are not used any where can be removed.

srikanth-vv (changes requested):

srikanth-vv (approved):
Looks good

PR #5156, opened 2024-04-12
DELIA-62404: Ensure Detached threads are quit before deactivating plu…

agampa263:
…gin to avoid crashes (System)

Reason for change: Detached threads are quit before deactivating plugin Test Procedure: Check no crashed is reported after call setmode api and deactivating Systemservices plugin Risks: Low
Priority: P1
Signed-off-by:AkshayKumar_Gampa <AkshayKumar_Gampa@comcast.com>


srikanth-vv (approved):
Looks good

PR #5162, opened 2024-04-15
RDK-48799: Enable/disable service worker support at runtime

filipe-norte-red:
Reason for change: Allow service worker support to be enabled only for apps that need it
Test Procedure: Confirm support is enabled/disabled using test page Risks: Low
Priority: P1


emutavchi (approved):
This depends on https://github.com/WebPlatformForEmbedded/WPEWebKit/pull/1312. Otherwise looks good to me.

modeveci (approved):

PR #5269, opened 2024-05-10
Locked API for HdcpProfile and locked API for XCast

kiran-kottarathil-infosys:
Updated the API lock implementation for HdcpProfile and XCast. Previously, the extension involved extending AbstractPluginWithApiAndIARMLock


kiran-kottarathil-infosys:
Need to add more details

PR #5201, opened 2024-04-24
SGMM393-181 : "onSystemPowerStateChanged" event is not getting triggered for ON state"

HarshiniRavichandran:
Fix setting LIGHT_SLEEP and STANDBY

It is required that LIGHT_SLEEP will behave in the same way as the
STANDBY. Therefore LIGHT_SLEEP state at the Power IARMMgr has been
changed as a fall through case to the STANDBY mode.

Moreover, the org.rdk.System plugin has to sent proper LIGHT_SLEEP
state if preferredStandbyMode is set to LIGHT_SLEEP.

This as well fixes the state reporting of the getPowerState method
in the org.rdk.System plugin.

PR #5316, opened 2024-05-20
SGMM393-181 : "onSystemPowerStateChanged" event is not getting triggered for ON state"

viveksinghnarwaria:
Fix setting LIGHT_SLEEP and STANDBY

It is required that LIGHT_SLEEP will behave in the same way as the
STANDBY. Therefore LIGHT_SLEEP state at the Power IARMMgr has been
changed as a fall through case to the STANDBY mode.

Moreover, the org.rdk.System plugin has to sent proper LIGHT_SLEEP
state if preferredStandbyMode is set to LIGHT_SLEEP.

This as well fixes the state reporting of the getPowerState method
in the org.rdk.System plugin.
Risk : Low

Signed-off-by: harshini_ravichandran@comcast.com


viveksinghnarwaria:
Cherry Picked from : https://github.com/rdkcentral/rdkservices/pull/5201

PR #5327, opened 2024-05-23
[WebKitBrowser] Updating methods deprecated since 2.42 WebKit API

VeithMetro:
`consoleMessageSentCallback()` has to be marked as `DEPRECATED`, since `webkit_console_message_get_text()` has been deprecated in version 2.40 of WebKit in this PR: https://github.com/WebKit/WebKit/pull/6784. It seems like at the moment there is no other way to get the console messages: https://bugs.webkit.org/show_bug.cgi?id=247791#c4.

### DIFF
--- a/WebKitBrowser/Extension/main.cpp
+++ b/WebKitBrowser/Extension/main.cpp
@@ -201,6 +201,7 @@ private:
                 G_CALLBACK(didStartProvisionalLoadForFrame), nullptr);
 #endif
     }
+PUSH_WARNING(DISABLE_WARNING_DEPRECATED_USE)
     static void consoleMessageSentCallback(VARIABLE_IS_NOT_USED WebKitWebPage* page, WebKitConsoleMessage* message, PluginHost* host)
     {
         string messageString = Core::ToString(webkit_console_message_get_text(message));
@@ -208,6 +209,7 @@ private:
 
         TRACE_GLOBAL(BrowserConsoleLog, (host->_consoleLogPrefix, messageString, line, 0));
     }
+POP_WARNING()
     static gboolean userMessageReceivedCallback(WebKitWebPage* page, WebKitUserMessage* message)
     {
         const char* name = webkit_user_message_get_name(message);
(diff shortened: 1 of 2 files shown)


modeveci (approved):

modeveci:
@emutavchi , can you review this PR?

WebKitBrowser/Extension/main.cpp
```diff
@@ -201,13 +201,15 @@ static class PluginHost {
                 G_CALLBACK(didStartProvisionalLoadForFrame), nullptr);
 #endif
     }
+PUSH_WARNING(DISABLE_WARNING_DEPRECATED_USE)
```
emutavchi, line 204: PUSH_WARNING / POP_WARNING  doesn't seem to be available in Thunder R2. We still have to build with 3 versions of Thunder: R2, R4.2 and R4.4. Could you please check if this will work with all 3 versions of Thunder?

emutavchi (changes requested):

WebKitBrowser/Extension/main.cpp
VeithMetro: Sorry, I was not aware we need this to work with Thunder R2, and you are correct that `PUSH_WARNING / POP_WARNING` are no available on R2.

In that case, I think we have two options here, either implement the `PUSH_WARNING / POP_WARNING` like it is in Portability here: https://github.com/rdkcentral/Thunder/blob/283b3d54334010403d85a4e69b3835de23e42331/Source/core/Portability.h#L100, or if the build does not need to be warning free, then we can just remove it. Thoughts @modeveci?

modeveci: @VeithMetro, for compatibility, it is better to get @MFransen69 feedback.

modeveci: Push/Pop warning checking may be removed from this PR, because the deprecated APIs (from wpe 2.40 until 2.45) brought back. As we do not have direct usage of any versions between 2.40-2.45. it may be good to not put any checks. Let me first confirm and then we can move on.

modeveci: @VeithMetro , we can remove POP/PUSH macros, I got confirmation! 2.38 and 2.46 will not give any warnings. 
Thank you very much! ;)

modeveci (approved):

MFransen69 (approved):

PR #5328, opened 2024-05-23
[WebKitBrowser] Updating IndexedDB to match version 2.42 of WPEWebKit

VeithMetro:
On top of that, base data and cache directories were added to WebsiteDataManager as persistent and volatile paths set through the Thunder configuration.


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2828,16 +2835,38 @@ static GSourceFuncs _handlerIntervention =
...
                     "local-storage-quota", localStorageDatabaseQuotaInBytes,
                     "indexeddb-directory", indexedDBPath,
-                    "per-origin-storage-quota", indexedDBSizeBytes,
+                    "origin-storage-ratio", originStorageRatio,
+                    "total-storage-ratio", totalStorageRatio,
+                    "base-data-directory", _service->PersistentPath(),
+                    "base-cache-directory", _service->VolatilePath(),
+                     nullptr);
+                g_free(indexedDBPath);
+#else
```
modeveci, line 2861: This is the 2.38 and earlier code-block. When this PR created there was no potential usage of indexdb. But with a prime app, there might be usage of indexdb. 

@emutavchi, to not change the config options for coming releases, shall we use same originstorageratio and totalstorageratio which are valid from 2.42 on (we can pass the total ratio empty as we can calculate from persistent path) from JSON config file? 

we are working on a document to explain all storage apis, but in the meantime you can find basic info from this ticket: https://github.com/WebPlatformForEmbedded/WPEWebKit/issues/1251#issuecomment-1847138858

If it is ok for you for 2.38 it will be like this:
auto* websiteDataManager = webkit_website_data_manager_new(
                     "local-storage-directory", wpeStoragePath,
                     "disk-cache-directory", wpeDiskCachePath,
                     "local-storage-quota", localStorageDatabaseQuotaInBytes,
                     "indexeddb-directory", indexedDBPath,
                     "per-origin-storage-quota", **CALCULATED FROM originStorageRatio**
                     "base-data-directory", _service->PersistentPath(),
                     "base-cache-directory", _service->VolatilePath(),
                      nullptr);

modeveci: FYI @VeithMetro , @MFransen69

VeithMetro: @modeveci I updated all *.config files which contained a `indexeddbsize` setting.

Two questions:

1. JSPP.config does not have `indexeddbsize` settings, but JSPP.conf.in does, should we add it to .config as well?
2. Should the `per-origin-storage-quota` be calculated as `LocalStorageSize` times `originStorageRatio` (in a separate `#if WEBKIT_CHECK_VERSION(2, 38, 0)`)?

modeveci: Hi @VeithMetro ,
1. Yes, you can add indexdbsize to JSPP.config, your PR adds config options to others as well. 
2. No, LocalStorageSize is different. You can calculate with "base-data-directory" which is persistentPath.

VeithMetro: I added indexdbsize to JSPP.config. But I am not completely sure about the `per-origin-storage-quota`.

Do we want to calculate the maximum possible size of a directory under `base-data-directory`, which like you said is `persistentPath`? In that case, if I am not mistaken, the `per-origin-storage-quota` would be equal to all free space on a disk which contains the `persistentPath' times `originStorageRatio`. Is this what we are looking for?

If so, I don't think there is a way to check a maximum possible size of a directory in Thunder, and I don't think there is a limit set on the maximum size of `persistentPath` in Thunder, so it is effectively a size of the whole mounted disk.

modeveci: RDK manifests set rdkservices' persistent path as /opt/persistent/rdkservices. And you are right that is basically pointing whole /dev/root filesystem. I think it is another discussion whether we will or want to give a maximum persistent storage size. 

So far,  to get the size of persistentPath, we can use Core::File::Size(theNameofDirectory)

**From 2.42 on, we can tell webkit to use as "total-storage-ratio" of persistent path for all domains. That will control and manage eviction algorithms.** 
E.g:
```
base-data-directory (persistent-path): 1GB
total-storage-ratio (0.1 of base-data-directory): 100MB
origin-storage-ratio(0.01 of base-data-directory): 10MB
```

**For 2.38, we can again fetch the size of persistent path by calling Core::File:Size API**
and e.g:
```
base-data-directory (persistent-path): 1GB
per-origin-storage-quota = 10MB = originStorageRatio(0.01) X persistentPath (1GB)
```

If @emutavchi and @MFransen69 are ok with this approach, then we can make changes.

emutavchi: @modeveci @VeithMetro /opt/persistent could be (and usually is) a separate partition. Perhaps 'WPEFramework::Core::Partition' can give you the size of partition and free space available there.

modeveci: Thank you for your feedback @emutavchi. to be more generic it is better to check the size with Core::File::Size. if the /opt/persistent is a partition it will be the size of that partition. if it is not a separate partition, It will be the size of /dev/root. With that, it can work for other platforms (including development platforms) as well.

VeithMetro: @modeveci is it now confirmed that we want to use `Core::File::Size` to get the free space available under `base-data-directory` path and then calculate the `per-origin-storage-quota`? 😄

modeveci: Let's start to use ::Size as it is generic. Please prepare your changes.

modeveci (dismissed):

MFransen69 (approved):

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2869,11 +2869,28 @@ static GSourceFuncs _handlerIntervention =
                      nullptr);
                 g_free(indexedDBPath);
 #else
+#if WEBKIT_CHECK_VERSION(2, 38, 0)
+                uint64_t indexedDBSizeBytes = 0;    // No limit by default, use WebKit defaults (1G at the moment of writing)
+                if (_config.OriginStorageRatio.IsSet() && _config.OriginStorageRatio.Value() != 0) {
+                    Core::File persistentPath(_service->PersistentPath());
+                    indexedDBSizeBytes = ((persistentPath.Core::File::Size() * _config.OriginStorageRatio.Value()) / 100);
```
asurdej-comcast, line 2876: Hi, @VeithMetro @modeveci
"persistentPath.Core::File::Size()" Does it truely return available disk space? With R4.2 it just returns directory size itself (without its contents) which is 4096 in all cases. Sources look similar in R4, master and other branches and there is just `stat()` call inside Thunder/Source/core/FileSystem.cpp:
```
#ifdef __POSIX__
        struct stat data;
        if (stat(_name.c_str(), &data) == 0) {
            _size = data.st_size;
```
On the other hand Core::Partition from the same FileSystem.h file correctly returns partition Size() and Free() space

modeveci: @asurdej-comcast You are right. I thought File::Size would do the same result but apparently not.
On the otherhand, Core::Partition uses statvfs which returns partition size when you passed a directory/file in the partition. It should serve the purpose. 
@VeithMetro Sorry, my misunderstading for File::Size, Can we use partition api?

modeveci (approved):

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -2811,20 +2814,24 @@ static GSourceFuncs _handlerIntervention =
                     wpeDiskCachePath = g_build_filename(g_get_user_cache_dir(), "wpe", "disk-cache", nullptr);
                 }
                 g_mkdir_with_parents(wpeDiskCachePath, 0700);
-
+#if WEBKIT_CHECK_VERSION(2, 42, 0)
```
emutavchi, line 2817: Can you enable this starting from 2.38, currently build is broken with:
```
WebKitImplementation.cpp:2883:44: error: 'indexedDBPath' was not declared in this scope
```

emutavchi (changes requested):

WebKitBrowser/WebKitImplementation.cpp
VeithMetro: Sure, done!

modeveci (approved):

PR #5350, opened 2024-05-29
RDKTV-28742: Activation of RA in RDKShell

fzahir786:
-To avoid launch of RA in factory mode


binuinbaraj (changes requested):
Please update the minor patch versions in RDKShell.cpp

binuinbaraj (approved):

binuinbaraj (approved):

PR #5370, opened 2024-06-03
RDK-49322: use manufacturer name from serialized location instead of …

praveenrajkg:
…serilization main


apatel859 (approved):

apatel859 (changes requested):
Pls update the change log file

praveenrajkg:
As per comment, added change.log file for systemservice plugin

apatel859 (approved):

PR #5372, opened 2024-06-03
Post a job to exit() instead of dispatching Deactivated on IShell

varumugam123:
This change is done to avoid crashes observed when WebKitImplementation posts a IShell::Job when WPEWebProcess is found unresponsive. But at the time of dispatching that job, the IShell handle seems invalid and this leads to crash.

This patch tries to avoid that path and simply exits the process.


emutavchi (approved):

MFransen69:
Mmm this seems more a workaround for another issue. The IShell (guess here you would mean the _service member) should be correctly recounted and should not become invalid, that indicates a problem with the refcount somewhere I think.  Even if the webprocess is frozen, WPEProcess should be correctly terminated (and will also be killed by Thunder including its child processes if it does not do that gracefully). 
Could you elaborate a bit more on the issue this change is trying to fix so we can see if we can fix the underlying issue instead?

varumugam123:
Hi @MFransen69, Following is the issue we have been investigating, WPEWebProcess crashed due to hang (i.e WPEWebProcess was killed by SIGFPE sent from WPEProcess aka UI Process). But immediately after that WPEProcess too gets crashed with SIGSEGV.

Below is the stack trace I obtained at the time of the crash : 
[WPEProcess-crash-on-WebProcess-hang.txt](https://github.com/user-attachments/files/15592825/WPEProcess-crash-on-WebProcess-hang.txt)

The crashed thread has these frames 
```
#0  0xacd53bc8 in typeinfo name for WPEFramework::PluginHost::ISubSystem::ISecurity () from ./usr/lib/wpeframework/proxystubs/libWPEFrameworkProxyStubs.so
#1  0x00023178 in WPEFramework::Core::ThreadPool::Minion::Process (this=this@entry=0x1d519f0) at /usr/src/debug/wpeframework/4.2+gitAUTOINC+8fa0abb34e-r0/git/Source/core/../core/ThreadPool.h:451
#2  0x000232ec in WPEFramework::Core::WorkerPool::Join (this=0x1d51920) at /usr/src/debug/wpeframework/4.2+gitAUTOINC+8fa0abb34e-r0/git/Source/core/../core/WorkerPool.h:380
#3  0x0001caaa in main (argc=<optimized out>, argv=<optimized out>) at /usr/src/debug/wpeframework/4.2+gitAUTOINC+8fa0abb34e-r0/git/Source/WPEProcess/Process.cpp:676
```

On checking frame 1, I see below
```
(gdb) frame 1
#1  0x00023178 in WPEFramework::Core::ThreadPool::Minion::Process (this=this@entry=0x1d519f0) at /usr/src/debug/wpeframework/4.2+gitAUTOINC+8fa0abb34e-r0/git/Source/core/../core/ThreadPool.h:451
451     /usr/src/debug/wpeframework/4.2+gitAUTOINC+8fa0abb34e-r0/git/Source/core/../core/ThreadPool.h: No such file or directory.

(gdb) print *this
$1 = {_parent = @0x1d51930, _dispatcher = 0x1d51c04, _adminLock = {m_syncMutex = {__data = {__lock = 0, __count = 0, __owner = 0, __kind = 1, __nusers = 0, {__spins = 0, __list = {__next = 0x0}}}, __size = '\000' <repeats 12 times>, "\001\000\000\000\000\000\000\000\000\000\000", __align = 0}}, _signal = {m_blManualReset = true, m_blCondition = false, m_syncAdminLock = {__data = {__lock = 0, __count = 0, __owner = 0, __kind = 0, __nusers = 0, {__spins = 0, __list = {__next = 0x0}}}, __size = '\000' <repeats 23 times>, __align = 0}, m_syncCondition = {__data = {__wseq = {__value64 = 0, __value32 = {__low = 0, __high = 0}}, __g1_start = {__value64 = 0, __value32 = {__low = 0, __high = 0}}, __g_refs = {0, 0}, __g_size = {0, 0}, __g1_orig_size = 0, __wrefs = 2, __g_signals = {0, 0}}, __size = '\000' <repeats 36 times>, "\002\000\000\000\000\000\000\000\000\000\000", __align = 0}}, _interestCount = {<std::__atomic_base<unsigned int>> = {static _S_alignment = 4, _M_i = 0}, <No data fields>}, _currentRequest = {_refCount = 0x1d413fc, _realObject = 0x1d413f0}, _runs = 57}

(gdb) info local
request = 0x1d413f0
job = <optimized out>

(gdb) print request
$2 = (WPEFramework::Core::IDispatch *) 0x1d413f0

(gdb) print *request
$3 = {<WPEFramework::Core::IDispatchType<void>> = {_vptr.IDispatchType = 0xb3f3b10c <vtable for WPEFramework::Core::ProxyObject<WPEFramework::PluginHost::IShell::Job>+8>}, <No data fields>}

(gdb) print *(PluginHost::IShell::Job*)request
$12 = {<WPEFramework::Core::IDispatch> = {<WPEFramework::Core::IDispatchType<void>> = {_vptr.IDispatchType = 0xb3f3b10c <vtable for WPEFramework::Core::ProxyObject<WPEFramework::PluginHost::IShell::Job>+8>}, <No data fields>}, _shell = 0x1d6e758, _state = WPEFramework::PluginHost::IShell::DEACTIVATED, _reason = WPEFramework::PluginHost::IShell::WATCHDOG_EXPIRED}

(gdb) print *(PluginHost::IShell*)((PluginHost::IShell::Job*)request)->_shell
$19 = {<WPEFramework::Core::IUnknown> = {<WPEFramework::Core::IReferenceCounted> = {_vptr.IReferenceCounted = 0xacd698c8 <vtable for WPEFramework::ProxyStubs::StateControlNotificationProxy+28>}, <No data fields>}, <No data fields>}
```

Here _shell has some non-null pointer, but upon trying to access it it crashes.

core/ThreadPool.h @ **4.2+gitAUTOINC+8fa0abb34e** has
```
422             void Process()
423             {
424                 _dispatcher->Initialize();
425
426                 while (_parent._queue.Extract(_currentRequest, infinite) == true) {
427
428                     ASSERT(_currentRequest.IsValid() == true);
429
430                     _runs++;
431
...
449                     IDispatch* request = &(*_currentRequest);
450
451                     _dispatcher->Dispatch(request);
452
...
```

MFransen69:
Hi @varumugam123 

We had a look and it is hard to understand the exact real root cause, looks more like memory corruption in general seen the callstack. As the exit is only done when there already is an issue we have no problem with it, guess you want to prevent the crash as it leads to a log being reported? 
Two remarks:
- please change the code a bit so that the exit is only done when WebKit is running OOP (otherise it will kill Thunder itself). e.g. you could use the IShell to call COMLink() to find out, if you are really afraid IShell is not reliable here anymore, which I'm not sure of, you could on start and keep that in a member.
- why would you do the exit in a job? There is a risk that also the job will not be handled correclty so I would just call exit immediately and not via a job.

varumugam123:
Hi @MFransen69, In practice WebKitBrowser is not supported to run as in-process-plugin. So I believe we need not to specially handle OOP case.

Also as to the deferred exit(), we had a race b/w a Deactivation request through Controller and the termination triggered from WebKitImplementation previously. We handled that race by sequencing the exit() to the worker thread (WebKitBrowser plugin's WPEProcess has only one worker), this way the external deactivation handling & the exit are kind of serialized.

MFransen69:
Hi @varumugam123 

Thanks for answering, as mentioned I'm fine with the job exit, could just imagine doing it directly could even be easier :)

On the in process handling, although it will not happen much besides for testing you can configure the plugin to run in process. Also as a generic concept (although using exit should be an exception :) ) and people looking at this code it should be that it will work both and out of process and not exit Thunder itself. Therefore I would like to ask to add the above and make sure exit is not called when the plugin is running in process.

MFransen69 (approved):

filipe-norte-red:
@MFransen69 , the above changes made their way to 24Q3 sprint, but are missing in the main branch. Is anything blocking the merge to main? 
Thanks

MFransen69:
Hi @filipe-norte-red I only approved the changes, think that would be a question to the submitter or Eugene.

PR #5388, opened 2024-06-06
SERXIONE-5242: After CDL, reboot device and stuck

cbucht200:
Reason for change: Crash on init due to ASSERT and uninitialized variable
Test Procedure: Test latest and reboot
Risks: Low
Signed-off-by: Chris Buchter


tdeva14 (approved):
LGTM

PR #5398, opened 2024-06-10
DELIA-65017 : Out-of-process improvements to PersistentStore

MManja792:
Merging the changes from Sprint 24Q2 ref PR#5178


npoltorapavlo:
This should be compatible with both "WPEFramework" and "Thunder". I'm not sure if replacing all "WPEFramework" to "Thunder" will do any good. Also DELIA-65017 is not approved for stable/release.

PR #5413, opened 2024-06-14
RDK-48604: New UserSettings Thunder Plugin

sithanda:
Adding doc entries for newly added UserSettings Plugin


sithanda:
Not required

PR #5415, opened 2024-06-14
RDKVREFPLT-1728 [BCM] [memcr] RDKShell OCIContainer memcr functionality support

DineshkumarJP:
Reason for change: Supported the OCIContainer memcr functionality.
Test Procedure: build and verify.
Risks: low.
Signed-off-by: Dinesh_kumar2@comcast.com

### DIFF
--- a/RDKShell/RDKShell.cpp
+++ b/RDKShell/RDKShell.cpp
@@ -782,7 +782,7 @@ namespace WPEFramework {
                     bool l2s = (gLaunchedToSuspended.find(mCallSign) != gLaunchedToSuspended.end());
                     gLaunchedToSuspendedMutex.unlock();
 
-                    if (!l2s && (mCallSign.find("Netflix") != std::string::npos || mCallSign.find("Cobalt") != std::string::npos))
+                    if (!l2s && (mCallSign.find("Netflix") != std::string::npos || mCallSign.find("Cobalt") != std::string::npos || mCallSign.find("Amazon") != std::string::npos || mCallSign.find("YouTube") != std::string::npos))
                     {
                         // call RDKShell.hibernate
                         std::thread requestsThread =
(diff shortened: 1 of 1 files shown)


RDKShell/RDKShell.cpp
```diff
@@ -6579,44 +6579,100 @@ namespace WPEFramework {
...
+                        std::cout << "Hibernation of " << callsign << " ignored!" << std::endl;
+                        JsonObject eventMsg;
+                        eventMsg["success"] = false;
+                        eventMsg["message"] = "hibernation blocked";
+                        notify(RDKShell::RDKSHELL_EVENT_ON_HIBERNATED, eventMsg);
+                        return;
                     }
+		    
+                    JsonObject eventMsg;
+		    if (!containerstatus0)
```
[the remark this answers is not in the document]
joseinweb: You can remove the containerstatus variable entirely and return the response in #6225

RDKShell/RDKShell.cpp
```diff
@@ -6633,6 +6689,41 @@ namespace WPEFramework {
...
+			else
+			{
+				response["message"] = "Container is not in a hibernate state that can`t be wakeup";
+				returnResponse(false);
+			}
+			if (!wakeupContainerResult["success"].Boolean())
+			{
+				status = false;
+				response["message"] = "Failed to restore/wakeup container";
+			}
```
joseinweb, line 6724: Shouldn't we add a returnresponse here?

joseinweb (changes requested):

RDKShell/RDKShell.cpp
```diff
@@ -6578,49 +6578,103 @@ namespace WPEFramework {
                         returnResponse(status);
                     }
                 }
+		// Container mode hibernate functionality handling
+		auto ociContainerPlugin = getOCIContainerPlugin();
+                if (!ociContainerPlugin)
+                {
+                    response["message"] = "OCIContainer plugin initialisation failed";
+                    returnResponse(false);
```
goruklu, line 6586: We don't use/enable OCIContainer plugin on some builds, this will break the hibernate feature.

goruklu (changes requested):

goruklu:
there are two way of using the memcr/hibernate feature:

1. Thunder's internal processcontainers methods for the Thunder plugins
2. OCIContainers hibernate/restore APIs (can be used for all containerized apps, native apps or thunder plugins)

for all Thunder plugins, we should be using the Thundercontroller apis.

OCIContainers hibernate/restore APIs should only be used for non-thunder plugin containers.

PR #5450, opened 2024-06-24
Fix to get Active Source Status when we put TV into HdmiCec Standby mode.

kpriya172:
Reason for change: Added fix to get status of "getActiveSourceStatus" when we put TV into org.rdk.HdmiCec_2.sendStandbyMessage.

Test Procedure: Build and verify. Use curl commands to verify the working using the methods getEnabled, getActiveSourceStatus & sendStandbyMessage of HdmiCec_2.

Risks: Low

Signed-off-by: Kumari Priyanka kumari_priyanka@comcast.com

### DIFF
--- a/HdmiCecSource/HdmiCecSource.cpp
+++ b/HdmiCecSource/HdmiCecSource.cpp
@@ -787,8 +787,16 @@ namespace WPEFramework
                         powerState = 0; 
                     }
                     else
+		    {
                         powerState = 1;
-
+		    }
+	            if(param->data.state.newState == IARM_BUS_PWRMGR_POWERSTATE_STANDBY)
+                    {
+                        // Handled standby event logic here
+                        isDeviceActiveSource = false;
+                        LOGINFO("ActiveSource isDeviceActiveSource status: %d \n", isDeviceActiveSource);
+                        HdmiCec_2::_instance->sendActiveSourceEvent();
+                    }
                 }
            }
        }


kpriya172:
Addressed the following review comments:

- create a new PR against the 24Q2 branch
- handle the changes in pwrMgrModeChangeEventHandler() instead of void HdmiCec_2Processor::process (const Standby &msg, const Header &header) function.
- Migrate HdmiCec_2 -> HdmiCecSource

HdmiCecSource/HdmiCecSource.cpp
```diff
@@ -787,8 +787,16 @@ namespace WPEFramework
                         powerState = 0; 
                     }
                     else
+		    {
                         powerState = 1;
-
+		    }
+	            if(param->data.state.newState == IARM_BUS_PWRMGR_POWERSTATE_STANDBY)
```
apatel859, line 793: what if device new state as DEEPSLEEP? can this logic fit in to above if else at line 785? also where do you send activesource true when device transition to ON state?

kpriya172: @apatel859, as per my understanding, we just need to send out the STANDBY event in order to put the TV in standby mode and IARM_BUS_PWRMGR_POWERSTATE_STANDBY does the job. IARM_BUS_PWRMGR_POWERSTATE_STANDBY_DEEP_SLEEP & IARM_BUS_PWRMGR_POWERSTATE_STANDBY_LIGHT_SLEEP are other 2 events but i dont see them getting used for hdmicec purposes. 

also for your 2nd query, we don't send the activesource to true; after giving the "HdmiCec_2.1.sendStandbyMessage" command, the TV goes in standby mode (turns off) and needs to wake up again using power button in remote. HdmiCec_2.1.getActiveSourceStatus command verifies the standby mode with the "false" status value.

HdmiCecSource/HdmiCecSource.cpp
```diff
@@ -787,8 +787,16 @@ namespace WPEFramework
...
+                    {
+                        // Handled standby event logic here
+                        isDeviceActiveSource = false;
+                        LOGINFO("ActiveSource isDeviceActiveSource status: %d \n", isDeviceActiveSource);
+                        HdmiCec_2::_instance->sendActiveSourceEvent();
```
viveksinghnarwaria, line 798: Hi @kpriya172 , There is no object HdmiCec_2. It is causing compilation failure. Please correct it. Also, please provide the latest log in the Broadcom accelerator device.

PR #5501, opened 2024-07-08
RDKDEV-1071 REFPLTV-2288 RDKServices : Wpeframework crash & restarting observed on setMode as Warehouse

selva11101992:
Reason for change: Added threadTimer.detach and removed join from stop timer to avoid first time setMode crash. While deactivating clear and join is called to avoid crash in deactivation.

Risks: Low

Signed-off-by: Selva Kumar MR selvakumar_mr@comcast.com


selva11101992:
https://github.com/rdk-e/rdkservices/pull/411/files

PR #5551, opened 2024-07-22
Rdk 39655

AnanthaC:
Notify the red recovery state changes


sborushevsky (approved):

anand-ky (changes requested):
Pls fix the L1/L2 errors. Need to merge the dependent IARM changes first so that CI tests pass.

PR #5556, opened 2024-07-23
[WebKitBrowserPlugin] improve handling of hang in Thunder

emutavchi:
Trying to gracefully shutdown browser pluing doesn't work if it is blocked in Thunder RPC.

Instead, kill the instance with SIGFPE so such crashes can be easier to find/track on crash portal by crash signal.


goruklu (approved):

emutavchi:
No feed back Metro since last week. Merging PR.

PR #5578, opened 2024-07-29
RDKTV-31562: Remove deprecated Cache APIs from System Service

ks734:


anand-ky (changes requested):
@ks734 - Pls update version in System services changelog (https://github.com/rdkcentral/rdkservices/blob/main/SystemServices/CHANGELOG.md) and source file (https://github.com/rdkcentral/rdkservices/blob/main/SystemServices/SystemServices.cpp#L68). Since this is a breaking change, bump up the major version number. So pls move from existing "2.3.2" to "3.0.0"

anand-ky (approved):

PR #5597, opened 2024-08-02
RDKTV-32271: Add additional Dolby Vision Game mode

tamilselvanak:
Reason for change: Header changes for DolbyGame Mode
Test Procedure:  refer ticket
Risks: None
Priority: P1
Signed-off-by: bp-tarumu252 <tamilselvan.arumugamkesavan@sky.uk>


AVOutput/AVOutputTV.cpp
```diff
@@ -2889,9 +2889,16 @@ namespace Plugin {
         std::string pqmode;
         std::string source;
         std::string format;
-        int lowLatencyIndex = 0;
+        int params[3]={0};
```
ddevad, line 2892: what are params here?  Can we give an appropriate name?

AVOutput/AVOutputTV.cpp
```diff
@@ -2911,23 +2918,30 @@ namespace Plugin {
             returnResponse(false);
         }
 
-        if( isSetRequired(pqmode,source,format) ) {
+        params[0]=lowLatencyIndex;
+        int retval= updateAVoutputTVParam("set","LowLatencyState",pqmode,source,format,PQ_PARAM_LOWLATENCY_STATE,params);
```
ddevad, line 2922: what are params here?  Can we give an appropriate name?

ddevad (changes requested):

AVOutput/AVOutputTV.cpp
tamilselvanak: These changes have been ported from the TVSettings plugin. The variable "params" is intended to hold the corresponding pqparams (brightness, contrast, lowlatency) values. However, it should be named "lowlatencyindex" since that name is already being used within this function.
We opted for a generic name like "params," and this value will be saved to the driver configuration file(ssmdata).

Currently, the generic name "params" is used in all set/reset functions. We can take this into account during the Llama AVOutput porting process.

AVOutput/AVOutputTV.cpp
tamilselvanak: These changes have been ported from the TVSettings plugin. The variable "params" is intended to hold the corresponding pqparams (brightness, contrast, lowlatency) values. However, it should be named "lowlatencyindex" since that name is already being used within this function.
We opted for a generic name like "params," and this value will be saved to the driver configuration file(ssmdata).

Currently, the generic name "params" is used in all set/reset functions. We can take this into account during the Llama AVOutput porting process.

ddevad (approved):

PR #5630, opened 2024-08-21
RDK-52419 : PrivacyMode support to SystemServices and Telemetry.

sborushevsky:


Telemetry/Telemetry.cpp
```diff
@@ -53,13 +54,18 @@
...
 #define DEFAULT_PROFILES_FILE "/etc/t2profiles/default.json"
 
+#define SYSTEMSERVICES_CALLSIGN "org.rdk.System"
+
 #define API_VERSION_NUMBER_MAJOR 1
 #define API_VERSION_NUMBER_MINOR 2
 #define API_VERSION_NUMBER_PATCH 2
 
 #ifdef HAS_RBUS
+#define RBUS_PRIVACY_MODE_EVENT_NAME "Device.X_RDKCENTRAL-COM_UserSettings.PrivacyModeChanged"
```
anand-ky, line 64: This should be "Device.X_RDKCENTRAL-COM_Privacy.PrivacyMode"

anand-ky (changes requested):

helpers/UtilsController.h
```diff
@@ -0,0 +1,190 @@
+#pragma once
```
anand-ky, line 1: We shouldn't use JSON-RPC for checking plugin status and activation. Pls use COM-RPC APIs similar to https://github.com/rdkcentral/rdkservices/blob/main/DisplaySettings/DisplaySettings.cpp#L215

anand-ky (changes requested):

helpers/UtilsController.h
sborushevsky: Updated  UtilsController.h to use com-rpc for activation.

anand-ky (approved):

PR #5632, opened 2024-08-22
DELIA-65834 : Skipped new mode request

ssitar583:
Reason for change: Throwing error in case of new mode request(when newmode requested is same as currentmode),thereby avoiding crash.

Test Procedure: 1.Set device to any of the modes(WAREHOUSE/EAS/NORMAL) for 30s.
		2.During that 30s,again make a request to set the same mode again.
		3.Check for WPEFramework crashes.

Risks: Low

### DIFF
--- a/NetworkManager/CMakeLists.txt
+++ b/NetworkManager/CMakeLists.txt
@@ -89,7 +89,10 @@ install(TARGETS ${MODULE_NAME} DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/${STORAGE
 # Build the implementation that runs out-of-process behind a COM-RPC interface
 
 if(ENABLE_GNOME_NETWORKMANAGER)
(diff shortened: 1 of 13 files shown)


SystemServices/SystemServices.cpp
```diff
@@ -1398,7 +1398,12 @@ namespace WPEFramework {
                     LOGWARN("request to switch to mode '%s' from mode '%s' \
                             with duration %d\n", newMode.c_str(),
                             oldMode.c_str(), duration);
-
+                     if(newMode == m_currentMode)
```
anand-ky, line 1401: Should we be checking for the same mode. We should check if existing setMode is in progress and the duration has not expired. So this should error out even if we set a different mode but before the previous setMode has expired.

anand-ky (changes requested):

PR #5684, opened 2024-09-07
RDK-52196: [SECVULN] Unsafe Use of strncpy C Functions

melhar098:
Reason for change: Resolve security vulnerabilities in rdkservices
Test Procedure: See ticket
Risks: Low
Priority: P1


XCast/XCast.cpp
```diff
@@ -943,29 +943,36 @@ void XCast::getUrlFromAppLaunchParams (const char *app_name, const char *payload
...
                 const char payload_key[] = "dialpayload=";
-                url_len -= sizeof(payload_key) - 1;
-                url_len -= strlen(payload);
                 if(url_len >= 0){
-                    if (has_query) strcat(url, "&");
-                    strcat(url, payload_key);
-                    strcat(url, payload);
-                    has_payload = 1;
+                    if (has_query) {
+                        snprintf(url + strlen(url), url_len, "&");
```
apatel859, line 953: third args should be format as used in line 946 "%s"

XCast/XCast.cpp
```diff
@@ -943,29 +943,36 @@ void XCast::getUrlFromAppLaunchParams (const char *app_name, const char *payload
...
                     LOGINFO("there is no enough room for payload\r\n");
                 }
             }
 
         if(additional_data_url != NULL){
-                if (has_query || has_payload) strcat(url, "&");
-                strcat(url, "additionalDataUrl=");
-            strcat(url, additional_data_url);
+                if ((has_query || has_payload) && url_len >= 0) {
+                    snprintf(url + strlen(url), url_len, "&");
```
apatel859, line 969: same as above comment, Also make sure in cutom build that after this change DAIL launch works fine so that we are not introducing any regression

PR #5717, opened 2024-09-17
RDKDEV-1084: Option to enable/disable webRTC ICE candidate filtering

marcin-mielczarczyk-red:
Added option to enable/disable webRTC ICE candidate filtering.
API to enable/disable webRTC ICE candidate filtering in WPE 2.38 has been delivered in below PR:
https://github.com/WebPlatformForEmbedded/WPEWebKit/pull/1402


pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-5062

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -3234,6 +3240,9 @@ static GSourceFuncs _handlerIntervention =
             // webaudio support
             WKPreferencesSetWebAudioEnabled(preferences, _config.WebAudioEnabled.Value());
 
+            // Turn on/off ICE candidate filtering
```
emutavchi, line 3243: Do you really use wpe-2.22 still? WebRTC is quite old there, it probably just drop the change.

emutavchi (changes requested):

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -3190,6 +3193,9 @@ static GSourceFuncs _handlerIntervention =
 
             auto preferences = WKPreferencesCreate();
 
+            // ICE candidate filtering
+            webkit_settings_set_enable_ice_candidate_filtering(preferences, _config.ICECandidateFilteringEnabled.Value());
```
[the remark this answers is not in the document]
marcin-mielczarczyk-red: > It doesn't make sense to use glib API with WK preferences. Perhaps merging problem?
Indeed, that was a merging problem. My original change was placed just after webaudio settings in GLIB part, but after merging to rdkservices upstream I didn't notice that this line was applied in the wrong place.

> I'd prefer to avoid binary dependency on webkit_settings_set_enable_ice_candidate_filtering. Please change to use object property setter instead.
Agreed and fixed.

WebKitBrowser/WebKitImplementation.cpp
marcin-mielczarczyk-red: We don't use 2.22 but I put it there just for consistency. It doesn't harm in case value is not set at all.

emutavchi: Since there is no use case (and an easy way to build it in sprint) I'd prefer to not add it. Could you please remove?

emutavchi (commented):
Otherwise looks good.

marcin-mielczarczyk-red:
> Since there is no use case (and an easy way to build it in sprint) I'd prefer to not add it. Could you please remove?
Done.

emutavchi (approved):

PR #5740, opened 2024-09-26
DELIA-65834 : Skipped new mode request

ssitar583:
Reason for change: Crash occurred when timer expired for Previous mode and trying to set the Normal mode and same time curl request trying to process it.Handled synchronization between the new mode curl request as well as setMode request for NORMAL(default) mode.

Test Procedure: 1.Set device to any of the modes(WAREHOUSE/EAS/NORMAL) for 30s.
			 2.During that 30s,again make a request to set the same/different mode again.
			 3.Check for crashes.


anand-ky:
@ssitar583 - Pls update CHANGELOG and version for this change. You can increment the patch version.

npoltorapavlo:
can you please stop using detach() in this repo. This is unloadable plugins, detach() is not appropriate. We had numerous crashes on stutdown due to people using detach() in this repo.

ssitar583:
@npoltorapavlo As part of DELIA-65834,I have just checked whether timer is in progress or not and no new detach calls were added as part of this handling.Those detach() calls were existing one.

npoltorapavlo:
@ssitar583 if you read my comment, it does not say you add detach(). It says it's wrong to use detach() in either form

PR #5741, opened 2024-09-26
RDK-53379: Deprecate DelayOffset APIs from DisplaySettings Thunder plugin

kprasad96:


apatel859 (approved):

kprasad96:
These APIs has to be removed from L1 and json file. So will bring in as a new PR

PR #5789, opened 2024-10-14
RDK-53610 - sample hello world plugin that uses auto generated stubs

madanagopalt:


HelloWorld/HelloWorldImplementation.h
```diff
@@ -0,0 +1,138 @@
...
+            Job() = delete;
+            Job(const Job&) = delete;
+            Job& operator=(const Job&) = delete;
+            ~Job() {
+                if (_helloWorldImplementation != nullptr) {
+                    _helloWorldImplementation->Release();
+                }
+            }
+
+            typedef websocketpp::client<websocketpp::config::asio_client> client;
```
mhughesacn, line 82: This is matching code from a websocket tutorial and needs to be credited.  1. Please insert here:
```
Copyright (c) 2014, Peter Thorson. All rights reserved.
Licensed under the BSD-3 License
```

2. Please append the same 2 lines to top level NOTICE
3. Please append a copy of the BSD-3 License to top level LICENSE.  (Unfortuately there isn't one already.)
4. This may require recipe checksum updates.

mfiess (approved):

PR #5795, opened 2024-10-16
RDK-53677: Analytics plugin phase2

adrianM27:


Analytics/IAnalytics.h
```diff
@@ -0,0 +1,60 @@
+/**
```
anand-ky, line 1: We shouldn't add the interface header file here directly. It should reside in ThunderInterfaces repo.

anand-ky (changes requested):

goruklu (dismissed):

PR #5880, opened 2024-11-22
RDKCOM-4606, RDKCOM-3765: New RemoteControl Plugin APIs

klu339:


egalla204 (approved):

egalla204 (changes requested):
you'll need to update the .md file as well for documentation.  I think there is also a CHANGELOG file you need to do as well

PR #5957, opened 2024-12-20
SERXIONE-6655: Xumo Content Store - Age Ratings

Dosakaya:
SERXIONE-6655: Xumo Content Store - Age Ratings - Failure in get/setBlockNotRatedContent Thunder API call.

Reason for Change: Matching curl calls to spec.
Test Procedure: Run usersettings curl commands on box.
Risks: HIGH
Priority: P0


anand-ky (approved):
Pls make a followup change to make a major version update since this is a breaking change and update CHANGELOG and versioning.

PR #5969, opened 2025-01-03
RDK-52028: Add CMS and WB to AVOutput

tamilselvanak:


upatel-sky (approved):
I think we do need to use capDetails_t. We can make use of capVectors_t with array size 1 which is same as capDetails_t. We can do this later as part of DV Mode redesign we are planning.

PR #6033, opened 2025-02-03
RDK-55619: Add reliable Maintenance Task status reporting

gomathishankar37:


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -140,7 +140,7 @@ string moduleStatusToString(IARM_Maint_module_status_t &status)
         case MAINT_LOGUPLOAD_ERROR:
             ret_status="MAINTENANCE_LOGUPLOAD_ERROR";
             break;
-        case MAINT_PINGTELEMETRY_COMPLETE:
+        case MAINT_PINGTELEMETRY_COMPLETE: /* TODO: Can these be removed */
```
tdeva14, line 143: Yes, lets remove anything unused

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -215,7 +215,7 @@ namespace WPEFramework {
         cSettings MaintenanceManager::m_setting(MAINTENANCE_MGR_RECORD_FILE);
 
         string task_names_foreground[]={
-            "/lib/rdk/RFCbase.sh",
+            "/lib/rdk/Start_RFC.sh",
             "/lib/rdk/swupdate_utility.sh >> /opt/logs/swupdate.log",
```
tdeva14, line 219: Can we rename swupdate_utility as well to Start_swupdate, incase its not too much of an ask?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -406,20 +406,58 @@ namespace WPEFramework {
...
                 cmd += "\0";
                 m_task_map[tasks[i]] = true;
 
                 if ( !m_abort_flag ){
                     LOGINFO("Starting Script (SM) :  %s \n",cmd.c_str());
-                    system(cmd.c_str());
-
-                    LOGINFO("Waiting to unlock.. [%d/%d]",i+1,(int)tasks.size());
-                    task_thread.wait(lck);
+                    task_status = system(cmd.c_str());
```
tdeva14, line 418: Where are we logging the detailed return status of this call?

Not sure if I can ask this. Why aren't we using v_secure_system() ?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -406,20 +406,58 @@ namespace WPEFramework {
...
+
+		    while(task_status != 0 && retry_count > 0 && !m_abort_flag){
```
tdeva14, line 420: Do we really need this loop? Can't we just `sleep`, decrement the iterator `i` and just `continue` if system() call failed?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -406,20 +406,58 @@ namespace WPEFramework {
...
+			task_status = system(cmd.c_str());
+			retry_count--;
+		    }
+		    if (task_status != 0){
+		        LOGINFO("Task Failed even after retry, setting task as Error");
+
+			const char* current_task = nullptr;
+			int complete_status = 0;
+
+			if (cmd.find("Start_RFC.sh") != string::npos){
```
tdeva14, line 433: These conditions are not efficient. We need to see if we can re-use any of the available maps or create one (but ensure there aren't too many)

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -406,20 +406,58 @@ namespace WPEFramework {
...
+			else if (cmd.find("swupdate_utility.sh") != std::string::npos) {
+			    current_task = task_names_foreground[1].c_str();
+			    complete_status = DIFD_COMPLETE;
+			} 
+			else if (cmd.find("Start_uploadSTBLogs.sh") != std::string::npos) {
+			    current_task = task_names_foreground[2].c_str();
+			    complete_status = LOGUPLOAD_COMPLETE;
+			}
+
+			if (current_task && m_task_map[current_task] != true) {
```
tdeva14, line 446: Not sure what this block does

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -406,20 +406,58 @@ namespace WPEFramework {
...
+			    current_task = task_names_foreground[0].c_str();
+			    complete_status = RFC_COMPLETE;
+			} 
+			else if (cmd.find("swupdate_utility.sh") != std::string::npos) {
+			    current_task = task_names_foreground[1].c_str();
+			    complete_status = DIFD_COMPLETE;
+			} 
+			else if (cmd.find("Start_uploadSTBLogs.sh") != std::string::npos) {
+			    current_task = task_names_foreground[2].c_str();
+			    complete_status = LOGUPLOAD_COMPLETE;
```
tdeva14, line 443: We need to make sure we end up in MAINTENANCE_INCOMPLETE task in case all attempts to start the task fails. Refer [1226](https://github.com/rdkcentral/rdkservices/blob/9883ef5c47507350b9a4694f5b85bec06abc792b/MaintenanceManager/MaintenanceManager.cpp#L1226)

tdeva14 (changes requested):
Pls refer to my comments

MaintenanceManager/MaintenanceManager.cpp
gomathishankar37: Will remove PINGTELEMETRY Status as they are not being used

MaintenanceManager/MaintenanceManager.cpp
gomathishankar37: Yes, we can rename it if @satya200 too is onboard with us, since this is a critical script and many references to "swupdate_utility.sh" might be present

MaintenanceManager/MaintenanceManager.cpp
gomathishankar37: Just added this loop just to have this code scalable, just incase if the retry count need to be increased we can just modify the `TASK_RETRY_COUNT` macro value in header file. If this is not required, we can just follow below approach as you suggested
```
task_status <- system() call
if task_status non-zero
    sleep(TASK_RETRY_DELAY)
    task_status <- system() call
    if task_status non-zero
        Task status <- Error
        continue
wait for task complete and proceed to next task
```

MaintenanceManager/MaintenanceManager.cpp
gomathishankar37: Yes, We can create a vector pair with status and taskname or a tuple with status, taskname and timeout value specifically for this purpose

MaintenanceManager/MaintenanceManager.cpp
gomathishankar37: for the Maintenance Cycle to end up with `MAINTENANCE_INCOMPLETE`, we might need to Skip the tasks (via ABORT) than via ERROR, we have ABORT Status only for FWDOWNLOAD Task, should we bring in ABORT Statuses for RFC and LogUpload as well for uniformity in this use case?

MaintenanceManager/MaintenanceManager.cpp
gomathishankar37: This block sets the ERROR status for current Task running, used similar approach as used in [IarmEventHandler()#L1143](https://github.com/rdkcentral/rdkservices/blob/9883ef5c47507350b9a4694f5b85bec06abc792b/MaintenanceManager/MaintenanceManager.cpp#L1143)

MaintenanceManager/MaintenanceManager.cpp
satya200: If we rename it, we have to modify multiple places, including test automation. So better we can keep same name.

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -215,7 +209,7 @@ namespace WPEFramework {
         cSettings MaintenanceManager::m_setting(MAINTENANCE_MGR_RECORD_FILE);
 
         string task_names_foreground[]={
-            "/lib/rdk/RFCbase.sh",
+            "/lib/rdk/Start_RFC.sh",
             "/lib/rdk/swupdate_utility.sh >> /opt/logs/swupdate.log",
```
gomathishankar37, line 213: Handle log redirect in wrapper script?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -215,7 +209,7 @@ namespace WPEFramework {
         cSettings MaintenanceManager::m_setting(MAINTENANCE_MGR_RECORD_FILE);
 
         string task_names_foreground[]={
-            "/lib/rdk/RFCbase.sh",
+            "/lib/rdk/Start_RFC.sh",
```
gomathishankar37, line 212: Can the wrapper and task scripts be added from a MACRO?

MaintenanceManager/MaintenanceManager.cpp
gomathishankar37: More info using WEXITSTATUS for system() call

MaintenanceManager/MaintenanceManager.cpp
gomathishankar37: Marking this as resolved

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -510,6 +560,181 @@ namespace WPEFramework {
...
+        }
+
+        // TBD: Check Implementation & add proper handling
+        void MaintenanceManager::timer_handler(int signo)
+        {
+            if (signo == SIGALRM)
+            {
+                LOGERR("[TIMEOUT HANDLER] Timeout reached for %s. Aborting task...", currentScript.c_str());
+                abortTask(currentScript.c_str(), SIGTERM);
+            }
```
gomathishankar37, line 735: Add logic to 
- send MAINT_<TASK>_ERROR
- thread notify to proceed to next task

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -510,6 +560,181 @@ namespace WPEFramework {
...
+        }
+
+        // TBD: Check Implementation & add proper handling
+        void MaintenanceManager::timer_handler(int signo)
+        {
+            if (signo == SIGALRM)
+            {
+                LOGERR("[TIMEOUT HANDLER] Timeout reached for %s. Aborting task...", currentScript.c_str());
+                abortTask(currentScript.c_str(), SIGTERM);
```
gomathishankar37, line 734: Don't set SIGTERM, let SIGABRT be sent as default signal

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -1655,24 +1921,25 @@ namespace WPEFramework {
                         LOGINFO("Task[%d] is false \n",i);
                     }
                 }
-
+                deleteTimer();
```
gomathishankar37, line 1924: can this be stopTimer()?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -1655,24 +1921,25 @@ namespace WPEFramework {
...
                 result=true;
             }
             else {
-                LOGERR("Failed to stopMaintenance without starting maintenance \n");
+                LOGERR("Failed to stopMaintenance without starting maintenance");
+                deleteTimer();
```
gomathishankar37, line 1929: Can this be stopTimer()?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -214,14 +204,25 @@ namespace WPEFramework {
 
         cSettings MaintenanceManager::m_setting(MAINTENANCE_MGR_RECORD_FILE);
 
+        /* Static Member Definitions */
+        timer_t MaintenanceManager::timerid;
+        string MaintenanceManager::currentScript;
```
gomathishankar37, line 209: currentScript -> currentTask

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -214,14 +204,25 @@ namespace WPEFramework {
...
+        bool MaintenanceManager::scriptCompleted = false;
```
gomathishankar37, line 210: scriptCompleted -> taskCompleted

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -375,52 +373,102 @@ namespace WPEFramework {
...
+                                m_task_map[current_task] = false;
+                            }
+                            stopTimer();
+                        }
+                    }
+                    else
+                    {
+                        stopTimer();
+                        LOGINFO("Waiting to unlock.. [%d/%d]", i + 1, (int)tasks.size());
+                        task_thread.wait(lck);
```
gomathishankar37, line 466: Add stopTimer() here after thread unblock
Add task_ as prefix for timer APIs

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -375,52 +373,102 @@ namespace WPEFramework {
...
+                {
+                    LOGINFO("Starting Script (SM) :  %s \n", cmd.c_str());
+                    currentTask = cmd;
+                    if (retry_count == TASK_RETRY_COUNT){
+                        LOGINFO("Starting Timer for %s \n", cmd.c_str());
+                        startTimer();
+                    }
+                    task_status = system(cmd.c_str());
+
+                    if (task_status != 0)
```
gomathishankar37, line 424: Add code comment for better understanding

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -375,52 +373,102 @@ namespace WPEFramework {
...
                 cmd += " &";
                 cmd += "\0";
                 m_task_map[tasks[i]] = true;
 
-                if ( !m_abort_flag ){
-                    LOGINFO("Starting Script (SM) :  %s \n",cmd.c_str());
-                    system(cmd.c_str());
+                if (!m_abort_flag)
+                {
+                    LOGINFO("Starting Script (SM) :  %s \n", cmd.c_str());
```
gomathishankar37, line 416: Script -> Task rename

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -375,52 +373,102 @@ namespace WPEFramework {
...
+                            retry_count--;
+                            continue;
+                        }
+                        else{
+                            LOGINFO("Task Failed even after retries, setting task as Error");
+
+                            const char *current_task = nullptr;
+                            int complete_status = 0;
+
+                            for (size_t j = 0; j < (sizeof(task_names_foreground)/sizeof(task_names_foreground[0])); j++){
```
gomathishankar37, line 440: Prefer a Mapping between task_names_foreground and task_complete_status to optimize this code snippet

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -375,52 +373,102 @@ namespace WPEFramework {
...
-                    LOGINFO("Waiting to unlock.. [%d/%d]",i+1,(int)tasks.size());
-                    task_thread.wait(lck);
+                            if (current_task && !m_task_map[current_task])
+                            {
+                                LOGINFO("Ignoring Error Event for Task: %s", current_task);
+                            }
+                            else if (current_task)
+                            {
+                                SET_STATUS(g_task_status, complete_status);
+                                task_thread.notify_one();
```
gomathishankar37, line 455: Handle Task map and need not notify thread

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -375,52 +373,102 @@ namespace WPEFramework {
...
 #endif
             std::unique_lock<std::mutex> lck(m_callMutex);
-            for( i = 0; i < tasks.size() && !m_abort_flag; i++) {
+            for (i = 0; i < static_cast<int>(tasks.size()) && !m_abort_flag; i++)
+            {
+                int task_status = -1;
                 cmd = tasks[i];
                 cmd += " &";
                 cmd += "\0";
                 m_task_map[tasks[i]] = true;
```
gomathishankar37, line 412: Have this to the happy path else

gomathishankar37, line 412: to 422

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -375,52 +373,102 @@ namespace WPEFramework {
...
+                            if (current_task && !m_task_map[current_task])
+                            {
+                                LOGINFO("Ignoring Error Event for Task: %s", current_task);
+                            }
+                            else if (current_task)
+                            {
+                                SET_STATUS(g_task_status, complete_status);
+                                task_thread.notify_one();
+                                LOGINFO("Error encountered in %s script task", current_task);
+                                m_task_map[current_task] = false;
```
gomathishankar37, line 457: to 426

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -510,6 +559,202 @@ namespace WPEFramework {
...
+        {
+            if (signo == SIGALRM)
+            {
+                LOGERR("Timeout reached for %s. Aborting task...", currentTask.c_str());
+                MaintenanceManager::_instance->abortTask(currentTask.c_str());
+
+                const char* failedTask = nullptr;
+                int complete_status = 0;
+
+                for(size_t j = 0; j < (sizeof(task_names_foreground)/ sizeof(task_names_foreground[0])); j++){
```
gomathishankar37, line 738: use task to status mapping

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -510,6 +559,202 @@ namespace WPEFramework {
...
+            }
+        }
+
+        /* TBD: static method uses MaintenanceManager instance to access non-static members and methods */
+        void MaintenanceManager::timer_handler(int signo)
+        {
+            if (signo == SIGALRM)
+            {
+                LOGERR("Timeout reached for %s. Aborting task...", currentTask.c_str());
+                MaintenanceManager::_instance->abortTask(currentTask.c_str());
```
gomathishankar37, line 733: ignore aborting the task but just set the task to Error

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -281,227 +350,316 @@ namespace WPEFramework {
...
+                    {
+                        task_status = system(cmd.c_str());
+                    }
+                    /* Set task_status purposefully to non-zero value to verify failure logic*/
+                    // task_status = -1;
+                    if (task_status != 0) /* system() call fails */
+                    {
+                        LOGINFO("%s invocation failed with return status %d", tasks[i].c_str(), WEXITSTATUS(task_status));
+                        m_task_map[tasks[i]] = false;
+                        if (retry_count > 0 && isTaskTimerStarted)
```
tdeva14, line 513: Whats the use of checking `isTaskTimerStarted` here?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -510,129 +668,497 @@ namespace WPEFramework {
...
+                {
+                    SET_STATUS(MaintenanceManager::_instance->g_task_status, complete_status);
+                    MaintenanceManager::_instance->task_thread.notify_one();
+                    LOGINFO("Set %s Task to ERROR", failedTask);
+                    MaintenanceManager::_instance->m_task_map[failedTask] = false;
+                }
+            }
+            else
+            {
+                LOGERR("Did not received SIGARLM");
```
tdeva14, line 950: Can we print the signal no here too just in case?

MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -510,129 +668,497 @@ namespace WPEFramework {
...
+                if (failedTask && !MaintenanceManager::_instance->m_task_map[failedTask])
+                {
+                    LOGINFO("Ignoring Error Event for Task: %s", failedTask);
+                }
+                else if (failedTask)
+                {
+                    SET_STATUS(MaintenanceManager::_instance->g_task_status, complete_status);
+                    MaintenanceManager::_instance->task_thread.notify_one();
+                    LOGINFO("Set %s Task to ERROR", failedTask);
+                    MaintenanceManager::_instance->m_task_map[failedTask] = false;
```
tdeva14, line 945: lets do this as the first thing here and then the rest

tdeva14 (changes requested):
Added minor comments. Mostly require cosmetic changes. Pls check

MaintenanceManager/MaintenanceManager.cpp
gomathishankar37: Additional condition to just see if the timer is Started, if not we just directly make the Task status to **ERROR**

tdeva14 (approved):
LGTM

PR #6078, opened 2025-02-20
DELIA-66742: Prevent Maintenance Status change on invalid stopMaintenance() jsonRPC call

gomathishankar37:


tdeva14 (dismissed):
LGTM

tdeva14 (changes requested):
Handle these cases as well

Requirement:
Calling stopMaintenance() stops any ongoing maintenance and the maintenance thread joins successfully

Test Cases:
Pre-requisite: Maintenance is running i.e. Maintenance thread is in execution (doesn't matter what it is doing)
Case #1: Maintenance thread is in network check loop
Case #2: Maintenance thread is in knowWhoAmI() loop
Case #3: Maintenance thread in blocked on WhoAmI event
Case #4: Maintenance thread is in maintenance task loop
Case #5: Maintenance thread is blocked on maintenance task event

PR #6103, opened 2025-02-28
RDK-56526: DS setForceHDRMode API - allow only subset of HDR modes

Srigayathry:
RDK-56526: DS setForceHDRMode API - allow only subset of HDR modes

Reason for change: Add subset of HDR modes which are allowed for forcing and check before calling setForceHDR API

Test Procedure: build and verify
Risks: Medium
Priority: P1
Signed-off-by: Srigayathry Pugazhenthi <srigayathry.pugazhenthi@sky.uk>


Srigayathry:
Not required

PR #6168, opened 2025-03-26
DELIA-66742: stopMaintenance in MM

gshankar37:


MaintenanceManager/MaintenanceManager.cpp
```diff
@@ -2292,36 +2291,32 @@ namespace WPEFramework
...
+                {
+                    LOGERR("task_stopTimer() did not stop the Timer...");
+                }
+                task_thread.notify_one();
+                if (m_thread.joinable())
+                {
+                    m_thread.join();
+                    LOGINFO("Thread joined successfully");
+                }
+                if (UNSOLICITED_MAINTENANCE == g_maintenance_type && !g_unsolicited_complete)
```
tdeva14, line 2308: This check must not be moved here. This is the one that prevents unsolicited maintenance from being called twice

tdeva14 (changes requested):

PR #6214, opened 2025-04-25
RDK-56367: Disable allm during power mode change

Srigayathry:
Reason for change: ALLM should be disabled after deepsleep/standby mode

Test Procedure: build and verify
Risks: Medium
Priority: P1
Signed-off-by: Srigayathry Pugazhenthi srigayathry.pugazhenthi@sky.uk


Srigayathry:
Not required.

PR #6221, opened 2025-04-28
DELIA-67918 : parse parameter as strgified json

npoltorapavlo:
Reason for change: users of api want to send stringified json
Test Procedure: None
Risks: None


bobseamon:
Keep up the great work!

PR #6324, opened 2025-07-09
test

benjaminqcox:


benjaminqcox:
This pull request includes changes to the `AVOutputTV.cpp` file within the `Plugin` namespace. The modifications include the addition of a hardcoded password, removal of a `returnResponse` call, and adjustments to variable initialization. Below is a summary of the most important changes:

### Security Concerns:
* Added a hardcoded password (`char* myPassword = "plaintext";`) in the `tvVideoFormatChangeHandler` function, which poses a significant security risk and should be addressed.

### Functional Changes:
* Removed the `returnResponse(true)` call from a function, which may affect the response handling logic. This change should be reviewed to ensure it does not cause unintended behavior.

### Code Simplification:
* Modified the initialization of the `TR181_ParamData_t` variable by removing zero-initialization (`TR181_ParamData_t param = {0};`) and replacing it with default initialization (`TR181_ParamData_t param;`). This change simplifies the code but should be verified for compatibility with the rest of the logic.

PR #6380, opened 2025-09-17
Safety checks to avoid crash on systemservices

preeja33:
This PR is to add some safetychecks especially on string manipulation side to avoid crashes reported around getTerritory/setterritory().

### DIFF
--- a/SystemServices/SystemServices.cpp
+++ b/SystemServices/SystemServices.cpp
@@ -2627,6 +2627,7 @@ namespace WPEFramework {
 	uint32_t SystemServices::setTerritory(const JsonObject& parameters, JsonObject& response)
 	{
 		bool resp = false;
+		std::lock_guard<std::mutex> lock(m_territoryMutex);
 		if(parameters.HasLabel("territory")){
 			makePersistentDir();
 			string regionStr = "";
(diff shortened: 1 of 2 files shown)


SystemServices/SystemServices.cpp
```diff
@@ -2627,6 +2627,7 @@ namespace WPEFramework {
 	uint32_t SystemServices::setTerritory(const JsonObject& parameters, JsonObject& response)
 	{
 		bool resp = false;
+		std::lock_guard<std::mutex> lock(m_territoryMutex);
```
srikanth-vv, line 2630: Please make this as const.

SystemServices/SystemServices.cpp
```diff
@@ -2704,6 +2705,7 @@ namespace WPEFramework {
 	uint32_t SystemServices::getTerritory(const JsonObject& parameters, JsonObject& response)
 	{
 		bool resp = true;
+		std::lock_guard<std::mutex> lock(m_territoryMutex);
```
srikanth-vv, line 2708: Please make this as const.

srikanth-vv (changes requested):

PR #6415, opened 2025-10-13
DELIA-68745 (Rdkemw 7050) (#6354)

npoltorapavlo:
* RDKEMW-7050 : fix coverity issues

Reason for change: initialize sqlite handle with nullptr to mute coverity report.
Test Procedure: None
Risks: None

* RDKEMW-7050 : fix coverity issues

Reason for change: return value from a function
using a reference parameter.
Test Procedure: None
Risks: None

---------


.github/workflows/PersistentStoreTests.yml
```diff
@@ -19,12 +19,6 @@ jobs:
           - src: 'PersistentStore/sqlite/l1test'
             build: 'build/sqlitel1test'
             test: 'sqlitel1test'
-          - src: 'PersistentStore/l0test'
```
anand-ky: Are we removing L0 & L1 tests?

PR #6425, opened 2025-10-29
DELIA-69357: Crash when deactivating Usersettings

preeja33:
Reason for change: Crash when deactivating Usersetting plugin
Test Procedure: compiled and verified
Priority: P1
Risks: Low
Version:Minor
Signed-off-by: preeja [preeja_raveendran@comcast.com](mailto:preeja_raveendran@comcast.com)


npoltorapavlo:
@preeja33  , you were supposed to add Unregister() call, that's it. It's only change that needs to be done here.

UserSettingsImplementation is ref counted, it can be referenced from elsewhere , and so not destroy when plugin deactivates. If you want to test, try to find where interface is used and deactivate those plugins as well.

PR #6427, opened 2025-10-31
DELIA-69147:In several places in RDKservices uninitialized variables are used

agampa263:
Reason for change: Fixing usage of uninitialized variables
Test Procedure: Regression testing on effected plugins
Risks: Low

Signed-off-by:AkshayKumar_Gampa <AkshayKumar_Gampa@comcast.com>

### DIFF
--- a/AVOutput/AVOutputTV.cpp
+++ b/AVOutput/AVOutputTV.cpp
@@ -3706,8 +3706,8 @@ namespace Plugin {
             gainInfo["to"]   = stoi(info.rangeVector[1]);
             response["gainInfo"]=gainInfo;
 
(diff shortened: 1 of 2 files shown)


sborushevsky (approved):

StateObserver/StateObserver.cpp
```diff
@@ -589,6 +591,7 @@ namespace WPEFramework {
 				}
 
 			}
+		}
 
 			response["properties"]=response_arr;
```
srikanth-vv: In case of error, what would be the output of the json request. Make sure the response still comes as valid error message.

srikanth-vv (changes requested):

PR #6467, opened 2026-06-04
RDKEMW-18711:  VG is not working in Xione Dhruv build

sergiygladkyy:
Read THUNDER_ACCESS port as JSON number and use loopback host

Reason for change:
- TTS OOP plugin /etc/WPEFramework/config.json fallback read "port" and "binding" as Core::JSON::String
- Core::JSON::String::Value() returns the quoted form for JSON string fields, so "binding":"0.0.0.0" yields an endpoint of "0.0.0.0":9998 (literal quotes) which Core::NodeId cannot parse
- An OOP child connects back to main on the same host; the listen binding is not the right connect target. Read port as DecUInt32 and use loopback.

Test Procedure: described in the ticket
Implements: code changes in TextToSpeech/impl
Risks: Low
Source: COMCAST
License: Apache-2.0
Upstream-Status: Pending

Version: patch

### DIFF
--- a/TextToSpeech/impl/SatToken.cpp
+++ b/TextToSpeech/impl/SatToken.cpp
@@ -50,10 +50,9 @@ string SatToken::getSecurityToken() {
         if(file.Open(true)) {
             JsonObject config;
             if(config.IElement::FromFile(file)) {
-                Core::JSON::String port = config.Get("port");
-                Core::JSON::String binding = config.Get("binding");
-                if(!binding.Value().empty() && !port.Value().empty())
-                    endpoint = binding.Value() + ":" + port.Value();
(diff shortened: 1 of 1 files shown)


sergiygladkyy:
Closing in favor of an equivalent patch against rdkcentral/entservices-mediaanddrm — that's the actual source-of-truth for the TextToSpeech plugin shipped in RDK-E develop builds (recipe entservices-mediaanddrm.bb in meta-rdk-video pins it as SRCREV). This rdkservices repo is the RDK-V line; the same SatToken JSON-parse bug exists here too but is not on the path for the Dhruv VG regression on RDKEMW-18711. Reopening or filing a separate RDK-V backport later if needed.

